PR #1, opened 2019-11-19
Allow re-signing to work in the vagrant/ansible env

alanhlam:
Since nebula-cert does not allow for host.crt or host.key to be overwritten when present, subsequent ansible runs break. This change will have ansible remove previously signed certificates before resigning in the vagrant test environment.


rawdigits:
👍

PR #3, opened 2019-11-19
make nebula a service that can install itself

rawdigits:
** Work in progress **

Allows the nebula binary to install itself under the standard service manager on many platforms via the github.com/kardianos/service library.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ all:
 	make NEBULA_CMD_PATH=${NEBULA_CMD_PATH} bin-windows
 
 bin:
-	go build -ldflags "-X main.Build=$(BUILD_NUMBER)" -o ./nebula ./cmd/nebula
+	go build -ldflags "-X main.Build=$(BUILD_NUMBER)" -o ./nebula ${NEBULA_CMD_PATH}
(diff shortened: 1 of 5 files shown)


nbrownus:
This would help to resolve #9 if we included a tap driver

rawdigits:
also #50

Makefile
```diff
@@ -1,55 +1,56 @@
...
-	make bin-arm6
-	make bin-arm64
-	make bin-darwin
-	make bin-windows
+	make NEBULA_CMD_PATH=${NEBULA_CMD_PATH} bin-linux
+	make NEBULA_CMD_PATH=${NEBULA_CMD_PATH} bin-arm
+	make NEBULA_CMD_PATH=${NEBULA_CMD_PATH} bin-arm6
+	make NEBULA_CMD_PATH=${NEBULA_CMD_PATH} bin-arm64
+	make NEBULA_CMD_PATH=${NEBULA_CMD_PATH} bin-darwin
+	make NEBULA_CMD_PATH=${NEBULA_CMD_PATH} bin-windows
```
wadey: hmm, why do we list these all as separate `make` commands instead of just making them dependencies of `all`, like:

    all: bin-linux bin-arm bin-arm6 bin-arm64 bin-darwin bin-windows

Seems like that would be simpler and then we don't have to worry about the env vars thing. Can you test that?

PR #16, opened 2019-11-21
Support Path MTU Discovery

wadey:
Overview
--------

This change allows the discovery and propagation of the path MTU. Until further
testing in the wild is done, you must set `tun.path_mtu_discovery = true` to
enable it.

This is change meant for situations like on AWS where you have a default
interface with MTU 9001 that only supports 1500 to internet destinations. It is
not anticipated to be used to discover the MTU over the internet, only to
the local gateway (although it should work in all cases)

Set `tun.mtu` to the highest mtu supported minus 60. Nebula will check for
DontFragment on outgoing packets and respond with ICMP "Destination Unreachable
Fragmentation Needed" if the packet is too large, which allows TCP
connections to the tun device to automatically set MSS (see RFC 1191).

An example config that could be used:

    preferred_ranges:
      - 10.0.0.0/9
    tun:
      path_mtu_discovery: true
      mtu: 8941 # (9001 - 60)

This would advertise a high MTU on the tun device, and use path MTU
discovery to reduce the MTU on any paths that end up over the internet.

Design
------

The initial implementation is only for Linux.

There are multiple ways to implement this, each with their own trade
offs:

1) Use IP_PMTUDISC_WANT on the outside socket.

- Pros: simple support for sending packets that want fragmentation.
- Cons: Need cache invalidation. There could be a race between checking
  the current MTU and the OS discovering the actual MTU if other
  processes are connecting to the same remote.

2) Use IP_PMTUDISC_DO and manualy fragment large packets.

- Pros: Don't need cache invalidation.
- Cons: Nebula meta packets would need fragmentation support,
  complicating the implementation.

3) Use IP_PMTUDISC_DO and use raw sockets to send large packets.

- Pros: Don't have to deal with fragmentation logic in Nebula.
- Cons: Two code paths for sending, and requires raw socket support.

4) Use IP_PMTUDISC_DO and use SO_REUSEPORT with a second socket to send
   large packets.

- Pros: Same as (3) but you don't need raw sockets.
- Cons: Using SO_REUSEPORT allows weird things to happen if you
  accidentally start two Nebula processes or something else listening on
  the same port.

After reviewing these options, (1) was selected for the first
implementation because it is fairly easy and the race conditions should
be rare. Switching to (2) or (3) in the future might be a good idea if
we review the trade offs more.

Implementation
--------------

`IP_RECVERR` is set on the outside UDP socket so that we can receive the
FragmentationNeeded packets that come from upstream. We set up
a goroutine to poll for these events and update our cache of the route MTU.

When a new route is attempted, we first look up IP_MTU on the socket to
see what the OS has currently cached as the MTU for the route. Because
we are using IP_PMTUDISC_WANT we will only receive error events when
the OS cached IP_MTU is wrong, so we need to ensure we check it before
sending packets. We internally cache the MTU for 1 minute before checking
IP_MTU again.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -96,6 +96,16 @@ tun:
   tx_queue: 500
   # Default MTU for every packet, safe setting is (and the default) 1300 for internet based traffic
   mtu: 1300
+  # NOTE: EXPERIMENTAL
+  # Enable discovery of path MTU. This is meant for situations like on AWS
+  # where you have a default interface with mtu 9001 that only supports 1500 to
+  # internet destinations. This feature will allow the mtu to automatically be
+  # set as high as possible for both local connections and internet
+  # connections. Set `tun.mtu` to the highest mtu supported minus 60 (for AWS,
+  # when using Transit Gateway, use 8440, otherwise use 8941).
+  # Nebula will check for DontFragment on outgoing packets and respond with
+  # ICMP FragmentationNeeded if the packet is too large.
+  path_mtu_discovery: false
   # Route based MTU overrides, you have known vpn ip paths that can support larger MTUs you can increase/decrease them here
   routes:
     #- mtu: 8800
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/flynn/go-shlex v0.0.0-20150515145356-3f9db97f8568 // indirect
 	github.com/flynn/noise v0.0.0-20180327030543-2492fe189ae6
 	github.com/golang/protobuf v1.3.1
+	github.com/google/gopacket v1.1.17
 	github.com/imdario/mergo v0.3.7
 	github.com/kardianos/service v1.0.0
 	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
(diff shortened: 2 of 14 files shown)


gboddin:
Hey,

Could this missing feature prevent protocols like vxlan ( that won't fragment ) from working ?

ATM running vxlan with a default mtu of 1500 breaks under nebula 1.4.0 with mashed-up packets received by the other end.

It might prevent most container solutions ( swarm, k8s ) from running their network over nebula.

nbrownus (commented):
> Hey,
> 
> Could this missing feature help protocols like vxlan ( that won't fragment ) ?
> 
> ATM running vxlan with a default mtu of 1500 breaks under nebula 1.4.0 with mashed-up packets received by the other end.
> 
> It might prevent most container solutions ( swarm, k8s ) from running their network over nebula as they require developers to specify custom network settings in their stack files.

Yes and you should be able to accomplish this today by ensuring your VXLAN MTU is no larger that nebulas `tun.mtu` setting.

gboddin:
> > Hey,
> > Could this missing feature help protocols like vxlan ( that won't fragment ) ?
> > ATM running vxlan with a default mtu of 1500 breaks under nebula 1.4.0 with mashed-up packets received by the other end.
> > It might prevent most container solutions ( swarm, k8s ) from running their network over nebula as they require developers to specify custom network settings in their stack files.
> 
> Yes and you should be able to accomplish this today by ensuring your VXLAN MTU is no larger that nebulas `tun.mtu` setting.

Yeah I confirmed it working with an MTU lower than Nebula's interface.

Sadly docker swarm ( for instance ) doesn't make it easy to fix the issue : https://github.com/moby/moby/issues/34981 , which makes re-using code for infra impossible  :') 

Thanks for the confirmation, I'll keep looking.

PR #17, opened 2019-11-21
Add a configuration option to change the DNS server listen port.

apognu:
Allowing the specification of a custom port for the embedded DNS server would allow nebula to integrate onto servers where a DNS server is already running.


coderobe:
The implementation at https://github.com/slackhq/nebula/pull/74 is a more fleshed out version of this, plus dns bind address configuration, also supporting the nebula-internal IP for binding, to prevent exposing it to the public interfaces.

rawdigits:
Agree, going to close this one out as #74 supercedes it. Thank you for the PR @apognu !

PR #23, opened 2019-11-22
Edit README.md

sclaire-1:
Made the first sentence of the second paragraph in the "What is Nebula?" section more concise by removing unnecessary words


rawdigits:
Thanks for the suggestion. The docs are going to be largely overhauled, so we will probably avoid making small changes for now, since big changes are really needed and will substantially improve all of this.

PR #28, opened 2019-11-22
Better error message when no config files found

### LINKED ISSUE #26: Imprecise error messages
Running `nebula -config /this/doesnt/exist` or `nebula -config /no/read/permissions` (nonexistent path and unreadable file, respectively) results in
```
FATA[0000] no pki.ca path or PEM data provided          
```

While generally correct, i suppose, this should maybe point out that the config file could not be found or read.

nbrownus:
Resolves #26

### DIFF
--- a/config.go
+++ b/config.go
@@ -40,6 +40,10 @@ func (c *Config) Load(path string) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


rawdigits:
👍

Issue #24, opened 2019-11-22
macOS Release Binaries Aren't Signed


obelisk:
macOS binaries should be signed to prevent tampering and eventually to notify gatekeeper that it is allowed to run.

rawdigits:
Now to learn about signing in OSX.... 
Step 1: find a mac

Yafsec:
For people having problems running the nebula binary on OSX for now:

-Select the nebula binary from within finder
-Press [control] and click the binary
-select open
-accept the warning and run the binary once

After this you can run it from the terminal without errors

RoelWKramer:
Since both binaries are command line utils, you can also remove them from quarantine:

`xattr -d com.apple.quarantine ~/bin/nebula-cert`

`xattr -d com.apple.quarantine ~/bin/nebula`

uluckas:
@RoelWKramer thanks, works like a charm but disables a security layer.

Therefor executable signing is still an important feature request.

Tycho-S:
This is especially important in corporate environments where gatekeeper (only run signed apps) is mandatory. It would really be great if they could be codesigned.

Slack's Mac app is signed properly (albeit using the tinyspeck.com signature, not slack's own) so the knowledge and keys/Apple subscription are all there :)

PR #29, opened 2019-11-22
Added basic macOS signing

obelisk:
In reference to issue #24 this adds a new make target for signing macOS binaries. I didn't integrate into release because it would seem weird to only have the macOS signing there and I don't have the moment right now to figure out the linux stuff and I don't know the Windows one.

But to my knowledge this is about what it should look like, the default TEAMID is the Slack one (pulled from the signed macOS Slack app) but can be overriden for enterprises deploying their own forks of nebula (or ones with unmerged master changes) with:

```
make TEAMID=<Some other TEAMID> sign-darwin
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 BUILD_NUMBER ?= dev+$(shell date -u '+%Y%m%d%H%M%S')
 GO111MODULE = on
+TEAMID = BQR82RBBHL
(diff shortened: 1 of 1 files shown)


directionless:
This probably needs to be reworked for notarization. But I might imagine there's some work there to support iOS

PR #39, opened 2019-11-23
*DO NOT MERGE* PSK Support

rawdigits:
** Work in progress, do not merge! **

Support user configured noise PSK mixin via an hkdf derived from a shared passphrase for all nodes. Currently only enforcing 8 character string for deriving this, but this may be considered insufficient, depending on how important we think this feature to be.

### DIFF
--- a/handshake_ix.go
+++ b/handshake_ix.go
@@ -75,14 +75,24 @@ func ixHandshakeStage1(f *Interface, addr *udpAddr, hostinfo *HostInfo, packet [
 	if h.RemoteIndex == 0 {
 		ci := f.newConnectionState(false, noise.HandshakeIX, f.PSK, 0)
 		// Mark packet 1 as seen so it doesn't show up as missed
-		ci.window.Update(1)
 
 		msg, _, _, err := ci.H.ReadMessage(nil, packet[HeaderLen:])
 		if err != nil {
 			l.WithError(err).WithField("udpAddr", addr).
 				WithField("handshake", m{"stage": 1, "style": "ix_psk0"}).Error("Failed to call noise.ReadMessage")
+			//return true
+			for _, key := range f.AltPSKs {
+				ci = f.newConnectionState(false, noise.HandshakeIX, key, 0)
+				// Mark packet 1 as seen so it doesn't show up as missed
+
+				msg, _, _, err = ci.H.ReadMessage(nil, packet[HeaderLen:])
+				if err == nil {
+					continue
+				}
+			}
 			return true
 		}
+		ci.window.Update(1)
 
 		hs := &NebulaHandshake{}
 		err = proto.Unmarshal(msg, hs)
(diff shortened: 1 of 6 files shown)


noise.go
```diff
@@ -58,3 +62,18 @@ func (s *NebulaCipherState) DecryptDanger(out, ad, ciphertext []byte, n uint64,
 		return []byte{}, nil
 	}
 }
+
+func sha256KdfFromString(secret string) ([]byte, error) {
+	if len(secret) < 8 {
```
blazindragon: nit: Constant or configuration instead of magic number?

noise.go
```diff
@@ -58,3 +62,18 @@ func (s *NebulaCipherState) DecryptDanger(out, ad, ciphertext []byte, n uint64,
...
+		err := ("PSK too short!")
+		return nil, fmt.Errorf("%s", err)
```
blazindragon:
```
var (
  // ErrPSKTooShort describes an error in which the PSK character length is shorter than the minimum required size of <?>
  ErrPSKTooShort = errors.New("PSK too short")
)
```

```
return nil, ErrPSKTooShort
```

Perhaps return an exported error so callers can type check via `errors.IsA()` or similar?

noise.go
```diff
@@ -58,3 +62,18 @@ func (s *NebulaCipherState) DecryptDanger(out, ad, ciphertext []byte, n uint64,
...
+	}
+	hmacKey := make([]byte, sha256.Size)
+	hash := sha256.New
+	hkdfer := hkdf.New(hash, []byte(secret), nil, nil)
+	n, err := io.ReadFull(hkdfer, hmacKey)
+	if n != len(hmacKey) || err != nil {
+		return nil, fmt.Errorf("%s", err)
```
blazindragon: fmt.Errorf does nothing but toss away the original type of the error, why not just `return nil, err` ? Or provide another message and wrap it with `%w` (see "Wrapping errors with %w" in https://blog.golang.org/go1.13-errors)

noise_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_sha256KdfFromString(t *testing.T) {
+	key := "tooshrt"
+	mac, err := sha256KdfFromString(key)
+	assert.Nil(t, mac)
+	assert.EqualError(t, err, "PSK too short!")
+
+	key = "goodkeyexceptthatitisterriblebecauseitiswordsbutwhatever"
```
blazindragon: 🤣 👍

handshake_ix.go
```diff
@@ -73,17 +73,28 @@ func ixHandshakeStage0(f *Interface, vpnIp uint32, hostinfo *HostInfo) {
...
-			return true
+			for _, key := range f.AltPSKs {
+				ci = f.newConnectionState(false, noise.HandshakeIX, key, 0)
+
+				msg, _, _, err = ci.H.ReadMessage(nil, packet[HeaderLen:])
+				if err != nil {
+					//l.WithError(err).WithField("udpAddr", addr).
+					//	WithField("handshake", m{"stage": 1, "style": "ix_psk0"}).Error("Failed to call noise.ReadMessage")
+				} else {
+					break
```
blazindragon: nit: Slap in a comment that explains this picks the first parseable (non-erorr) alternate PSK in the order provided?

PR #41, opened 2019-11-23
Add an error (non fatal) when a lighthouse host has no static entry

### LINKED ISSUE #40: Add a warning if anything in lighthouse hosts does not have a static_host_map entry.
Not all static_host_map entries are necessarily lighthouses, but all lighthouses _should_ be listed in the static_host_map table.

To avoid accidental misconfiguration, we should validate that every item in lighthouse: hosts has a corresponding static_host_map entry on startup and (warn?|fatal exit?) if that is not the case.

rawdigits:
fixes #40

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -53,16 +53,16 @@ func NewLightHouse(amLighthouse bool, myIp uint32, ips []string, interval int, n
 	return &h
 }
 
-func (lh *LightHouse) ValidateLHStaticEntries() (bool, error) {
+func (lh *LightHouse) ValidateLHStaticEntries() error {
 	for lhIP, _ := range lh.lighthouses {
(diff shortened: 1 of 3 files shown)


lighthouse.go
```diff
@@ -53,6 +53,18 @@ func NewLightHouse(amLighthouse bool, myIp uint32, ips []string, interval int, n
 	return &h
 }
 
+func (lh *LightHouse) ValidateLHStaticEntries() error {
+	for lhIP, _ := range lh.lighthouses {
+		for ip, _ := range lh.staticList {
+			if lhIP == ip {
+				continue
```
wadey: should be `break` instead of `continue`? needs to break out of the inner loop to continue the outer loop

PR #47, opened 2019-11-25
fix: allow macos nodes to ping themselves

### LINKED ISSUE #38: [macOS] Unable to ping its own IP
Lighthouse (192.168.100.1): a DigitalOcean droplet with public IP.
Laptop (192.168.100.2): MacbookAir 2018 Catalina.
Config: I use the example config (with lighthouse public IP)

I can ping from Laptop to Lighthouse, and from Lighthouse to Laptop. But on the Laptop, I cannot ping its own IP (192.168.100.2).

Screenshot 2019-11-23 at 11 50 21 PM

On another Linux laptop with IP (192.168.100.3), I can ping its own IP.

Is it the limitation of macOS `utun` or somehow I need to config my Mac?

logikaljay:
fixes #38 

There is probably a better way to achieve this, but it works..
1. Delete 192.168.100.2 route
2. Create a new 192.168.100.2/32 route with a gateway of 127.0.0.1

What doesn't work?

Because we are creating our own route - it isn't cleaned up when nebula exits.

### DIFF
--- a/tun_darwin.go
+++ b/tun_darwin.go
@@ -46,6 +46,13 @@ func (c *Tun) Activate() error {
 	if err = exec.Command("route", "-n", "add", "-net", c.Cidr.String(), "-interface", c.Device).Run(); err != nil {
 		return fmt.Errorf("failed to run 'route add': %s", err)
 	}
+	if err = exec.Command("route", "-n", "delete", c.Cidr.IP.String()).Run(); err != nil {
+		return fmt.Errorf("failed to run 'route delete': %s", err)
+	}
+	if err = exec.Command("route", "-n", "add", fmt.Sprintf("%s/32", c.Cidr.IP.String()), "127.0.0.1").Run(); err != nil {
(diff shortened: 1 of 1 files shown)


nfam:
@logikaljay Is it possible that we can do something as following
1. The gateway is configurable, default is 127.0.0.1.
2. Trap SIG_INT, SIG_TERM,.. and cleanup the route.

logikaljay:
Hey @nfam, 

I am all for making improvements - I think cleaning up the route on SIGINT, SIGTERM is a great idea.

However we should think of a way to programatically get the loopback device address as adding OS-specific arguments could become confusing for the users.
If there is no way to reliably get the loopback address, we can resort to adding an argument - however the term "gateway" might be a bit misleading for the users.

logikaljay:
This is most likely fixed in https://github.com/slackhq/nebula/releases/tag/v1.5.0

PR #48, opened 2019-11-25
feat: adding docker support

### LINKED ISSUE #25: Docker image
When someone has some time, a Docker image would be awesome :)

logikaljay:
fixes #25 

Use a docker build envrionment and a seperate runtime environment to keep the image size down.

Build environment image size: 359mb
Runtime environment image size: 26.7mb

How to build:
```
$ docker build -t nebula .
```

How to use:
1. Copy (or create) your ca.crt and ca.key file in the CWD
```
$ docker run -it --rm \
  -v $(pwd):/app \
  nebula \
  nebula-cert ca -name "Myorganization, Inc"
```

2. Generate a certificate for the container
```
$ docker run -it --rm \
  -v $(pwd):/app \
  nebula \
  nebula-cert -name "container1" -groups "docker"
```

3. Create your [config.yaml](https://github.com/slackhq/nebula/blob/master/examples/config.yaml) for the container1 node.

4. Create the container and join the nebula
```
$ docker run -it --rm \
  --cap-add=NET_ADMIN --device /dev/net/tun \
  -v $(pwd):/app \
  nebula \
  nebula -config config.yaml
```

5. confirm that the host machine cannot ping another node
```
$ ping 192.168.100.1
PING 192.168.100.1 (192.168.100.1): 56 data bytes
Request timeout for icmp_seq 0
Request timeout for icmp_seq 1
Request timeout for icmp_seq 2
^C
--- 192.168.100.1 ping statistics ---
4 packets transmitted, 0 packets received, 100.0% packet loss
```

6. confirm that your docker container can ping another node
```
$ docker exec -it 5a87 sh
/app # ping 192.168.100.1
PING 192.168.100.1 (192.168.100.1): 56 data bytes
64 bytes from 192.168.100.1: seq=0 ttl=64 time=38.749 ms
64 bytes from 192.168.100.1: seq=1 ttl=64 time=38.388 ms
^C
--- 192.168.100.1 ping statistics ---
2 packets transmitted, 2 packets received, 0% packet loss
round-trip min/avg/max = 38.388/38.568/38.749 ms
```

I don't like that I have to add NET_ADMIN capabilities to the docker container.
I don't like that I have to provide the /dev/net/tun device to the docker container.

Maybe someone with more experience can resolve these two limitations

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM golang:alpine as buildenv
+
+RUN mkdir -p /build/bin
+ADD . /build/
+WORKDIR /build
+RUN apk add make
+RUN make
+RUN make install
+
+FROM alpine:3.7
+COPY --from=buildenv /build/nebula /usr/bin/nebula
+COPY --from=buildenv /build/nebula-cert /usr/bin/nebula-cert
+WORKDIR /app
+CMD ["nebula"]


michabbb:
@logikaljay can you show us your dockerfile?

logikaljay:
@michabbb I don't have one, however if you wanted to just add nebula to your Dockerfile you could just do something like...

```
FROM node:latest

WORKDIR /usr/local/bin
RUN wget https://github.com/slackhq/nebula/releases/download/v1.0.0/nebula-linux-amd64.tar.gz
RUN tar zxf nebula-linux-amd64.tar.gz

WORKDIR /app
COPY package.json .
RUN npm i
COPY ./src /app/src/

ENTRYPOINT npm start
```

You would have to make sure you have `/dev/net/tun` available on the host, supply it to the container when starting it, along with any certificates that you require.
You would also have to provide the `--cap-add=NET_ADMIN` flag when starting the docker container.

michabbb:
thanks a lot 👍

toonsevrin:
For anyone having issues with running inside kubernetes. This should be executed when your container starts (so `entrypoint.sh`), combined with the NET_ADMIN capability you're set:
```bash
# Create the tun device path if required
if [ ! -d /dev/net ]; then mkdir /dev/net; fi
if [ ! -e /dev/net/tun ]; then  mknod /dev/net/tun c 10 200; fi
```

afritzler:
For the Kubernetes case: Not sure if I would put this into the `entrypoint.sh`. I would suggest using an `InitContainer` to create the corresponding devices in conjunction with setting the correct `PodSecurityContext`: 

```    
securityContext:
  capabilities:
    add: ["NET_ADMIN"]
```

toonsevrin:
@afritzler what's the advantage of init over an entrypoint? I think your nebula instance requires NET_ADMIN in either case, no?

PR #56, opened 2019-12-01
Add linux-386, linux-ppc64le targets

nilium:
This patch set adds support for the linux/386 and linux/ppc64le
(GOOS/GOARCH) targets. The 386 target can't currently be built due to
the use of the syscall package and just because there's no
udp_linux_386.go. To fix this, use of the syscall package for Linux has
been replaced with the golang.org/x/sys/unix package (which already had
some sparse use). The latter target, ppc64le, should just require the
udp_linux_ppc64le.go changes, since 386 appears to have the largest
gaps in the syscall package.

Given that the 64 and 32-bit pairs of targets have the same
PrepareRawMessages support, it might be better to push these to use
comment-based build tags instead of filename-based ones.

This patch set is primarily intended to support packaging on Void Linux
(because it supports i686 and ppc64le), but may be useful for smaller
systems where there's little need for 64-bit architectures (not sure
embedded applies here, but the term's been stretched in recent years).

This has been confirmed working on ppc64le hardware (thanks to @q66 for
testing on PPC hardware) as well as with an i686 kernel in a VM (by
connecting to an ARM lighthouse and communicating with a third amd64
machine).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ BUILD_NUMBER ?= dev+$(shell date -u '+%Y%m%d%H%M%S')
 GO111MODULE = on
 export GO111MODULE
 
-all: bin-linux bin-arm bin-arm6 bin-arm64 bin-darwin bin-windows bin-mips bin-mipsle bin-mips64 bin-mips64le
+all: bin-linux bin-linux-386 bin-linux-ppc64le bin-arm bin-arm6 bin-arm64 bin-darwin bin-windows bin-mips bin-mipsle bin-mips64 bin-mips64le
 
 bin:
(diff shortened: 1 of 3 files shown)


nilium:
Sorry for the delay on this, been a hectic week. Anyway, testing on a 386 kernel / VM worked, so I'm removing the do-not-merge.

nbrownus:
If you rebase this branch on master you should be able to drop the `udp_linux` files you've added.

nilium:
I’ll do that in a bit, only have my phone at the moment. Should help keep things a little cleaner.

nilium:
There we go.

nbrownus:
Thanks!

PR #58, opened 2019-12-02
add possibility to specify output directory for sign command

reddec:
Add possibility to specify output directory for sign command

Adds `--out <directory>` to `sign` command, where `<directory>` is a path to output directory. If directory not exists, it will be created.

Example:

```
$ ls
ca.crt  ca.key  nebula  nebula-cert
$ nebula-cert sign --out certs --duration 1h -name client1 -ip 1.1.1.1/32
$ ls
ca.crt  ca.key  certs  nebula  nebula-cert
$ ls certs
client1.crt  client1.key
```

P.S.

Also some typo fix in help

### DIFF
--- a/cmd/nebula-cert/sign.go
+++ b/cmd/nebula-cert/sign.go
@@ -8,11 +8,12 @@ import (
 	"io/ioutil"
 	"net"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
-	"golang.org/x/crypto/curve25519"
 	"github.com/slackhq/nebula/cert"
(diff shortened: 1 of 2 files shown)


rawdigits:
Thanks for the PR! We already have options for -out-crt [path] and -out-key [path], so we will probably  avoid adding this, since it overlaps with existing functionality an may add confusion.

reddec:
Sad( It let's you specify only one parameter instead of two. Plus it creates directory if needed and stores keys more organized. 
Additionally the feature keeps backward compatibility.
Feels like more convenient for common use.

Issue #57, opened 2019-12-02
Suggestion: Unify example config file extension 


zhechenzhu:
I deployed nebula on some of my machines after listening to a recent episode of Linux Unplugged. It worked really well. Thank you. 

One issue that I ran into while setting up systemd service on a linux machine is that, due to the yaml file extension mismatch between the provided example config file and the service file, the service refuse to start. The example config file is named "config.yaml" while the config file path in the systemd service file is "/etc/nebula/config.yml". Toke me a while to figure out.

I would suggest unifying the file extensions to save some head scratch for new users.

rawdigits:
Good suggestion, and thanks for noting this! We will probably change the example file name.

PR #60, opened 2019-12-02
rename file to .yml to prevent confusion with example start scripts

rawdigits:
what it says on the tin..


rawdigits:
addresses #57

PR #67, opened 2019-12-04
Make static_host_map work as documented :trollface:

LeeBrotherston:
The example config states that the static_host_map is expressed as such:

"{nebula ip}": ["{routable ip/dns name}:{routable port}"]

However if you use a dns name, net.ParseIP will return nil, and there is no
check for an error condition.  Therefore I have added a few lines of (probably
awful) code that checks for nil being returned and instead attempts to lookup
the provided string using dns.


rawdigits:
I think it works as documented. Your change would allow for:
`"{nebula hostname/ip}": ["{routable ip/dns name}:{routable port}"]`
instead of
`"{nebula ip}": ["{routable ip/dns name}:{routable port}"]`
but we definitely prefer to force the identifier in the map to be an IP, since that's the primary key/common denominator of a nebula identity.

I've tested existing nebula with a dns name as the value(s) and it works as it says on the tin.

PR #68, opened 2019-12-04
Added Dockerfile

afritzler:
Added a multi stage Dockerfile for `nebula` and `nebula-certs` binary.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+#############      builder       #############
+FROM golang:1.13.4 AS builder
+
+WORKDIR /go/src/github.com/slackhq/nebula
+
+COPY . .
+
+RUN make
+
+# #############      nebula        #############
+FROM alpine:3.10.3 AS nebula
+
+RUN apk add --update libc6-compat
+
+COPY --from=builder /go/src/github.com/slackhq/nebula/nebula /nebula
+COPY --from=builder /go/src/github.com/slackhq/nebula/nebula-cert /nebula-cert
+
+WORKDIR /
+
+ENTRYPOINT ["/nebula"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,21 @@
 #############      builder       #############
-FROM golang:1.14.4 AS builder
+FROM golang:1.16.0 AS builder
 
 WORKDIR /go/src/github.com/slackhq/nebula
 
 COPY . .
 
-RUN make
+RUN make bin
 
 # #############      nebula        #############
-FROM alpine:3.12.0 AS nebula
+FROM alpine:3.13.2 AS nebula
 
 RUN apk add --update libc6-compat
 
 COPY --from=builder /go/src/github.com/slackhq/nebula/nebula /nebula
 COPY --from=builder /go/src/github.com/slackhq/nebula/nebula-cert /nebula-cert
+COPY entrypoint.sh /entrypoint.sh
 
 WORKDIR /
 
-ENTRYPOINT ["/nebula"]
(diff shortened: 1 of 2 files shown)


pfandzelter:
Great idea! Why did you decide to use libc6-compat and an alpine image rather than using FROM scratch and compiling with static linking? Also, doesn't "make" compile all binaries while we would just need "make bin-linux"?

afritzler:
@pfandzelter since I want to run nebula in my Kubernetes cluster, I would like to have a proper OS in my container so I can `exec` into it. Regarding the `make` you are of course right. Let me fix that. At the end it doesn't matter since it is a multi stage build but for the sake of speeding things up that totally makes sense.

kojishoda:
Thanks
Nice to meet my entrepreneur group friend <http://twitter@KhoonSiak>


Andreas Fritzler <notifications@github.com> 於 2019年12月19日 週四 上午12:21寫道：

toonsevrin:
Seems like this is going stale. Could this be merged and added to CD?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN make
+
+# #############      nebula        #############
+FROM alpine:3.10.3 AS nebula
+
+RUN apk add --update libc6-compat
+
+COPY --from=builder /go/src/github.com/slackhq/nebula/nebula /nebula
+COPY --from=builder /go/src/github.com/slackhq/nebula/nebula-cert /nebula-certs
```
toonsevrin, line 16: You probably want to change this to
```dockerfile
 COPY --from=builder /go/src/github.com/slackhq/nebula/nebula-cert /nebula-cert
```

afritzler: Good point! Thank you.

toonsevrin:
Gj @afritzler!

There's one more thing that I also wrote in #48 : If you add this to your entrypoint.sh (and use that as the entrypoint), the tunnel device is setup automatically and you only need NET_ADMIN.
```bash
# Create the tun device path if required
if [ ! -d /dev/net ]; then mkdir /dev/net; fi
if [ ! -e /dev/net/tun ]; then  mknod /dev/net/tun c 10 200; fi
/nebula "$@"
```

ieugen:
would love to see this merged

afritzler:
I hope I will find some time soon to include a proper `entrypoint.sh` with the changes proposed by @toonsevrin .

ieugen:
It would be nice to see this through, It will close a few issues/pulls.

https://github.com/slackhq/nebula/pull/48
https://github.com/slackhq/nebula/issues/25

and this one.

afritzler:
Ok, I added the suggested `entrypoint.sh` to the Dockerfile now.

therealklanni:
Why hasn't this been reviewed after nearly a year?

JohnMaguire:
I'm closing this issue out in lieu of #736 which uses an up-to-date Golang runtime and includes docs in the README. We're still sorting out what kind of Docker support users are looking for in #25.

> There's one more thing that I also wrote in #48 : If you add this to your entrypoint.sh (and use that as the entrypoint), the tunnel device is setup automatically and you only need NET_ADMIN.
> 
> ```shell
> # Create the tun device path if required
> if [ ! -d /dev/net ]; then mkdir /dev/net; fi
> if [ ! -e /dev/net/tun ]; then  mknod /dev/net/tun c 10 200; fi
> /nebula "$@"
> ```

@toonsevrin A point of confusion here - my understanding is that on Linux, CAP_NET_ADMIN allows for the creation of tun devices. Can you help me understand what problem `entrypoint.sh` is solving? Thanks!

PR #74, opened 2019-12-08
Allow configuration of dns listener host/port

coderobe:
This allows configuration of the DNS listener host and port.

I had to move the DNS initialization to a later point in nebula startup to allow the node's nebula IP to be used as DNS bind host, as the interface has to be ready for the bind to succeed in that case. I don't think that should cause any side effects.

I've also added documentation of the config options to the example config. This is backwards-compatible, as it only adds config options. I opted to not move `serve_dns` into `lighthouse.dns.` for that reason.

Common usage scenarios would be changing the port in case another resolver is running on :53, or changing the host to localhost/<nebula node IP> should one not want to publish this on all interfaces.

I have tested this.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -27,6 +27,10 @@ lighthouse:
   # serve_dns optionally starts a dns listener that responds to various queries and can even be
   # delegated to for resolution
   #serve_dns: false
+  #dns:
+    # The DNS host defines the IP to bind the dns listener to. This also allows binding to the nebula node IP.
+    #host: 0.0.0.0
+    #port: 53
   # interval is the number of seconds between updates from this node to a lighthouse.
   # during updates, a node sends information about its current IP addresses to each node.
   interval: 60
--- a/main.go
+++ b/main.go
@@ -195,7 +195,6 @@ func Main(configPath string, configTest bool, buildVersion string) {
 		l.Warn("lighthouse.am_lighthouse enabled on node but upstream lighthouses exist in config")
 	}
 
-	serveDns := config.GetBool("lighthouse.serve_dns", false)
 	lightHouse := NewLightHouse(
 		amLighthouse,
 		ip2int(tunCidr.IP),
@@ -206,11 +205,6 @@ func Main(configPath string, configTest bool, buildVersion string) {
 		udpServer,
 		punchBack,
 	)
-
-	if amLighthouse && serveDns {
-		l.Debugln("Starting dns server")
-		go dnsMain(hostMap)
-	}
 
 	//TODO: Move all of this inside functions in lighthouse.go
 	for k, v := range config.GetMap("static_host_map", map[interface{}]interface{}{}) {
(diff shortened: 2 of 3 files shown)


rawdigits:
This looks good. I'll test on my side this week. serve_dns is sadly lacking in tests, so may see about adding some.

rawdigits:
@coderobe what do you think of adding this to the config options that can be HUP-able? I think it could be very useful.

coderobe:
Sure, I've pushed a commit that adds HUP-ability. This has been tested and verified to be working by me.

coderobe:
Also resolved the conflict with your current `master` branch.

examples/config.yml
```diff
@@ -27,6 +27,10 @@ lighthouse:
   # serve_dns optionally starts a dns listener that responds to various queries and can even be
   # delegated to for resolution
   #serve_dns: false
```
nbrownus: Now that we have a `lighthouse.dns` parent, moving this to `lighthouse.dns.enabled` would help keep things clean. We'd need to add a deprecation notice and allow folks to use the old config path for the time being as well.

main.go
```diff
@@ -313,6 +308,12 @@ func Main(configPath string, configTest bool, buildVersion string) {
 	attachCommands(ssh, hostMap, handshakeManager.pendingHostMap, lightHouse, ifce)
 	ifce.Run(config.GetInt("tun.routines", 1), udpQueues, buildVersion)
 
+	// Start DNS server last to allow using the nebula IP as lighthouse.dns.host
+	if amLighthouse && serveDns {
```
nbrownus: It would be great if the reload support also worked for the `lighthouse.serve_dns` and `lighthouse.dns.enabled` config options to turn on/off the dns server with HUP.

examples/config.yml
coderobe: I intentionally didn't do this to preserve compatibility, i'd rather not deprecate config options that hit a stable release. Are you sure you want me to do that?

main.go
coderobe: Good point. I'll try to get that in today.

examples/config.yml
nbrownus: I'm fine not having you write the support for it, but I think we should be moving to keep config as clean as we can. Happy to write it up after this merge.

main.go
coderobe: Hmm, this is actually tricky because dns.Server.isStarted is not exported in miekg/dns, so we cannot easily figure out whether a DNS server is running without keeping state about that within nebula as well.

Without this we would have to restart the DNS server on every HUP, even when no config option has changed, simply because there's no way to figure out whether it's already running or needs to be started.

nbrownus: hmm, not the end of the world. I think the next move on the dns server is to get it out of leaking itself onto the global scope like it does. Would then lend itself to keeping that state.

nbrownus:
Thanks!

PR #82, opened 2019-12-11
Use Github Actions for Linux, Windows and macOS

wadey:
This allows us to run automated builds and tests for the three OSes.


wadey:
The tests will fail until #81 is merged

PR #88, opened 2019-12-12
Add gofmt Github action

wadey:
This runs gofmt for all pull requests, to make sure no files need to be reformatted.

Also, run gofmt on all current files so it will pass.

When it fails, it gives a diff of what needs to change. See here for an example of how a failure looks: https://github.com/slackhq/nebula/runs/345051948

### DIFF
--- a/.github/workflows/gofmt.yml
+++ b/.github/workflows/gofmt.yml
@@ -1,0 +1,29 @@
+name: gofmt
+on:
+  push:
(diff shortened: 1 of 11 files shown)


wadey:
> Gonna be gnarly the first time around :|

It's actually not bad, this PR fixes all of the current issues. Only import ordering issues.

Issue #90, opened 2019-12-12
error in README.md documentation - node names


bmullan:
In the README.md section titled:
  
**4. Nebula host keys and certificates generated from that certificate authority**

it names 3 nodes lighthouse1, host1, host3 but in the commands following it uses node names of lighthouse1, laptop, server1 and host3...

> This assumes **you have three nodes, named lighthouse1, host1, host3**. You can name the nodes any way you'd like, including FQDN. You'll also need to choose IP addresses and the associated subnet. In this example, we are creating a nebula network that will use 192.168.100.x/24 as its network range. This example also demonstrates nebula groups, which can later be used to define traffic rules in a nebula network.
> 
> ./nebula-cert sign -name "**lighthouse1**" -ip "192.168.100.1/24"
> ./nebula-cert sign -name "**laptop**" -ip "192.168.100.2/24" -groups "laptop,home,ssh"
> ./nebula-cert sign -name "**server1**" -ip "192.168.100.9/24" -groups "servers"
> ./nebula-cert sign -name "**host3**" -ip "192.168.100.9/24"

PR #94, opened 2019-12-13
Fix for Issue #90

infinytum:
Fixed the wrong node names mentioned


infinytum:
How do I sign your CLA ? I cant seem to find the button...

PR #114, opened 2019-12-19
Move tun mtu failing to an error log instead of fatal

### LINKED ISSUE #112: Restricted linux container permission problem
On some restricted linux containers you can not set the TX queue length for tap devices. I had to comment out the following:

`--- a/tun_linux.go
+++ b/tun_linux.go
@@ -193,7 +193,7 @@ func (c Tun) Activate() error {
        // Set the transmit queue length
        ifrq := ifreqQLEN{Name: devName, Value: int32(c.TXQueueLen)}
        if err = ioctl(fd, unix.SIOCSIFTXQLEN, uintptr(unsafe.Pointer(&ifrq))); err != nil {
-               return fmt.Errorf("failed to set tun tx queue length: %s", err)
+               //return fmt.Errorf("failed to set tun tx queue length: %s", err)
        }

        // Bring up the interface
%`

In general I would believe that this situation should not be an error but just a warning.

nbrownus:
As noted in the comment, the route table should have the MTU set as well. Failing to establish the route is a fatal.

closes #112

### DIFF
--- a/tun_linux.go
+++ b/tun_linux.go
@@ -187,7 +187,8 @@ func (c Tun) Activate() error {
 	// Set the MTU on the device
 	ifm := ifreqMTU{Name: devName, MTU: int32(c.MaxMTU)}
 	if err = ioctl(fd, unix.SIOCSIFMTU, uintptr(unsafe.Pointer(&ifm))); err != nil {
-		return fmt.Errorf("failed to set tun mtu: %s", err)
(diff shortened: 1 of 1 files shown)


jum:
This does not fix my problem, as my server does bail out at line 197, the result of SIOCSIFTXQLEN. The setting of the MTU was no problem for me.

PR #117, opened 2019-12-19
fix(nebula-cert): flag duration is optional, so reflect this is the cli help

zeisss:
`nebula-cert sign` defaults the `--duration` flag to 1 second before the CA expires, so it is not required.

See https://github.com/slackhq/nebula/blob/master/cmd/nebula-cert/sign.go#L102-L104

### DIFF
--- a/cmd/nebula-cert/sign.go
+++ b/cmd/nebula-cert/sign.go
@@ -36,7 +36,7 @@ func newSignFlags() *signFlags {
 	sf.caCertPath = sf.set.String("ca-crt", "ca.crt", "Optional: path to the signing CA cert")
 	sf.name = sf.set.String("name", "", "Required: name of the cert, usually a hostname")
 	sf.ip = sf.set.String("ip", "", "Required: ip and network in CIDR notation to assign the cert")
-	sf.duration = sf.set.Duration("duration", 0, "Required: how long the cert should be valid for. Valid time units are seconds: \"s\", minutes: \"m\", hours: \"h\"")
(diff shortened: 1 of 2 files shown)


nbrownus:
We should add a note about the computed default value in the description since we can't elude to it by actually setting a default value at the flag level.

zeisss:
@nbrownus any suggestions? I thought the same but nothing short + reasonable came to mind.

nbrownus:
"The default is 1 second before the signing cert expires" seems fine to me

zeisss:
Feels a bit long, but ok :)

```
21:04 $ ./nebula-cert sign -h
Usage of ./nebula-cert sign <flags>: create and sign a certificate
  -ca-crt string
    	Optional: path to the signing CA cert (default "ca.crt")
  -ca-key string
    	Optional: path to the signing CA key (default "ca.key")
  -duration duration
    	Optional: how long the cert should be valid for. The default is 1 second before the signing cert expires. Valid time units are seconds: "s", minutes: "m", hours: "h"
  -groups string
    	Optional: comma separated list of groups
  -in-pub string
    	Optional (if out-key not set): path to read a previously generated public key
  -ip string
    	Required: ip and network in CIDR notation to assign the cert
  -name string
    	Required: name of the cert, usually a hostname
  -out-crt string
    	Optional: path to write the certificate to
  -out-key string
    	Optional (if in-pub not set): path to write the private key to
  -subnets string
    	Optional: comma seperated list of subnet this cert can serve for
```

PR #138, opened 2019-12-28
alert in logs if nebula certificate is close to expiry

alanhlam:
* Check whether the in-memory nebula certificate is about to expire. 
* Write/alert in logs based on time left (`expiry_check.time_left`) on certificate and based on interval (`expiry_check.log_interval`) defined in nebula configs.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -14,8 +14,10 @@ pki:
   expiry_check:
     enabled: true
     # time remaining on the nebula certificate before expiry alerts appear in logs
+    # default is 5d
     time_left: 72h
     # how often the alerting will appear in logs
+    # default is 60m
     log_interval: 60m
 
 # The static host map defines a set of hosts with fixed IP addresses on the internet (or any network).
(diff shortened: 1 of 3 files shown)


examples/config.yml
```diff
@@ -10,6 +10,13 @@ pki:
   #blacklist is a list of certificate fingerprints that we will refuse to talk to
   #blacklist:
   #  - c99d4e650533b92061b09918e838a5a0a6aaee21eed1d12fd937682865936c72
+  #expiry_check looks at the in-memory cert and will begin alerting in logs if the host nebula certificate is close to expiry
```
nbrownus: Lets document the default values here for folks

expiry_check.go
```diff
@@ -0,0 +1,68 @@
...
+func CertTTLInvalid(cert *cert.NebulaCertificate, timeLeft time.Duration) bool {
+	return cert.Details.NotAfter.Before(time.Now().Add(timeLeft))
+}
+
+func (e *CertExpiryCheck) logExpiryCert() {
+	go func() {
+		for {
+			select {
+			case <-e.Ticker.C:
+				if CertTTLInvalid(e.Interface.certState.certificate, e.TimeLeft) {
```
nbrownus: We should log the time to expiry as well

PR #146, opened 2019-12-30
Add plist for MacOS launchctl to start nebula automatically

harpchad:


jrop:
It would also be great to add this to the Homebrew formula (formulae?) so that the following would work:

```
brew install nebula
brew services start nebula
```

As it stands currently, the following error is output:

```
brew services start nebula
Error: Formula `nebula` has not implemented #plist or installed a locatable .plist file
```

nbrownus:
For osx you should have the `-service` flag that should be able to handle these cases, I believe `-service install` is what you want.

NodeGuy:
I don't get an error message but it doesn't seem actually to start nebula either:

```shell
% sudo brew services start nebula
Password:
Warning: Taking root:admin ownership of some nebula paths:
  /usr/local/Cellar/nebula/1.4.0/bin
  /usr/local/Cellar/nebula/1.4.0/bin/nebula
  /usr/local/opt/nebula
  /usr/local/opt/nebula/bin
  /usr/local/var/homebrew/linked/nebula
This will require manual removal of these paths using `sudo rm` on
brew upgrade/reinstall/uninstall.
==> Successfully started `nebula` (label: homebrew.mxcl.nebula)

% ps -A | grep nebula
48374 ttys001    0:00.00 grep nebula
```

sanderlegit:
note, the config must have extension .yml (not .yaml)

> `sudo brew services start nebula` worked for me but I had to move my configuration into `/usr/local/etc/nebula` first.

PR #149, opened 2020-01-03
Lighthouse config reload on SIGHUP support

wadey:
This change moves the logic for creating the `LightHouse` structure out of main.go and into `NewLightHouseFromConfig`. It then makes the following config values reload on SIGHUP by registering a reload callback:

- `static_host_map`
- `lighthouse.hosts`
- `lighthouse.interval`
- `punch_back`

This makes it possible to update the static hosts or lighthouses and do a reload, without having to restart the Nebula process.

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -68,8 +68,19 @@ func NewLightHouseFromConfig(config *Config, myIp uint32, nebulaPort int, pc *ud
 		punchBack:    config.GetBool("punch_back", false),
 	}
 
+	err := h.loadStaticHosts(config)
+	if err != nil {
+		return nil, err
+	}
+
+	config.RegisterReloadCallback(h.reload)
+
+	return &h, nil
+}
+
+func (lh *LightHouse) loadStaticHosts(config *Config) error {
 	rawLighthouseHosts := config.GetStringSlice("lighthouse.hosts", []string{})
-	if h.amLighthouse && len(rawLighthouseHosts) != 0 {
+	if lh.amLighthouse && len(rawLighthouseHosts) != 0 {
 		l.Warn("lighthouse.am_lighthouse enabled on node but upstream lighthouses exist in config")
 	}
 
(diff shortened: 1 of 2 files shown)


lighthouse.go
```diff
@@ -53,6 +55,114 @@ func NewLightHouse(amLighthouse bool, myIp uint32, ips []uint32, interval int, n
...
+	}
+
+	return nil
+}
+
+func (lh *LightHouse) reload(config *Config) {
+	if config.GetBool("lighthouse.am_lighthouse", false) != lh.amLighthouse {
+		l.Warn("Changing lighthouse.am_lighthouse with SIGHUP not supported. Ignored.")
+	}
+	lh.interval = config.GetInt("lighthouse.interval", 10)
```
nbrownus: Something to note here, if the original config was borked with an insanely high interval a HUP won't help.

lighthouse.go
```diff
@@ -53,6 +55,114 @@ func NewLightHouse(amLighthouse bool, myIp uint32, ips []uint32, interval int, n
...
+			removedStatic = append(removedStatic, vpnIP)
+		}
+	}
+	for vpnIP := range lh.lighthouses {
+		if _, ok := newLh.lighthouses[vpnIP]; !ok {
+			removedLighthouses = append(removedLighthouses, vpnIP)
+		}
+	}
+
+	for _, vpnIP := range removedLighthouses {
```
nbrownus: This seems highly likely to cause a concurrent read/write access panic in both maps and our current RWMutex is entirely for the `addrMap`, seems we need at least one more now.

This also appears to only remove the diff set and not add any new hosts?

main.go
```diff
@@ -193,70 +190,14 @@ func Main(configPath string, configTest bool, buildVersion string) {
...
-				}
-				lightHouse.AddRemote(ip2int(vpnIp), NewUDPAddr(ip2int(ip), uint16(port)), true)
-			}
-		}
-	}
-
-	err = lightHouse.ValidateLHStaticEntries()
 	if err != nil {
-		l.WithError(err).Error("Lighthouse unreachable")
+		l.Fatalf("Failed to initialize lighthouse handler: %v", err)
```
nbrownus: I'd leave the `WithError` and keep this more helpful error message without the formatting. `Failed to initialize lighthouse`

jaysoncena:
Is it possible to add reloading of `pki.blacklist`?

wadey:
Closed in favor of #649

PR #151, opened 2020-01-05
patch for packaging

jeffw387:
This systemd service file could be directly used in Arch packages if it referenced nebula at /usr/bin instead of /usr/local/bin. Would there be a problem with defaulting to that? 

What's the use case for /usr/local/bin? Manual installation location?

### DIFF
--- a/examples/service_scripts/nebula.service
+++ b/examples/service_scripts/nebula.service
@@ -8,7 +8,7 @@ SyslogIdentifier=nebula
 StandardOutput=syslog
 StandardError=syslog
 ExecReload=/bin/kill -HUP $MAINPID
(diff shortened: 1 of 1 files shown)


nbrownus:
I'm a bit unfamiliar with the packaging standards for arch. Does it cause some downstream issues if the unit doesn't come from this repo?

jeffw387:
I'm not entirely sure to be honest.

nbrownus:
Found the bug in arch and left a comment. I'd rather not modify the current example unless others are depending on it. Happy to add a seperate file in `dist` or help the maintainer over there in whichever way they want to proceed.

jeffw387:
Sounds good to me, thanks for looking

PR #163, opened 2020-01-14
tun_darwin 

harpchad:
Two changes:

- Remove water and replace with sys call for tun setup
- Support named interfaces

**This needs testing**

I submitted the PR to get some more help reviewing and testing. I've tested on Catalina.

If anyone has pointers on sending route messages to an AF_ROUTE socket so we can replace the execs I'd love to see an example.


tun_darwin.go
```diff
@@ -2,66 +2,298 @@ package nebula
...
-	return err
+func (t *Tun) WriteRaw(b []byte) error {
+	var nn int
+
+	// add packet information header
+	var h [4]byte
+	h[0] = 0x00
+	h[1] = 0x00
+	h[2] = 0x00
+	h[3] = unix.AF_INET
```
nbrownus: Can you provide a link for the description of this behavior?

tun_darwin.go
```diff
@@ -2,66 +2,298 @@ package nebula
...
+	b = append(h[:], b[:]...)
```
nbrownus: This will trash the GC. Ideally we'd find a way to pass a slice with `h` prepended and avoid the copy. That said, darwin is not a likely maximum performance target.

tun_darwin.go
harpchad: It's doing the same thing IFF_NO_PI does on Linux (doesn't appear that flag exists in Darwin/BSD).  We have to strip the 4 byte protocol information header header on read, and add it on write.

tun_darwin.go
```diff
@@ -2,66 +2,298 @@ package nebula
...
+	}
+
+	// Set the MTU on the device
+	ifm := ifreqMTU{Name: devName, MTU: int32(t.DefaultMTU)}
+	if err = ioctl(fd, unix.SIOCSIFMTU, uintptr(unsafe.Pointer(&ifm))); err != nil {
+		return fmt.Errorf("Failed to set tun mtu: %v", err)
+	}
+
+	/*
+		// Set the transmit queue length
```
nbrownus: This seems to be managed by some ring buffer sizes and slot size. I need to read the code some more, but if we want to target performance we'd want to expose these options in config

These control options in specific
`UTUN_OPT_SLOT_SIZE`
`UTUN_OPT_NETIF_RING_SIZE`
`UTUN_OPT_TX_FSW_RING_SIZE`
`UTUN_OPT_RX_FSW_RING_SIZE`
`UTUN_OPT_MAX_PENDING_PACKETS`

nbrownus:
Whew there's a lot to learn here. I believe this will come in handy for establishing routes without `exec`.

`newroute` function https://opensource.apple.com/source/network_cmds/network_cmds-307.0.1/route.tproj/route.c.auto.html

tun_darwin.go
Dreamacro: what about https://golang.org/pkg/net/#Buffers ?

wadey:
i've updated this branch up to master, and also removed the `os/exec` usage (replaced with proper AF_ROUTE socket calls).

This is ready for more testing!

nbrownus (dismissed):

nbrownus (approved):

PR #166, opened 2020-01-16
add CHANGELOG.md

wadey:
This is the initial creation of a CHANGELOG. When we are ready to cut
a release we can change the version at the top and update the links at
the bottom.


rawdigits:
LGTM! Thanks for adding this @wadey !

PR #183, opened 2020-02-08
fix linting errors

tux21b:
This pull request currently contains 4 commits:

**1. simplify code**

This change removes unnecessary complicated constructs. For example                                                                                                                                                                                    ``if x == true {`` is changed to ``if x {``, and                                                                                                                                                                                                           ``if x == true { return true } else { return false }`` is changed                                                                                                                                                                                        to ``return x``.

No behavioral changes have been made in this commit. 

**2. 'safe' changes**

This change replaces ``timer.Tick`` with ``timer.NewTicker`` (and defers                                                                                                                                                                                       a call to ``timer.Stop`` to avoid theoretically leaking goroutines).                                                                                                                                                                                        The current code seems to be fine as is, because the loops are                                                                                                                                                                                         long running and can not be exited at the moment, but I think it is                                                                                                                                                                                    still good habit to write well-behaving code.

Additionally a small buffer is added to the ``os.Signal`` channel                                                                                                                                                                                          (according to the docs, those channels must be buffered because                                                                                                                                                                                        the signal handler can not block and can not do a synchronized                                                                                                                                                                                         send).

Neither of those changes should change the behavior in any way.   

**3. removed unused code**

Those variables are currently unused and its properly safe to                                                                                                                                                                                          remove them. They might indicate some other bugs though, so **please                                                                                                                                                                                     review**. 

**4. improve test error handling**

This change adds lots of error checks to the existing test cases.

I have also rewritten all usages of ``assert.Nil(err)`` to                                                                                                                                                                                               ``assert.NoError(err)``. The later gives a clear indent what the assertion is about (especially if its written like                                                                                                                                                                                           ``assert.Nil(createCA())``) and the error messages in the case of                                                                                                                                                                                        an failure are much clearer ("unexpected error").

No behavoral changes are introduced. This change only adds additional checks to the existing test cases.

---

There is also a 5th commit which is currently not part of this pull request. It adds missing error handling to the remaining program. Since those changes will affect the behavior, those should be reviewed more carefully. I will prepare another pull request for them in the future.

### DIFF
--- a/bits.go
+++ b/bits.go
@@ -18,7 +18,7 @@ type Bits struct {
 func NewBits(bits uint64) *Bits {
 	return &Bits{
 		length:             bits,
-		bits:               make([]bool, bits, bits),
+		bits:               make([]bool, bits),
 		current:            0,
 		lostCounter:        metrics.GetOrRegisterCounter("network.packets.lost", nil),
 		dupeCounter:        metrics.GetOrRegisterCounter("network.packets.duplicate", nil),
@@ -28,7 +28,7 @@ func NewBits(bits uint64) *Bits {
 
 func (b *Bits) Check(i uint64) bool {
 	// If i is the next number, return true.
-	if i > b.current || (i == 0 && b.firstSeen == false && b.current < b.length) {
+	if i > b.current || (i == 0 && !b.firstSeen && b.current < b.length) {
 		return true
 	}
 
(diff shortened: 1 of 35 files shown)


wadey:
I agree with most of these changes, but need to read it over again.

Can you tell me what linter and settings you are using?

tux21b:
Sure, the issues where found using golangci-lint (https://github.com/golangci/golangci-lint) in the current version with the default config (i.e. no config file). I use it for all my projects and can highly recommend it.

nbrownus:
We have drifted horribly from this PR, we can either have @tux21b apply the linting fixes to master or we can fold the linter into our workflow. My preference is to include the lint in a github action. @wadey?

PR #184, opened 2020-02-11
Support unsafe_routes on Windows

numkem:
Tested on Windows 2019 on AWS.


wadey:
Thank you, I tested this out and it works great.

Can you update the executable to be the full path:

    `C:\Windows\System32\route.exe`

Like we just merged with #191?

numkem:
This should do it!

PR #202, opened 2020-03-07
Upgrade to Protocol Buffers APIv2

wadey:
- https://blog.golang.org/a-new-go-api-for-protocol-buffers

I can connect to hosts running the old version with this build so it
looks like our signature validation is still compatible. No changes to our code required except for changing the import path.


nbrownus:
We have progressed beyond the versions listed in this `go.mod`, do we still have valuable work here?

wadey:
We updated our protobuf dependency in #369 and merged the Makefile changes as part of that as well, so we can close this now.

PR #210, opened 2020-03-27
add configurable punching delay because of race-condition-y conntracks

rawdigits:
Added a delay to punching via lighthouse signal to deal with race conditions in conntrack implementations, which seems to cause issues with random routers. This makes it much more likely that hole punching works.


rawdigits:
LGTM

PR #211, opened 2020-03-28
Build debian packages for nebula

gebi:
Hi,

As it seems now viable to use nebula with the last hole punching improvements i've polished our debian package building code :)

This can be used to build various distribution package formats (it uses `fpm` at it's core) so building rpm and packages for other distributions is easy.

This should not be thought as throwing something over the fence, i would also be willing to help you maintain that part in the future.
(i've had no time to play with github actions till now, but maybe that can be included directly)

### DIFF
--- a/contrib/deb/Dockerfile
+++ b/contrib/deb/Dockerfile
@@ -1,0 +1,21 @@
+FROM debian:buster
+MAINTAINER Michael Gebetsroither <michael@mgeb.org>
+
+# Create the builder, if you have fpm on your system just execute the script below on the host
+# Used packages:
+#   - make libffi-dev libtool (are needed because of https://github.com/ffi/ffi/issues/611)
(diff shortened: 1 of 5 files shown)


nbrownus:
We are not building/deploying packages in this repo due to complexity of the problem. Ideally folks would use the official packages built by their respective OS or maintain their own to suit their needs.

Here are a few that exist today:
https://tracker.debian.org/pkg/nebula
https://rpmfind.net/linux/rpm2html/search.php?query=nebula&submit=Search+...&system=&arch=
https://pkgs.alpinelinux.org/packages?name=nebula&branch=edge&repo=&arch=&maintainer=
https://archlinux.org/packages/extra/x86_64/nebula/
https://formulae.brew.sh/formula/nebula

Thanks!

PR #213, opened 2020-03-31
setup default config location

reddec:
Default location of configuration file lets distribute archive with configuration and nebula execution file to the target users (possibly not very technical educated) and let them join a network just by "doble-click" on executable. Otherwise consumers must be familiar with command line interface.

### DIFF
--- a/cmd/nebula/main.go
+++ b/cmd/nebula/main.go
@@ -16,7 +16,7 @@ import (
 var Build string
 
 func main() {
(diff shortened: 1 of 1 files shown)


tarrenj:
I like this! @reddec, would you mind submitting a PR to https://github.com/tarrenj/nebula/tree/dev so that the contribution is attributed to you within my fork please?

johnmaguire:
Thanks for the PR - I think for your specific use case, you can ship a small shell script that calls nebula with the correct file path.

PR #216, opened 2020-04-02
Add a systemd unit for arch and a wireshark dissector

nbrownus:
- First cut at a unified wireshark dissector
- Systemd unit for arch in repo

### DIFF
--- a/dist/arch/nebula.service
+++ b/dist/arch/nebula.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=nebula
+Wants=basic.target
+After=basic.target network.target
+
+[Service]
+SyslogIdentifier=nebula
+StandardOutput=syslog
+StandardError=syslog
+ExecReload=/bin/kill -HUP $MAINPID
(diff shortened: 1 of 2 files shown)


virtadpt:
It might be worth moving that .service file out of the Arch-specific directory.  It looks a lot like the one I use on Raspbian, Ubuntu, and Arch.

TheTyrius:
Sorry if this is obvious, but why are there two (seemingly) very similar service files in the source tree (under examples and now dist)?

nbrownus:
The one in examples uses a path to the binary that most distros won't be
using. This lead to an issue with arch specifically where the package did
not contain a unit file. Instead of breaking anyone relying on the one in
examples we chose to publish a new one in a very specific place to allow
arch to move forward.

willemw12:
I agree with @virtadpt: there is nothing distro-specific with that "Arch" systemd unit file (for creating an installable package). Except maybe syslog?

Issue #201, opened 2020-03-04
V1.1 nebula-linux-mips Fails with Illegal Instruction on OpenWrt (SOLVED)


fghorow:
I just attempted to execute the binary packaged at https://github.com/slackhq/nebula/releases/download/v1.1.0/nebula-linux-mips.tar.gz on an OpenWrt MIPS 24kc architecture machine.

As stated in the title, it immediately failed with an error message of "Illegal Instruction". The machine is confirmed via `lscpu` to be a Big-Endian configuration.
 
My deep suspicion, having previously seen this error message only upon trying to boot Debian on OpenWrt boxes, is that the nebula binary requires either hardware FPU instructions, or the kernel to be compiled with SoftFPU. The distribution OpenWrt kernels for MIPS 24kc are not compiled with SoftFPU.

If my suspicion is correct -- something which would take me a significant amount of time to verify -- the only workaround I know of requires setting up an OpenWrt cross-compilation chain and building the whole thing from source.  

Is there some way to avoid emitting FPU instructions in nebula binaries? That *might* be a quicker way to check my suspicion...

In any case, I'm submitting this issue so future OpenWrt users can see that this is a known issue.

fghorow:
OK. I can confirm that the issue was a lack of soft-float support.

Current workaround to get a working nebula executable (tested under Ubuntu 19.10 cross-compilation, with its go1.13.1 package installed and configured to be the default go compiler):

1.  clone the nebula source from this repo.

3.  set these environment variables:
`GOOS=linux
 GOARCH=mips
 GOMIPS=softfloat`
  That last variable is NOT set by default in the nebula build environment.

10. compile:
`go build -trimpath -ldflags "-s -w" -o ./nebula "./cmd/nebula"`
I am no expert on the go build system. Those `-s -v` flags produce a stripped binary. Perhaps there is also some other state created by previous compilation attempts that affected my results. YMMV. The resulting binary is in `./nebula`

11. (optional) compress the executable:
`upx nebula` 
I needed to do this in order to fit onto my router's (very) limited storage space. Again, YMMV.

Once those steps are done on my Ubuntu box, I copied the executable over to my router and executed it. Instead of complaining about an instruction error, it now presents me with the `usage` information, which I take as evidence of a successful "hello world".

Hope this helps somebody else!

wadey:
Thanks for giving details on how to make it work! I'll look into having the release process build a `mips-softfloat` binary.

[...]

PR #231, opened 2020-05-26
Build mips-softfloat

wadey:
This makes GOARM more generic and does GOMIPS in a similar way to
support mips-softfloat. We also set `-ldflags "-s -w"` for
mips-softfloat to give the best chance of the binary working on these
small devices.

Fixes: #201

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -279,6 +279,16 @@ jobs:
           asset_name: nebula-linux-mips64le.tar.gz
           asset_content_type: application/gzip
 
+      - name: Upload linux-mips-softfloat
+        uses: actions/upload-release-asset@v1.0.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create_release.outputs.upload_url }}
+          asset_path: ./linux-latest/nebula-linux-mips-softfloat.tar.gz
+          asset_name: nebula-linux-mips-softfloat.tar.gz
+          asset_content_type: application/gzip
+
       - name: Upload freebsd-amd64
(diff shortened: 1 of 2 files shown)


actronzs:
is it possible to make an ramips-softfloat version for download?
on my router with MediaTek MT7628AN ver:1 eco:2 the mips-softfloat does not work:
/usr/bin/nebula: line 1: syntax error: unexpected "(" 

tnx

wadey:
> is it possible to make an ramips-softfloat version for download

@actronzs It looks like that processor is little endian, have you tried the plain `mipsle` build first? If that doesn't work and you do need softfloat, I'll prepare a `mipsle-softfloat` build you can test as well.

actronzs:
@wadey 
with mipsle:

nebula -config /etc/nebula/config.yml
Illegal instruction

wadey:
@actronzs can you test the `mipsle-softfloat` build here? https://github.com/wadey/nebula/releases/tag/v1.3.0-mipsle-softfloat

actronzs:
@wadey thank you very much for creating a mipsle-softfloat build

sadly, the same result: Illegal instruction.

I have now tested all builds of nebula that are available for download, unfortunately all without success.

I was hoping that this build would work, now I can only pray that a ramips build will be available one day. or throw the router in the bin, which would be sad. because the price, size and performance of the TP-Link TL-WR902AC v3 would be perfect for us.

the rest of the routers we have (all slightly different models) work perfectly with the mips-softfloat build.

wadey:
It looks like you need to rebuild your kernel with FPU emulation support, some openwrt builds don't have it enabled by default.

See this thread which discusses the solution: 

https://github.com/golang/go/issues/18880

PR #244, opened 2020-06-24
Dont hold locks in ssh commands

psanford:
Previously we were holding hostmap and lighthouse read locks for the
duration of the commands `list-hostname` and `list-lighthouse-addrmap`
(including the network write time). For busy lighthouses this could
block host updates for a significant amount of time.

Now we make a copy of the data we're going to serialize and then
release the lock before marshaling it and sending it back down to the
ssh client.

### DIFF
--- a/ssh.go
+++ b/ssh.go
@@ -326,8 +326,7 @@ func attachCommands(ssh *sshd.SSHServer, hostMap *HostMap, pendingHostMap *HostM
 func sshListHostMap(hostMap *HostMap, a interface{}, w sshd.StringWriter) error {
 	fs, ok := a.(*sshListHostMapFlags)
 	if !ok {
-		//TODO: error
-		return nil
+		return fmt.Errorf("flags not of type *sshListHostMapFlags")
 	}
 
 	type hostInfo struct {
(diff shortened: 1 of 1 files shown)


ssh.go
```diff
@@ -378,49 +382,42 @@ func sshListHostMap(hostMap *HostMap, a interface{}, w sshd.StringWriter) error
...
+	}
+
+	type lighthouseInfo struct {
+		VpnIP net.IP   `json:"vpnIp"`
+		Addrs []string `json:"addrs"`
 	}
 
 	lightHouse.RLock()
-	defer lightHouse.RUnlock()
+	lighthouses := make([]lighthouseInfo, 0, len(lightHouse.addrMap))
```
nbrownus, line 394: Set the length and avoid `append`

ssh.go
```diff
@@ -378,49 +382,42 @@ func sshListHostMap(hostMap *HostMap, a interface{}, w sshd.StringWriter) error
...
+	for k, v := range lightHouse.addrMap {
+		addrs := make([]string, 0, len(v))
```
nbrownus, line 396: Same story as above

ssh.go
```diff
@@ -324,48 +326,50 @@ func attachCommands(ssh *sshd.SSHServer, hostMap *HostMap, pendingHostMap *HostM
...
+		RemoteIndex    uint32                  `json:"remoteIndex"`
+		RemoteAddrs    []*udpAddr              `json:"remoteAddrs"`
+		CachedPackets  int                     `json:"cachedPackets"`
+		Cert           *cert.NebulaCertificate `json:"cert"`
+		MessageCounter *uint64                 `json:"messageCounter"`
 	}
 
 	hostMap.RLock()
-	defer hostMap.RUnlock()
+	hosts := make(map[uint32]hostInfo, len(hostMap.Hosts))
```
nbrownus, line 343: I am unsure if we'd be willing to change this now, but we could turn this into an array and speed up/simplify this flow.

PR #246, opened 2020-06-26
trigger handshakes when lighthouse reply arrives

wadey:
Currently, we wait until the next timer tick to act on the lighthouse's
reply to our HostQuery. This means we can easily add hundreds of
milliseconds of unnecessary delay to the handshake. To fix this, we
can introduce a channel to trigger an outbound handshake without waiting
for the next timer tick.

A few samples of cold ping time between two hosts that require a
lighthouse lookup:

    before (v1.2.0):

    time=156 ms
    time=252 ms
    time=12.6 ms
    time=301 ms
    time=352 ms
    time=49.4 ms
    time=150 ms
    time=13.5 ms
    time=8.24 ms
    time=161 ms
    time=355 ms

    after:

    time=3.53 ms
    time=3.14 ms
    time=3.08 ms
    time=3.92 ms
    time=7.78 ms
    time=3.59 ms
    time=3.07 ms
    time=3.22 ms
    time=3.12 ms
    time=3.08 ms
    time=8.04 ms

I recommend reviewing this PR by looking at each commit individually, as
some refactoring was required that makes the diff a bit confusing when
combined together.

### DIFF
--- a/handshake_manager.go
+++ b/handshake_manager.go
@@ -200,6 +200,15 @@ func (c *HandshakeManager) AddVpnIP(vpnIP uint32) *HostInfo {
 	// We lock here and use an array to insert items to prevent locking the
 	// main receive thread for very long by waiting to add items to the pending map
 	c.OutboundHandshakeTimer.Add(vpnIP, c.config.tryInterval)
+
+	// If this is a static host, we don't need to wait for the HostQueryReply
+	// We can trigger the handshake right now
+	if _, ok := c.lightHouse.staticList[vpnIP]; ok {
+		select {
+		case c.trigger <- vpnIP:
+		default:
+		}
+	}
 	return hostinfo
 }
 
--- a/handshake_manager.go
+++ b/handshake_manager.go
@@ -16,21 +16,24 @@ const (
 	DefaultHandshakeTryInterval = time.Millisecond * 100
 	DefaultHandshakeRetries     = 20
 	// DefaultHandshakeWaitRotation is the number of handshake attempts to do before starting to use other ips addresses
-	DefaultHandshakeWaitRotation = 5
+	DefaultHandshakeWaitRotation  = 5
+	DefaultHandshakeTriggerBuffer = 64
 )
 
 var (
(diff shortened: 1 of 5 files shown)


lighthouse.go
```diff
@@ -358,6 +361,13 @@ func (lh *LightHouse) HandleRequest(rAddr *udpAddr, vpnIp uint32, p []byte, c *c
 			ans := NewUDPAddr(a.Ip, uint16(a.Port))
 			lh.AddRemote(n.Details.VpnIp, ans, false)
 		}
+		if lh.handshakeTrigger != nil {
```
nbrownus, line 364: Would be nice to fix tests to avoid this nil check in production

wadey: fixed in latest commit (this was actually unnecessary, a non-blocking send to a nil channel just skips the send).

I also added a test.

handshake_manager.go
```diff
@@ -80,69 +94,86 @@ func (c *HandshakeManager) NextOutboundHandshakeTimerTick(now time.Time, f EncWr
...
+			// We continue to query the lighthouse because hosts may
+			// come online during handshake retries. If the query
+			// succeeds (no error), add the lighthouse info to hostinfo
+			ips := c.lightHouse.QueryCache(vpnIP)
+			// If we have no responses yet, or only one IP (the host hadn't
+			// finished reporting its own IPs yet), then send another query to
+			// the LH.
+			if len(ips) <= 1 {
+				ips, err = c.lightHouse.Query(vpnIP, f)
+			}
```
forfuncsake, line 124: Is there a scenario where a host may have multiple IPs in the cache when we get here, but they're all stale (e.g. a mid-handshake roam)? If so, could this change cause the handshake to be eternally stuck in this state?

wadey: This code block only every runs when remote == nil, which only happens if there was no cache available in previous runs (once any ip is available, a remote will be selected). So this change here won't change current behavior here except to stop sending an extra lighthouse query that we won't use the response from.

PR #249, opened 2020-07-01
Fix up the tun for android

nbrownus:
This adds a new tun file for android, very similar to linux but we take care of `Activate` before `nebula` boots.

### DIFF
--- a/tun_android.go
+++ b/tun_android.go
@@ -1,0 +1,69 @@
+package nebula
+
+import (
+	"fmt"
+	"io"
+	"net"
+	"os"
+
+	"golang.org/x/sys/unix"
+)
+
+type Tun struct {
+	io.ReadWriteCloser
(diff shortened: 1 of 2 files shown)


tun_android.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		if n == 0 {
+			return io.ErrUnexpectedEOF
+		}
+	}
+}
+
+func (c Tun) Activate() error {
+	c.Device = "android"
```
wadey, line 67: Why set this here instead of on line 30?

nbrownus: To give the function something to do. I am very indifferent about where to place this but it felt inline with other `Activate`s where the device name is learned.

Technically, we can set the name on android but it's a fairly useless thing to do IMO, everything else is established beforehand.

PR #252, opened 2020-07-03
Make Interface.Inside an interface type

forfuncsake:
This commit updates the Interface.Inside type to be a new interface
type instead of a *Tun. This will allow for an inside interface
that does not use a tun device, such as a single-binary client that
can run without elevated privileges.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -2,6 +2,8 @@ package nebula
 
 import (
 	"errors"
+	"io"
+	"net"
 	"os"
 	"time"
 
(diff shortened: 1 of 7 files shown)


wadey:
Ah, correction, there is another one that matters:

- https://github.com/slackhq/nebula/blob/4756c9613dbe175de60d46e76b5c106a03381b02/interface.go#L147

But, I still think one spot changing from struct -> interface on the hot path should be okay.

PR #255, opened 2020-07-05
Added --watch-config flag

### LINKED ISSUE #254: Tracker: Updating the configuration automatically without sigHUP with --watch-config
If it's okay with you I'd like to add a new parameter `--watch-config` that uses [fsnotify](https://github.com/fsnotify/fsnotify) to watch the configuration file for changes.

The default will false (which means the config is only reloaded on a HUP signal). When set to true, the config will still reload on HUP but also when fsnotify detects a change.

I'll create a PR if this is okay to add. I really do not see the point in running sidecars and all that for this functionality.

toonsevrin:
The `--watch-config` flag sets up file watchers within the binary. When a config file or a pki file changes, it reloads the config.

The default is of course false (SIGHUP is then the only update mechanism).

**Things to be discussed:**
I simply call reloadConfig() when a file update is watched. I'm not sure what happens when the `pki.cert` is updated before the `pki.key`. The ideal behavior is to wait for them to update if they do not match, but after a little bit give it out and error out of the program. This is not implemented as for now but let's discuss it.

I have added a bunch of debug logging, feel free to update these.

**Features that may seem strange at first:**
- \[[Source](https://github.com/fsnotify/fsnotify/issues/254)\] Watches the directories instead of the files themselves. If you watch a file it will only trigger an event once. Note that if the file read fails for some reason, I will assume the file has changed, but still write the error to the log. I'm not sure if it's necessary to log this as an error, because this is something that is to be expected to happen sometimes.
- \[[Source](https://github.com/fsnotify/fsnotify/issues/324#issuecomment-601885573)\] Added checksum.go file: I cache the file checksums here, when a file is edited, it will create a lot of notifications: 3 on my machine. 

Resolves #254 
Edits by maintainers are allowed!

### DIFF
--- a/checksum.go
+++ b/checksum.go
@@ -1,0 +1,24 @@
+package nebula
+
+import (
+	"crypto/md5"
+	"encoding/hex"
+	"io"
+	"os"
+)
+
+// returns the checksum of the file at the filePath
+func md5sum(filePath string) (string, error) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
+
+	hash := md5.New()
+	if _, err := io.Copy(hash, file); err != nil {
+		return "", err
+	}
+
+	return hex.EncodeToString(hash.Sum(nil)), nil
+}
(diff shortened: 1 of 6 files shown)


nbrownus:
Since `nebula` supports pointing `-config` at a directory, modifying any single file can cause a reload, which may not be appropriate since we can not know if you are done modifying other files in the tree. The only way out is to ask that the thing modifying notify `nebula` when it is done.

Issue #172, opened 2020-01-21
Request: Create a "snap" for easy installation on various Linux platforms.


jrop:
Packaging nebula up into a Snap would allow users to install it via the following command on various linux distros:

`snap install --classic nebula`

An intro on packaging up a Go application into a "snap" can be found here:

https://snapcraft.io/first-snap#go

ArmoredBlood:
I'm looking into it. I have a product usecase that requires nebula be in a snap package. No guarantee on when it'll be done as it's my first snap, but I'm actively working on it

jrop:
Yeah, I wanted to look into it as well, but other time commitments take precedence.

jwallden:
I did some testing yesterday and with classic confinement there is absolutely no problems in getting this up and running. However, for getting the Nebula interface up and running you only need the network and network-control plugs so maybe its better to go the strict confinement route. That presents some challenges on where to store the created ca certs (and others) during creation but with the personal-files plug we can get around that to.

jwallden:
I have pushed a working snap to the snap store and are awaiting review and the grant to autoconnect network-control and the use of system-files. [Here is the link to the request over att Snapcraft](https://forum.snapcraft.io/t/request-for-auto-connect-of-network-control-and-the-use-of-plug-system-files-for-snap-nebula/15648)

I have published this snap under the name nebula. I'm of course happy to transfer ownership of this snap. Probably cleaner to have it published by SlackHQ. Anyhow, here is the link to the snap in its current state: https://github.com/jwallden/nebula-snap

jwallden:
Just want to let you know that with some minor changes the snap is now available in the Snapstore. Still haven't got approval for the autoconnect so have to do that manually for now.
I'd appreciate if someone could test it and let me know if any issues arise.

nbrownus:
Closing, please re-open #257 if there is still a desire to transfer the snap package.

Issue #101, opened 2019-12-17
Packaging


hugglesfox:
It would be great if Nebula provided deb and rpm packages to simplify the install and upgrade process.

SaigyoujiYuyuko233:
agree! :) but the installation is simple. It just is a simple runnable file.

coderobe:
These should ideally be provided by downstream packagers of the relevant ecosystems to prevent incompatibilities and ensure distribution-local packaging standards - [NixOS](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/networking/nebula/default.nix) and [Arch Linux](https://www.archlinux.org/packages/community/x86_64/nebula/) are already doing that.
Consider bringing it up to the package maintainer(s) of the relevant distributions

rawdigits:
I brought this in a comment on one of the arstechnica articles about nebula as well. We'd definitely prefer if other folks who enjoy using nebula might take up the task of packaging it for various distros/platforms, as @coderobe has done for Arch.

https://arstechnica.com/gadgets/2019/12/how-to-set-up-your-own-nebula-mesh-vpn-step-by-step/?comments=1&post=38425931

toonsevrin:
I'll write a snapcraft package, hopefully this can then be CDed with the [slack publisher](https://snapcraft.io/slack).

hlovdal:
I have created  [rpm packages for my own usage](https://github.com/hlovdal/nebula-overlay-networking) that I have used on all my Fedora/Centos machines for a couple of months now. With those it is only install package, edit configuration file and then start and enable service.

(Now my package is "cheating" in the build process by just pulling the release binaries from this repository, which is why I have not made any pull requests yet.)

But related to this, I think just `nebula` is a very poor package name since

1. It is not very descriptive.
2. There exists other existing packages already named [nebula](https://rpmfind.net/linux/RPM/fedora/34/x86_64/n/nebula-0.2.3-24.fc34.x86_64.html).

I chose to name the package `nebula-overlay-networking` to avoid these issues.

Now that there exists a [pull request for a Debian package](https://github.com/slackhq/nebula/pull/514) I think it would be good to have a discussion on package name to avoid unnecessary naming fragmentation (like for example [`kdeconnect-kde`, `kdeconnect-plasma`, `kdeconnect` or `kde-connect`](https://invent.kde.org/network/kdeconnect-kde#on-linux)).

Any maintainer opinions on package names?

hlovdal:
[...]

PR #257, opened 2020-07-08
Snap package

jwallden:
This is work to build Nebula as a snap package with reference to #172 
This snap is already published on https://snapcraft.io/nebula and if there is interest I'm happy to transfer ownership of that to the Nebula project


jwallden:
Oh, and the files are available in https://github.com/jwallden/nebula-snap/
I put the README in snap/local for this pull request.

toonsevrin:
Hooray! Also solves #101

toonsevrin:
One more thing: Perhaps it would be nice to add the following to the Readme:


## Install without any confinement
Run the following command to install Nebula without any confinement.
```shell
sudo snap install nebula --devmode
```

jwallden:
Good idea. I'll add it.

toonsevrin:
Dang, I keep bugging you here but is there any reason why you are using `SNAP_COMMON` instead of `SNAP_USER_COMMON`? The latter would allow you to run without sudo I think! :tada:

jwallden:
It's fine! Glad to have the conversation.

The SNAP_USER_COMMON is not restored on a snap refresh. This means, at least as I understand it, that e g the keys and config used in Nebula would get deleted when the package is updated.
See here: https://snapcraft.io/docs/environment-variables

However, when I look at it now it seems that SNAP_COMMON is also not backed up and restored. So actually I'm not sure why I chose SNAP_USER_COMMON. It is a while since I did this but from what I recall, back when I tested things SNAP_COMMON seemed lik a better option.

Now, In the big scheme of things I'm not really sure that it would matter with regards to running it with sudo. I think that since we're using network-control sudo would be required anyway.

toonsevrin:
@jwallden :)) what the documentation says about both SNAP_COMMON and SNAP_USER_COMMON is that the data is shared accross all versions of your snap. This means if you upgrade or downgrade, you actually retain your data, while with SNAP_DATA, it is lost. There you go!

And true, all though all the side-utilities (signing, keygen...) would work without sudo, which is somewhat nice.

jwallden:
Unclear if it's considered actively or not. I haven't got any reaction from Nebula mainainers whatsoever :)

nbrownus:
@jwallden we have been very hesitant to handle packaging in this repo, we did add docker support though so I could see adding snap to our release pipeline if you are still interested.

Closing for now, please re-open if you are still interested.

PR #258, opened 2020-07-10
Fix localOnly inversion; some idiomatic adjustments

forfuncsake:
Just a few things that stood out to be while I was sneaking a look at your branch.
There is one legit bug (`!localOnly`)

### DIFF
--- a/control.go
+++ b/control.go
@@ -16,7 +16,7 @@ type Control struct {
 }
 
 type ControlHostInfo struct {
-	VpnIP          net.IP                  `json:"vpnIp"`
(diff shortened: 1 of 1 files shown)


control.go
```diff
@@ -16,7 +16,7 @@ type Control struct {
 }
 
 type ControlHostInfo struct {
```
forfuncsake, line 18: The pointer type fields in this struct look risky to me, particularly `MessageCounter` and `CurentRemote` which appear to return the pointer to the nebula-internal values, therefore allowing modification.

nbrownus:
I'll be incorporating a few of these changes shortly, thanks for looking!

PR #260, opened 2020-07-27
Experiment: Modular functionality using go plugins

forfuncsake:
The commits in this PR are a first attempt at an experimental plugin framework along with a reference implementation of a basic configurable plugin to send/receive packets.

### DIFF
--- a/plugin/echo/main.go
+++ b/plugin/echo/main.go
@@ -1,0 +1,91 @@
+package main
+
+import (
+	"encoding/binary"
+	"fmt"
+	"net"
+	"time"
+)
+
+// P is the export required by the nebula plugin interface.
+var P = Echo{}
+
+// Echo provides the implementation of a Nebula Message Plugin interface that will echo
+// a payload to a remote nebula nodes log.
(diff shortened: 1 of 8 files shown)


plugin/echo/main.go
```diff
@@ -0,0 +1,91 @@
...
+	payload string
+	period  time.Duration
+}
+
+// Name returns the name of this plugin.
+func (e *Echo) Name() string {
+	return "echo"
+}
+
+// Configure accpets a config map that will be used to define the behaviour of the plugin.
```
forfuncsake, line 26:
```suggestion
// Configure accepts a config map that will be used to define the behaviour of the plugin.
```

forfuncsake:
For posterity, there were offline discussions about this experiment with the final resolution being that allowing arbitrary code execution via plugins created security risks that were going to be difficult to mitigate.
The preferred option for in-process custom payload handling was to enable nebula to be imported as a package with hooks available to send and receive those payloads.

That experiment is now being tracked in #310.

PR #261, opened 2020-07-28
add inside/outside hot path benchmarks

wadey:
These tests will be a bit relative, but can be used to compare the
results before and after a PR. We can also verify that no allocations
are introduced on the hot path, perhaps in a GitHub actions check.

    BenchmarkInsideHotPath/AESGCM-4         	  406822	      2879 ns/op	 521.06 MB/s	       1 B/op	       0 allocs/op
    BenchmarkInsideHotPath/ChaChaPoly-4     	  368092	      3278 ns/op	 457.61 MB/s	       1 B/op	       0 allocs/op
    BenchmarkOutsideHotPath/AESGCM-4        	 1723149	       718 ns/op	2090.07 MB/s	       0 B/op	       0 allocs/op
    BenchmarkOutsideHotPath/ChaChaPoly-4    	 1000000	      1108 ns/op	1354.02 MB/s	       0 B/op	       0 allocs/op

Because Interface.outside is not an interface, we have to actually
create a udp socket to write outgoing packets to in BenchmarkInside, so
the results look slower as they involve the kernel write path while
BenchmarkOutside is fully contained.

But we could use this benchmark to compare the results before and after
changing Interface.outside to an interface, as an example.

### DIFF
--- a/inside_test.go
+++ b/inside_test.go
@@ -1,0 +1,132 @@
+package nebula
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/flynn/noise"
+	"github.com/sirupsen/logrus"
+	"github.com/slackhq/nebula/cert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
+)
+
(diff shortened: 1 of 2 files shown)


nbrownus:
Closing this since #739 covers the same bits

PR #275, opened 2020-08-11
Align linux UDP performance optimizations with configuration

forfuncsake:
While attempting to run nebula on an older Synology NAS, it became
apparent that some of the performance optimizations effectively
block support for older kernels. The `recvmmsg` syscall was added in
linux kernel `2.6.33`, and the Synology ds212j (among other models)
is pinned to `2.6.32.12`.

Similarly, `SO_REUSEPORT` was added to the kernel in the 3.9 cycle.
While this option has been backported into _some_ older trees, it
is missing from the Synology kernel.

This commit allows nebula to be run on linux devices with older
kernels if the config options are set up with a single listener
and a UDP listen batch size of 1 (`listen.batch = 1`).

### DIFF
--- a/udp_linux.go
+++ b/udp_linux.go
@@ -161,37 +161,6 @@ func (u *udpConn) ListenOut(f *Interface) {
 	}
 }
 
-func (u *udpConn) Read(addr *udpAddr, b []byte) ([]byte, error) {
-	var rsa rawSockaddrAny
-	var rLen = unix.SizeofSockaddrAny
-
-	for {
-		n, _, err := unix.Syscall6(
-			unix.SYS_RECVFROM,
-			uintptr(u.sysFd),
-			uintptr(unsafe.Pointer(&b[0])),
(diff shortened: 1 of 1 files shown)


darren:
wow, this really helps! We spent days trying to figure out why nebula won't work on our old machines but to no avail.

Is there anyway to check whether `recvmmsg` is available?  if not, can the default value of config `batch: 64` be changed   value to 1?  because`tun.routines` already defaults to 1. So that user on old kernel will work if he uses the default configuration.

forfuncsake:
Changing the default value of `listen.batch` to 1 would likely hurt performance for a lot of the user base on upgrade if they are using the default value (currently `64`) in their deployments, so that's probably not the right path forward.

As far as checking for availability of the syscall, perhaps a change to the log output if the syscall fails to suggest config tuning makes sense... dependent on this change getting in first though.

PR #277, opened 2020-08-12
Use Go 1.15

wadey:
Update all CI checks and release process to use the latest patch version
of go1.15.

Note: We may want to hold off on merge for a week or so just to make sure there are no major issues in the new Go release.


rawdigits:
👍 lgtm

PR #279, opened 2020-08-12
More like a library

nbrownus:


udp_linux.go
```diff
@@ -87,8 +87,17 @@ func NewListener(ip string, port int, multi bool) (*udpConn, error) {
...
+	return nil
+}
+
+func (ua *udpAddr) Copy() udpAddr {
+	nu := udpAddr{
+		IP:   ua.IP,
+		Port: ua.Port,
+	}
+
+	return nu
```
wadey, line 100: You should be able to just do `return ua` here, since returning a struct instead of a pointer will effectively do the copy.

wadey (approved):
Overall, I think this looks good :+1: One small comment but we don't need to make the change.

We should update the documentation to say that we have removed support for `udp.routines` and `tun.routines`. I'm not sure if anyone was actively using those, or if perhaps we should try to update this PR to retain their support.

wadey (approved):
These changes look good to me :+1:

Issue #283, opened 2020-08-16
Multiple data races found


darren:
There're panics during handshake:

```go
DEBU[0011] Hostmap index deleted                         hostMap="map[indexNumber:1651690631 mapName:main mapTotalSize:1]"
DEBU[0011] Sending 0 stored packets                      certName=lighthouse-03 vpnIp=10.8.1.252
DEBU[0011] Generated index                               index=2257837831
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1513f67]

goroutine 13 [running]:
github.com/slackhq/nebula.ixHandshakeStage0(0xc000312160, 0xc00a0801fc, 0xc0002c03c0)
        github.com/slackhq/nebula/handshake_ix.go:40 +0x5a7
github.com/slackhq/nebula.(*Interface).getOrHandshake(0xc000312160, 0xa0801fc, 0x2329)
        github.com/slackhq/nebula/inside.go:106 +0x13b
github.com/slackhq/nebula.(*Interface).consumeInsidePacket(0xc000312160, 0xc0001ac000, 0x8c, 0x2329, 0xc0001ba000, 0xc0001ba010, 0xc, 0xc, 0xc0001ae500, 0x2329, ...)
        github.com/slackhq/nebula/inside.go:32 +0xf9
github.com/slackhq/nebula.(*Interface).listenIn(0xc000312160, 0x0)
        github.com/slackhq/nebula/interface.go:164 +0x13c
created by github.com/slackhq/nebula.(*Interface).Run
        github.com/slackhq/nebula/interface.go:125 +0x559
```

And enable race build found these data races:

```go
1. ==================
WARNING: DATA RACE
Read at 0x00c000348288 by goroutine 14:
  github.com/slackhq/nebula.(*HandshakeManager).handleOutbound()
      github.com/slackhq/nebula/handshake_manager.go:149 +0x3ed
  github.com/slackhq/nebula.(*HandshakeManager).NextOutboundHandshakeTimerTick()
      github.com/slackhq/nebula/handshake_manager.go:97 +0xb9
  github.com/slackhq/nebula.(*HandshakeManager).Run()
      github.com/slackhq/nebula/handshake_manager.go:83 +0x1ea

Previous write at 0x00c000348288 by goroutine 18:
  github.com/slackhq/nebula.ixHandshakeStage0()
[...]

PR #284, opened 2020-08-22
Fix a data race while sending metrics found in issue #283 case 3

darren:
```go
3. ==================
WARNING: DATA RACE
Write at 0x00c00030e020 by goroutine 17:
  sync/atomic.AddInt64()
      runtime/race_amd64.s:276 +0xb
  github.com/slackhq/nebula.(*Interface).sendNoMetrics()
      github.com/slackhq/nebula/inside.go:226 +0x9c
  github.com/slackhq/nebula.(*Interface).send()
      github.com/slackhq/nebula/inside.go:214 +0x149
  github.com/slackhq/nebula.(*Interface).readOutsidePackets()
      github.com/slackhq/nebula/outside.go:94 +0x1213
  github.com/slackhq/nebula.(*udpConn).ListenOut()
      github.com/slackhq/nebula/udp_generic.go:109 +0x3b5
  github.com/slackhq/nebula.(*Interface).listenOut()
      github.com/slackhq/nebula/interface.go:147 +0x15e

Previous read at 0x00c00030e020 by goroutine 18:
  github.com/slackhq/nebula.(*Interface).consumeInsidePacket()
      github.com/slackhq/nebula/inside.go:58 +0x892
  github.com/slackhq/nebula.(*Interface).listenIn()
      github.com/slackhq/nebula/interface.go:164 +0x178

```

### DIFF
--- a/inside.go
+++ b/inside.go
@@ -55,7 +55,8 @@ func (f *Interface) consumeInsidePacket(packet []byte, fwPacket *FirewallPacket,
 	dropReason := f.firewall.Drop(packet, *fwPacket, false, hostinfo, trustedCAs)
 	if dropReason == nil {
 		f.sendNoMetrics(message, 0, ci, hostinfo, hostinfo.remote, packet, nb, out)
-		if f.lightHouse != nil && *ci.messageCounter%5000 == 0 {
+		mc := atomic.LoadUint64(ci.messageCounter)
+		if f.lightHouse != nil && mc%5000 == 0 {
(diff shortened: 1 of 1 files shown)


wadey:
Thanks for doing this! I actually think we should tweak this a little bit to remove the extra load from the atomic int since this is on the hot path. `sendNoMetrics` currently already does an atomic increment on this variable, can you update `sendNoMetrics` to return the new value for the counter and then this part you changed can just use the returned value?

wadey (approved):
LGTM, thanks for making the suggested changes :+1:

PR #288, opened 2020-08-26
Add disable timestamp option

CzBiX:


examples/config.yml
```diff
@@ -156,6 +156,8 @@ logging:
   level: info
   # json or text formats currently available. Default is text
   format: text
+  # Disable timestamp logging. useful when output is redirected to logging system that already adds timestamps. Default is false
+  disable_timestamp: true
```
wadey, line 160:
```suggestion
  disable_timestamp: false
```

Our standard is to make the value defined in the example config be the default value, so that if someone just uses the example config verbatim, it will contain all default settings.

wadey (changes requested):
This looks great besides needing to fix the default value in the example config. Thanks!

wadey (approved):
Thanks :+1:

Issue #116, opened 2019-12-19
Use the wintun driver on Windows


rawdigits:
The existing driver we use comes from openvpn and has .. issues. We should use this nice simple thing instead:

https://www.wintun.net/

wadey:
Agreed! I thought about opening the same issue.

Issue #9, opened 2019-11-19
Fatal Error: TAP driver


voldemortensen:
```Activate failed: Failed to find the tap device in registry with specified ComponentId 'tap0901', TAP driver may be not installed```

The docs don't mention the need for a TAP driver. Which driver is recommended?

rawdigits:
Doh! Are you on Windows, by chance? If so, installing OpenVPN will add a tuntap driver we can use. I'll keep this issue open and update the docs. We also need a better solution for installation on windows.

voldemortensen:
Yes, this particular machine is a Windows machine (I meant to mention that). Thanks!

voldemortensen:
Circling back to confirm that installing OpenVPN did indeed solve the issue.

lixin9311:
you can find a standalone TUN/TAP driver for Windows here. (bottom of the page)

https://build.openvpn.net/downloads/releases/

good for guys don't want to install extra software

rawdigits:
Thanks to the pointer to a standalone driver! We'll use this in the documentation.

erayd:
The latest version of the OpenVPN (2.5.0) driver did not work out of the box for me on Windows 10 x64.

Changing the registry value `HKLM\SYSTEM\CurrentControlSet\Control\Class\{4d36e972-e325-11ce-bfc1-08002be10318}\0016\ComponentId` from `root\tap0901` to `tap0901` resolved the issue.

MKRhere:
Thanks @erayd; this helped.
I previously had tinc installed. So terminating tincd, modifying the registry value and running nebula worked fine.

stealthmode404:
> The latest version of the OpenVPN (2.5.0) driver did not work out of the box for me on Windows 10 x64.
> 
> Changing the registry value `HKLM\SYSTEM\CurrentControlSet\Control\Class\{4d36e972-e325-11ce-bfc1-08002be10318}\0016\ComponentId` from `root\tap0901` to `tap0901` resolved the issue.

It appears that the registry is not always at 0016. In mine, it was 0010. Basically, it's easier to identify it with the description. It was mentioned as "TAP-Windows Adapter V9" in the ProductName and this helped in identifying.

I changed the `ComponentId` from `tap_ovpnconnect` to `tap0901`
![Screenshot 2021-06-20 at 9 54 03 PM](https://user-images.githubusercontent.com/1282847/122681577-88206200-d212-11eb-80d6-8239c0665b98.png)

[...]

PR #289, opened 2020-08-26
Add Wintun support

CzBiX:
Users required to install wireguard to get wintun driver.


CzBiX:
Fix #116

wadey:
This is awesome, thank you so much!

To merge this, I think we need to make sure this won't break any existing users of Nebula that are using the OpenVPN tun driver. Could you update your patch to retain support for the old driver, but select the wintun driver if it is installed? I'm thinking perhaps we have `tun_water_windows.go` and `tun_wintun_windows.go` and we have a small stub in `tun_windows.go` to select which one to use based on the installed driver. If you have a better idea for how to go about this, feel free to try another route too. I just want to make sure we don't leave any existing users scratching their heads when they upgrade (at least have a deprecation period for folks to switch over).

In the future after this PR has been merged I'd like to set up a build process to create a `.msi` that includes the wintun driver as well, which will also hopefully solve this for new users.

Thanks again, let us know if you aren't interested in making those changes as well and we can take the reigns.

CzBiX:
@wadey Done, please review.

CzBiX:
I don't know why, Microsoft Security will reported virus.
I tinkered a bit, it won't complain now.

wadey:
Thanks for updating your PR, I just want to do a little more testing with this before we merge :+1:

wadey:
I was working on finishing up some testing for this, but they just announced today the release of Wintun 0.9 which is supposed to make distribution of the driver much easier. So I guess we should look at what it will take to upgrade to 0.9

https://lists.zx2c4.com/pipermail/wireguard/2020-November/005921.html

CzBiX:
The new version wintun is much easier to integrated.
But I found that wintun is licensed at GPL, which is incompatible with MIT.
Just send an email about license change to wireguard mail list.

ofer-dev:
I think https://github.com/WireGuard/wintun/commit/6af568d1e0edaa6809d99e7be5e13825c5e8bee6 solves license compatibility issues. wintun.h is now duel licensed under GPL and MIT.

CzBiX:
> I think [WireGuard/wintun@6af568d](https://github.com/WireGuard/wintun/commit/6af568d1e0edaa6809d99e7be5e13825c5e8bee6) solves license compatibility issues. wintun.h is now licensed under GPL and MIT.

Sorry, but it doesn't. The usage of wintun required linking to wintun library, which will make our code be GPL too.
In fact, I found that the TAP windows driver also be GPLv2 licensed. 

BTW, I haven't received any response from the author of wintun/wireguard yet.
It seems that this pr can be closed now.

wadey:
Did you send your email to Jason directly or to the mailing list? I was trying to find it on the mailing list and I didn't see it. Perhaps we are okay as long that file is MIT, but I need to brush up on GPL and DLLs.

Could you attach your previous email exchange about updating the header file license to this issue?

CzBiX:
I send the email to mailing list. and received "Your message to WireGuard awaits moderator approval". that's all.

ofer-dev:
Hey, the signed binary wintun.dll from https://www.wintun.net is under a different license then gpl. It is not an open source license but it allows to redistribute and link to wintun.dll as long as it is used only with the api defined in [wintun.h](https://github.com/WireGuard/wintun/blob/master/api/wintun.h).
Here is a copy of the [license](https://gist.github.com/ofer-dev/6ef9c46c5a6c03bf17cba206123e81ba).

wadey:
It looks like @ofer-dev is correct, and the .dll is distributed with a permissible license, one that we could bundle with an installer for Windows and not be concerned about licensing issues. The license is viewable in the wintun repo here: https://github.com/WireGuard/wintun/blob/master/prebuilt-binaries-license.txt

@CzBiX Would you like to take the effort of upgrading this PR for Wintun 0.10, or would you like to hand it over and we can work on it? Thanks again for your effort so far!

wadey:
Also, if you decide to update this PR, don't worry about how to bundle the .dll for distribution. We will work on a different PR to build an installer with the DLL during the release process.

CzBiX:
Since we will bundle the dll, I think we can remove `water tun` implementation now.

CzBiX:
@jamescorbett put the right `wintun.dll` into same folder with `nebula.exe`, then run nebula with administrator privileges.

gustavokei:
This seems to solve [#9](https://github.com/slackhq/nebula/issues/9#issuecomment-761974328) on Windows 10. I had the same problem.
It is also a much better solution than installing the entire OpenVPN package.

Got it working by building @CzBiX [fork](https://github.com/CzBiX/nebula/tree/wintun) and putting the `wintun.dll` into the same folder with `nebula.exe`
Hopefully this gets merged soon!

Awesome project, really appreciate your work!

paz:
any updates on this?

tskarman:
Just wanted to let you know that the CzBiX fork still appears to work well with the current Wintun version 0.13 (https://www.wintun.net/builds/wintun-0.13.zip)

Transfer speed on Windows is still markedly behind WireGuard in my situation but still ahead of the Nebula 1.4 Open-VPN-TAP-based version. But there's no point discussing that before there is a stable release.

wadey (approved):
I haven't been able to test this myself as I don't have a Windows system running currently, but the code looks good and I trust that it is running and working properly.

PR #293, opened 2020-08-30
Add support for specific Windows TAP adapter.

anedward01:
Based on rpooley's work, I added his modification to the most recent Nebula source. It creates a new entry in `config.yml` for `tun.interface_name` where the value is the name of the TAP adapter.

I also added a conditional for using a TAP adapter under ComponentId "root\0901" because adapters created by OpenVPN may label adapters as such instead of ComponentId "0901".

Creating an error message with newlines wasn't an option, so I created label "1:" and "2:" under `tun_windows.go` if there are no available adapters. There might be a better option available?

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -128,6 +128,8 @@ tun:
   disabled: false
   # Name of the device
   dev: nebula1
+  # Toggle this command and label after TUN device. This is a fix for Windows.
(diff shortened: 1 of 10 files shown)


nbrownus:
Closing as we've just moved to using `wintun` instead

PR #297, opened 2020-09-18
Add QR code support to `nebula-cert`

nbrownus:
Adds a `-qr` options to both `sign` and `ca` modes as well as a new `qr` mode to convert arbitrary files to qr encoded images.


cmd/nebula-cert/main.go
```diff
@@ -71,6 +71,8 @@ func main() {
 		err = printCert(args[1:], os.Stdout, os.Stderr)
 	case "verify":
 		err = verify(args[1:], os.Stdout, os.Stderr)
+	case "qr":
+		err = qr(args[1:], os.Stdout, os.Stdin)
```
forfuncsake, line 75: Just because you can, doesn't mean you should....

Arbitrary QR encoding is pretty far outside the scope of nebula-cert and definitely feels like a bolted on feature.

Perhaps adding the `-qr` flag on `print` mode fulfils the missing use case?

cmd/nebula-cert/ca.go
```diff
@@ -122,6 +125,10 @@ func ca(args []string, out io.Writer, errOut io.Writer) error {
 		return fmt.Errorf("refusing to overwrite existing CA key: %s", *cf.outKeyPath)
 	}
 
+	if *cf.qr && *cf.outCertPath == "ca.crt" {
+		*cf.outCertPath = "ca.png"
+	}
```
forfuncsake, line 130: The way this works is very specific and not very intuitive.
Combining these flags in the wrong way will lead to unexpected results.

I would expect use of a `-qr` flag to produce the png in addition to the text-based cert.

cmd/nebula-cert/main.go
nbrownus: Yes. That is way better. Thank you!

cmd/nebula-cert/ca.go
nbrownus: Yep. Totally agree again!

forfuncsake (dismissed):

forfuncsake (approved):

heyheyhello:
Hi did this get fully merged as per PR description or only a few of the features? I don't see the `qr` mode. The 1.4.0 release notes say it "is useful for configuring mobile clients" but I can only get a QR of the crt, not the key... so I can't actually configure a mobile client yet :upside_down_face: 

I ended up using https://github.com/fumiyas/qrc to transfer the crt and the key, but then I'm not using this PR :/

Would be great if Nebula supported an export of both crt and key since both are needed.

PR #299, opened 2020-09-22
use a buffered channel for signal.Notify

TimRots:
To avoid missing signals, the channel should be buffered and of the appropriate size. 
For a channel used for notification of just one signal value, a buffer of size 1 is sufficient, as per https://golang.org/pkg/os/signal/#Notify


wadey:
Thanks for your PR. Since this method only listens for one signal and then calls Stops and doesn't read from the channel again, we don't need a buffer on it.

We do need a buffer on this one though, if you want to change your PR to fix that one: https://github.com/slackhq/nebula/blob/v1.3.0/config.go#L117

PR #305, opened 2020-09-26
Tidy up some things

forfuncsake:
Just a couple of minor nits that I've noticed while reading through things recently.


nbrownus:
I believe a good deal of these have been addressed now, would love an update patch though.

PR #310, opened 2020-10-02
Experiment: Interface hooks

forfuncsake:
Following on from the plugin experiment in #260, this experiment provides an inverted implementation where nebula is used as a library/package and a wrapping command can use the library to hook into the "inside" interface.

Along with this PR, a demonstration app that uses this feature is being hosted at https://github.com/forfuncsake/nebula-hook.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - master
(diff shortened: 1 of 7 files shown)


nbrownus:
Closing for now, we can resurrect if/when there is a desire for this capability

PR #317, opened 2020-10-16
Added line to systemd config template to start Nebula before sshd

briankelly:
During shutdown, this will keep Nebula alive until after sshd is finished. This cleanly terminates ssh clients accessing a server over a Nebula tunnel. h/t @nbrownus 

I tested this on Ubuntu 20.04.1 LTS and Raspbian GNU/Linux 10 (buster).

### DIFF
--- a/examples/service_scripts/nebula.service
+++ b/examples/service_scripts/nebula.service
@@ -2,6 +2,7 @@
 Description=nebula
 Wants=basic.target
 After=basic.target network.target
+Before=sshd.service
 
 [Service]
 SyslogIdentifier=nebula


wadey (approved):
LGTM :+1:

since this is just a change to the example script, it should be safe to merge. I assume that the Before= still means that sshd.service won't be blocked from starting if Nebula fails to start?

briankelly:
> I assume that the Before= still means that sshd.service won't be blocked from starting if Nebula fails to start?

@wadey correct. I also tested this on raspbian by moving ca.crt and rebooting:

```
Oct 29 20:34:38 officepi nebula[428]: {"error":"unable to read pki.ca file /etc/nebula/ca.crt: open /etc/nebula/ca.crt: no such file or directory","level":"error","msg":"Failed to load ca from config","time":"2
Oct 29 20:34:38 officepi systemd[1]: nebula.service: Main process exited, code=exited, status=1/FAILURE
Oct 29 20:34:38 officepi systemd[1]: nebula.service: Failed with result 'exit-code'.
Oct 29 20:34:38 officepi sshd[477]: Server listening on 0.0.0.0 port 22.
Oct 29 20:34:38 officepi sshd[477]: Server listening on :: port 22.
Oct 29 20:34:38 officepi systemd[1]: Started OpenBSD Secure Shell server.
```

I could SSH to host on LAN without issue.

PR #320, opened 2020-10-22
Lighthouse handler optimizations

wadey:
We noticed that the number of memory allocations `LightHouse.HandleRequest` creates for each call can seriously impact performance for high traffic lighthouses. This PR introduces a benchmark in the first commit and then optimizes memory usage by creating a `LightHouseHandler` struct. This struct allows us to re-use memory between each lighthouse request (one instance per UDP listener go-routine).

Here is the comparison of before and after (`notfound` is that case that a query is made for a non-existent host. `found` is for the case that the host exists):

```
name                                old time/op    new time/op    delta
LighthouseHandleRequest/notfound-4     944ns ±11%     188ns ±15%   -80.07%
LighthouseHandleRequest/found-4       5.53µs ±40%    1.04µs ± 6%   -81.27%

name                                old alloc/op   new alloc/op   delta
LighthouseHandleRequest/notfound-4      264B ± 0%        0B       -100.00%
LighthouseHandleRequest/found-4       19.7kB ± 0%     0.1kB ± 0%   -99.68%

name                                old allocs/op  new allocs/op  delta
LighthouseHandleRequest/notfound-4      8.00 ± 0%      0.00       -100.00%
LighthouseHandleRequest/found-4         22.0 ± 0%       2.0 ± 0%   -90.91%
```

Thanks @forfuncsake and @psanford for the collaboration.

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -12,7 +12,7 @@ import (
 	"github.com/slackhq/nebula/cert"
 )
 
-var ErrQueryHostNotFound = errors.New("No host found")
+var ErrHostNotKnown = errors.New("host not known")
 
 type LightHouse struct {
 	sync.RWMutex //Because we concurrently read and write to our maps
@@ -116,7 +116,7 @@ func (lh *LightHouse) Query(ip uint32, f EncWriter) ([]udpAddr, error) {
 		return v, nil
 	}
 	lh.RUnlock()
-	return nil, ErrQueryHostNotFound
+	return nil, ErrHostNotKnown
 }
 
 // This is asynchronous so no reply should be expected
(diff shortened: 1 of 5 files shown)


lighthouse.go
```diff
@@ -281,9 +275,70 @@ func (lh *LightHouse) LhUpdateWorker(f EncWriter) {
...
+	details := lhh.meta.Details
+
+	details.Reset()
+	lhh.meta.Reset()
+	lhh.meta.Details = details
+
+	return lhh.meta
+}
+
+func (lhh *LightHouseHandler) resetIpAndPorts(n int) []*IpAndPort {
```
forfuncsake, line 315: After the latest cleanup commit, I think we should rename this to `resizeIpAndPorts` and remove the return value.

wadey: good point :+1:

adamrothman (approved):
Code looks good to me, but I lack context for the behavioral changes.

PR #323, opened 2020-10-29
i don't think this is used at all anymore

rawdigits:
prove me wrong


wadey (approved):
looks like you are correct :+1:

Issue #33, opened 2019-11-23
Question: NAT Setup


jatsrt:
I seem to be missing something important.  If I setup a mesh of hosts with all direct public IP addresses, it works fine.  However, if I have a network with a light house(public IP), then all nodes behind NAT, they will not connect to each other.  The lighthouse is able to communicate with all hosts, but hosts are not able to communicate with each other.  

Watching the logs I see connections trying to be made to both the NAT public, and the private IPs.

I have enabled punchy and punch back, but does not seem to help.

Hope it is something simple?

jatsrt:
Also, to note in this setup all nodes are behind different NATs on different networks.  Hub and spoke with the hub being the lighthouse and spokes going to hosts on different networks.

rawdigits:
My best guess (because I _just_ messed this up in a live demo), is that am_lighthouse may be set to "true" on the individual nodes.

Either way, can you post your lighthouse config and one of your node configs?

nfam:
Hi, I have the same issue. My lighthouse is on a DigitalOcean droplet with public IP. My MacBook and Linux Laptop at home are on the same network both connected to lighthouse. I can ping lighthouse from both laptop, but I cannot ping from one laptop to the other.

Lighthouse config
```yaml
pki:
  ca: /data/cert/nebula/ca.crt
  cert: /data/cert/nebula/lighthouse.crt
  key: /data/cert/nebula/lighthouse.key
static_host_map:
  "192.168.100.1": ["LIGHTHOUSE_PUBLIC_IP:4242"]
lighthouse:
  am_lighthouse: true
  interval: 60
  hosts:
listen:
  host: 0.0.0.0
  port: 4242
punchy: true
tun:
  dev: neb0
  drop_local_broadcast: false
  drop_multicast: false
  tx_queue: 500
  mtu: 1300
logging:
  level: info
  format: text
firewall:
  conntrack:
    tcp_timeout: 120h
    udp_timeout: 3m
    default_timeout: 10m
    max_connections: 100000
  outbound:
    - port: any
      proto: any
      host: any
  inbound:
    - port: any
      proto: icmp
      host: any
    - port: 443
      proto: tcp
      groups:
        - laptop
```

Macbook config
```yaml
pki:
  ca: /Volumes/code/cert/nebula/ca.crt
  cert: /Volumes/code/cert/nebula/mba.crt
  key: /Volumes/code/cert/nebula/mba.key
static_host_map:
[...]

PR #325, opened 2020-10-30
Delay opportunistic receive-side hole punching

rawdigits:
*WIP* DO NOT MERGE

It turns out that the opportunistic hole punching confused the hell out of linux conntrack. There ended up being additional races in the conntrack table and they were sometimes unresolvable. This is a temporary improvement, and expect some more tweaks in the near term.

...now for the good news: Hole punching works extremely well now! If you have the stomach for it, I'd encourage you to try this branch and see if things are oh-so-much-better.

I'll write up a much more detailed breakdown of the intention of the original feature and the issues it caused to share with the community soon.

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -395,7 +395,8 @@ func (hm *HostMap) Punchy(conn *udpConn) {
 	for {
 		for _, addr := range hm.PunchList() {
 			metricsTxPunchy.Inc(1)
-			conn.WriteTo([]byte{1}, addr)
+			_ = addr
+			//conn.WriteTo([]byte{1}, addr)
 		}
 		time.Sleep(time.Second * 30)
 	}
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -379,11 +379,11 @@ func (lh *LightHouse) HandleRequest(rAddr *udpAddr, vpnIp uint32, p []byte, c *c
 		}
 	case NebulaMeta_HostMovedNotification:
 	case NebulaMeta_HostPunchNotification:
-		/*
-			if !lh.IsLighthouseIP(vpnIp) {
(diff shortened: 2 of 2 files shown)


JohnMaguire:
I was experiencing issues similar to #33 and decided to try out this branch. Specifically:

A. DigitalOcean Lighthouse (Rain)
B. Local NAS on LAN (Acheron)
C. Local laptop on LAN (Macbook)

The LAN is using Google Wifi mesh routers, which some others in #33 are also using.

Prior to trying this patch, Lighthouse<->NAS communication worked, and Lighthouse<->Laptop communication worked. Attempting to communicate between Laptop<->NAS resulted in a debug log noting the other side an attempt at hole punching, followed by silence and no communication. (Neither ICMP pings nor SSH traffic succeeded.)

I should note that occasionally, communication between Laptop<->NAS would succeed, perhaps roughly until the Nebula service was restarted on one of the hosts. This led me to believe that a race may be the culprit.

I have only run this branch briefly - and I am only running it on the NAS as the CI seems to indicate it will not build on macOS, but it seems to resolve the issue. Laptop<->NAS communication is now working. I am able to ping the hosts from each other, and establish SSH connections.

JohnMaguire:
After more testing I believe the rest of my issues were caused by the numerous Docker bridge devices on this host. Once they were excluded from reporting to Lighthouse, connections seem to be established much faster and reliably.

The mobile app Pending tunnel view was very useful for figuring this out.

bsirang:
@rawdigits this PR sounds promising to me since I am having connectivity issues behind NATs...

Do you still plan on pushing this across the finish line? I notice there's a build failure here on this branch.

PR #329, opened 2020-11-03
add HostMap.RemoteIndexes

wadey:
This change adds an index based on HostInfo.remoteIndexId. This allows
us to use HostMap.QueryReverseIndex without having to loop over all
entries in the map (this can be a bottleneck under high traffic
lighthouses).

Without this patch, a high traffic lighthouse server receiving recv_error
packets and lots of handshakes, cpu pprof trace can look like this:

      flat  flat%   sum%        cum   cum%
    2000ms 32.26% 32.26%     3040ms 49.03%  github.com/slackhq/nebula.(*HostMap).QueryReverseIndex
     870ms 14.03% 46.29%     1060ms 17.10%  runtime.mapiternext

Which shows 50% of total cpu time is being spent in QueryReverseIndex.

This PR also combines a few commonly used together functions like `DeleteVpnIP` / `DeleteIndex` and `AddVpnIPHostInfo` / `AddIndexHostInfo` into one method so that we only have to grab the write lock once instead of for each function call.

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -244,8 +244,7 @@ func (n *connectionManager) HandleDeletionTick(now time.Time) {
 			if n.intf.lightHouse != nil {
 				n.intf.lightHouse.DeleteVpnIP(vpnIP)
 			}
-			n.hostMap.DeleteVpnIP(vpnIP)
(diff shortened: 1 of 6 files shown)


adamrothman (approved):
Lighthouse level-up! :shipit:

PR #331, opened 2020-11-12
don't mark handshake packets as "lost"

wadey:
Packet 1 is always a stage 1 handshake and packet 2 is always stage 2.
Normal packets don't start flowing until the message counter is 3 or
higher.

Currently we only receive either packet 1 or 2 depending on if
we are the initiator or responder for the handshake, so we end up
marking one of these as "lost". We should instead just always mark
packets 1 and 2 as being seen since a handshake cannot be successful
unless both have been transmitted and received.

### DIFF
--- a/handshake_ix.go
+++ b/handshake_ix.go
@@ -63,6 +63,10 @@ func ixHandshakeStage0(f *Interface, vpnIp uint32, hostinfo *HostInfo) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


forfuncsake (dismissed):
I've stared at words long enough to think I have a grasp on how this works now.
Looks reasonable to me.

nbrownus (approved):

PR #342, opened 2020-11-23
tun_disabled: reply to ICMP Echo Request

wadey:
This change allows a server running with `tun.disabled: true` (usually
a lighthouse) to still reply to ICMP Echo Request packets. This allows
you to "ping" the lighthouse Nebula IP as a quick check to make sure the
tunnel is up, even when running with tun.disabled.

This is still gated by allowing `icmp` packets in the inbound firewall
rules.

### DIFF
--- a/tun_disabled.go
+++ b/tun_disabled.go
@@ -1,26 +1,43 @@
 package nebula
 
 import (
+	"encoding/binary"
 	"fmt"
 	"io"
 	"net"
 	"strings"
 
+	"github.com/rcrowley/go-metrics"
+	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
 )
 
 type disabledTun struct {
-	block  chan struct{}
+	read   chan []byte
 	cidr   *net.IPNet
 	logger *log.Logger
+
+	// Track these metrics since we don't have the tun device to do it for us
+	tx metrics.Counter
+	rx metrics.Counter
 }
 
-func newDisabledTun(cidr *net.IPNet, l *log.Logger) *disabledTun {
-	return &disabledTun{
+func newDisabledTun(cidr *net.IPNet, queueLen int, metricsEnabled bool, l *log.Logger) *disabledTun {
+	tun := &disabledTun{
 		cidr:   cidr,
-		block:  make(chan struct{}),
+		read:   make(chan []byte, queueLen),
 		logger: l,
 	}
+
+	if metricsEnabled {
+		tun.tx = metrics.GetOrRegisterCounter("messages.tx.message", nil)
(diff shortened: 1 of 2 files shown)


tun_disabled.go
```diff
@@ -36,12 +53,69 @@ func (*disabledTun) DeviceName() string {
...
+	if !ok {
+		return 0, io.EOF
+	}
+
+	t.tx.Inc(1)
+	if l.Level >= logrus.DebugLevel {
+		t.logger.WithField("raw", prettyPacket(r)).Debugf("Write payload")
+	}
+
+	return copy(b, r), nil
```
forfuncsake, line 66: Just for some extra safety, consider:

```suggestion
	payloadLen = len(r)
	n := copy(b, r)
	if n != payloadLen {
		return payloadLen, fmt.Errorf("sent truncated payload %d/%d bytes", n, payloadLen)
	}
	
	return n, nil
```

tun_disabled.go
```diff
@@ -36,12 +53,69 @@ func (*disabledTun) DeviceName() string {
...
+
+	// We don't support fragmented packets
+	if b[7] != 0 || (b[6]&0x2F != 0) {
+		return false
+	}
+
+	buf := make([]byte, len(b))
+	copy(buf, b)
+
+	// Swap dest / src IPs and recalcualte checksum
```
forfuncsake, line 83:
```suggestion
	// Swap dest / src IPs and recalculate checksum
```

tun_disabled.go
```diff
@@ -36,12 +53,69 @@ func (*disabledTun) DeviceName() string {
...
+	copy(ipv4[12:16], b[16:20])
+	copy(ipv4[16:20], b[12:16])
+	ipv4[10] = 0
+	ipv4[11] = 0
+	binary.BigEndian.PutUint16(ipv4[10:], ipChecksum(ipv4))
+
+	// Change type to ICMP Echo Reply and recalculate checksum
+	icmp := buf[20:]
+	icmp[0] = 0
+	icmp[2] = 0
```
forfuncsake, line 94: `icmp[2]` is equivalent to `b[22]` which may lead to an out of bounds panic that we aren't protecting against.

wadey: good point, I'll update the check above to ensure the packet length is at least 28 bytes (minimum ICMP size)

tun_disabled.go
wadey: I'll add a check for this and also I'll check when we receive an ICMP Echo Request that it is <= mtu

adamrothman (approved):

forfuncsake (approved):

PR #345, opened 2020-11-24
Zero out allocations in connection manager

miranhpark:
not cleaning up makes others and myself a little nervous; adding on top of #340 - benchmarks locally similarly; thanks for the callout @forfuncsake

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -139,6 +139,12 @@ func (n *connectionManager) Start() {
 	go n.Run()
 }
 
+func memsetZero(a []byte) {
+	for i := range a {
+		a[i] = 0
+	}
+}
+
 func (n *connectionManager) Run() {
 	clockSource := time.Tick(500 * time.Millisecond)
 	p := []byte("")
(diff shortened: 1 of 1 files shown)


connection_manager.go
```diff
@@ -147,6 +153,9 @@ func (n *connectionManager) Run() {
 
 	for now := range clockSource {
 		n.HandleMonitorTick(now, p, nb, out)
+		memsetZero(p)
```
forfuncsake, line 156: `p` is never mutated, so this one is probably overkill (it's also a zero byte slice, so unless the compiler helps us out here, we would add the cost of a function call to do nothing).

forfuncsake (approved):

wadey:
`SendMessageToVpnIp` overwrites the contents of the nb and out slices when it runs, so they don't need to be reset to zero. We already rely on this behavior in many other places, I'm worried that adding zeroing in this spot may confuse people into thinking that the other uses in the code base aren't safe, when they are.

Instead I think we should just document these methods to say the contents of the slices will be overwritten.

PR #350, opened 2020-12-08
Including FreeBSD 12 rc.d service script

fearedbliss:
This PR implements a simple but clean FreeBSD 12 rc.d service script. It provides basic functionality for starting/stopping the service, including loading a custom nebula rc config (if any - defaults provided), log redirection, and passing the config parameter to the nebula process. There can probably be more things improved since this is my first FreeBSD rc.d script, but hopefully the more experienced FreeBSD Nebula users can further improve it.

### DIFF
--- a/examples/service_scripts/nebula
+++ b/examples/service_scripts/nebula
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+. /etc/rc.subr
+
+name="nebula"
+rcvar="nebula_enable"
+
+load_rc_config $name
+
+: ${nebula_log_file:="/var/log/nebula.log"}
+: ${config:="/usr/local/etc/nebula/config.yml"}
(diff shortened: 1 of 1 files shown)


nbrownus:
I am entirely unfamiliar with FreeBSD rc.d, is there a name we can give this that would be more descriptive in the examples directory?

If not we may need to move some things around, like `nebula.service` -> `systemd/nebula.service` and `nebula` -> `freebsd/nebula` to make it easier for folks

fearedbliss:
@nbrownus Hey Nate,

I think it's fine to either move it to a hierarchical structure for organization purposes (like you mentioned above), or you can just prefix the script with `freebsd` as well. I'm sure that if someone decides to package Nebula for FreeBSD's ports, they could easily rename the script as part of the install process.

Issue #280, opened 2020-08-13
make default route via nebula node (full subnet routing, 0.0.0.0/0, etc.)


darren:
Wireguard do someting magic while settting 
```
AllowedIPs = 0.0.0.0/0, ::/0
```
And all traffic will route by that wireguard peer.

But if set nebula's unsafe_route 0.0.0.0/0,   it will failed to start:
```
level=fatal msg="failed to set mtu 1300 on route 0.0.0.0/0; file exists"
```

hackxx:
In fact, the routing of wireguard is also done through system routing. It can also be realized by system routing！

ibehren1:
I would appreciate this option as well for certain members of the nebula.

nbrownus:
This is something we are interested in supported and I'll try to find time for it this year.

mrbluecoat:
ping

SherifNagy:
I ran a test, I might be totally wrong , however I think the issue with adding 0.0.0.0/0 via nebula IP, is that the handshake times out to reach the actual nebula server " at least this is what I think happens " so what I did was using https://www.procustodibus.com/blog/2021/03/wireguard-allowedips-calculator/ to exclude the nebula exit node IP , that gave me few routes to add, did that, reload my client and it works... I am sure there is a better way and maybe a correct way, but thought to put this here

nimblepenguin:
> I ran a test, I might be totally wrong , however I think the issue with adding 0.0.0.0/0 via nebula IP, is that the handshake times out to reach the actual nebula server " at least this is what I think happens " so what I did was using https://www.procustodibus.com/blog/2021/03/wireguard-allowedips-calculator/ to exclude the nebula exit node IP , that gave me few routes to add, did that, reload my client and it works... I am sure there is a better way and maybe a correct way, but thought to put this here

Did you add the routes under unsafe_routes? or did you add them on OS/System level (e.g. ip route add)?

SherifNagy:
> > I ran a test, I might be totally wrong , however I think the issue with adding 0.0.0.0/0 via nebula IP, is that the handshake times out to reach the actual nebula server " at least this is what I think happens " so what I did was using https://www.procustodibus.com/blog/2021/03/wireguard-allowedips-calculator/ to exclude the nebula exit node IP , that gave me few routes to add, did that, reload my client and it works... I am sure there is a better way and maybe a correct way, but thought to put this here
> 
[...]

Issue #145, opened 2019-12-30
Question: can a lighthouse be configured as a gateway to allow anonymous web browsing from a node?


russdirks:
I'm probably in way over my head on this, but am curious just the same.  If it's possible I would like to try to figure it out.  Probably need DNS on the lighthouse for sure.

rawdigits:
It is possible, but not currently well supported. You could set the lighthouse as an iptables masquerade host and route all traffic for 0.0.0.0/0 through it with linux routing. You'll also need 0.0.0.0/0 defined in `subnets` within the nebula certificate for the lighthouse.

Very much "probably works, but not supported" for the moment.

russdirks:
Thanks, I'll give that a try.

JohnMaguire:
Relevant: #280

PR #351, opened 2020-12-13
Add 0.0.0.0/0 unsafe route support

1dom:
Currently, setting an unsafe route of 0.0.0.0/0 results in Nebula creating a route to send everything down the tun device before the tunnel is established preventing handshake and breaking everything.

One way around this is to manually add a route directly to the external IP of the via node once Nebula is started to enable the handshake.

This PR attempts to make nebula check if the unsafe_route destination network contains any of the external IPs of the via node first, and then automagically add a route if so.

Note: I am not a Go programmer, so this may take a while to be usable for everyone else. Initial commit works for my setup though.

### DIFF
--- a/tun_common.go
+++ b/tun_common.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"strings"
 )
 
 const DEFAULT_MTU = 1300
@@ -12,6 +13,7 @@ type route struct {
 	mtu   int
 	route *net.IPNet
 	via   *net.IP
+	overrides []net.IPNet
 }
 
 func parseRoutes(config *Config, network *net.IPNet) ([]route, error) {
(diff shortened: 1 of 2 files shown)


JCBird1012:
This would fit a need I have, and bring Nebula one step closer to addressing #145.

fignew:
This seems like it would broaden the appeal of Nebula by allowing it to act as a typical VPN. Especially when you take into consideration the Android and iOS clients. (Unfortunately on iOS you can't have multiple VPNs so my traffic is in the open when using Nebula and public Wi-Fi)

monoidk:
Sidenote: For complex routing or those wishing a DIY routing setup on linux, implementing fwmark support #973 may be useful/cleaner, though probably overly complicated for just #145.

JohnMaguire:
Relevant: #280

johnmaguire:
This issue has been open for a few years and is still in a draft state, so I'm going to go ahead and close this out.

PR #357, opened 2021-01-03
adding example synology startup script and installation instructions

jdk:


examples/synology/Readme.md
```diff
@@ -0,0 +1,112 @@
...
+/usr/local/etc/rc.d/nebula.sh start
+/usr/local/etc/rc.d/nebula.sh stop
+/usr/local/etc/rc.d/nebula.sh restart
+```
+
+## Create a Startup Script
+
+Head over to the web portal of your synology and go to control panel. Scroll down to Task Scheduler.
+
+Click Create->Triggerd Tasks->User-defined scripts
```
forfuncsake, line 37:
```suggestion
Click Create->Triggered Tasks->User-defined scripts
```

examples/synology/Readme.md
```diff
@@ -0,0 +1,112 @@
+# Synology Installation
```
forfuncsake, line 1: For consistency within the repo, could you please adjust the case of this filename to `README.md`?

examples/synology/nebula.sh
```diff
@@ -0,0 +1,89 @@
...
+    ;;
+  status)
+    status
+    ;;
+  restart)
+    stop
+    start
+    ;;
+  *)
+    echo "Usage: $0 {start|stop|restart}"
```
forfuncsake, line 88: Is status omitted here for a reason?

examples/synology/nebula.sh
```diff
@@ -0,0 +1,89 @@
...
+
+SCRIPT="PATH_TO_CONFIG" # example /volume1/nebula/config.yml
+SUBNET="YOUR SUBNET IP" # example 10.1.0.0/24
+
+PIDFILE="PATH_TO_PID" # example - /volume1/nebula/nebula.pid
+LOGFILE="PATH_TO_LOG" # example - /volume1/nebula/nebula.log
+
+status() {
+  if [ -f $PIDFILE ]; then
+    echo 'Service running' >&2
```
forfuncsake, line 11: We should probably verify that this is not a stale pidfile via something like `kill -0 $(cat $PIDFILE)` here too.

examples/synology/Readme.md
```diff
@@ -0,0 +1,112 @@
...
+
+```
+/usr/local/etc/rc.d/nebula.sh restart &> /volume1/nebula/start.log
+```
+
+***NOTE*** - We are using Restart on purpose in case we did not cleanly shutdown. Our PID file we use to kill our service will still be running, and our start command will fail, so we purposely kill the script before re-running.
+
+## Create a Shutdown Script
+
+Click Create->Triggerd Tasks->User-defined scripts
```
forfuncsake, line 64:
```suggestion
Click Create->Triggered Tasks->User-defined scripts
```

examples/synology/Readme.md
```diff
@@ -0,0 +1,112 @@
...
+
+If the synology did not properly shutdown, the PID file could be present, and the nebula script will exit. You can always force remove the file, or force re-start nebula with the following command:
+
+```
+/usr/local/etc/rc.d/nebula.sh restart
+```
+
+### Route Not Present
+
+The script creates an additional route, check to see it's present:
```
forfuncsake, line 102: Would be helpful to document here that this is achieved by running the command `route`.

examples/synology/nebula.sh
```diff
@@ -0,0 +1,89 @@
...
+  if [ -f $PIDFILE ] && kill -0 $(cat $PIDFILE); then
+    echo 'Service already running' >&2
+    return 1
+  fi
+  printf 'Starting nebula service...' >&2
+  "$SCRIPT" &> "$LOGFILE" & echo $! > "$PIDFILE"
+  sleep 5
+
+  NEXT_WAIT_TIME=0
+  until [ $NEXT_WAIT_TIME -eq 120 ] || [ ! -z "$(ifconfig | grep nebula1)" ]; do
```
forfuncsake, line 26: `nebula1` here assumes that we're using the example config. The default value might be `tun<x>`, or may have been defined in the users own config file. It's probably worth making this one of the configurable values at the top of the script.

PR #367, opened 2021-01-19
allow multiple instances of daemon

waldner:
Use systemd's unit instance mechanism to allow running multiple nebula daemons with different configuration files.


nbrownus:
The file would need to be named `nebula@.service` and I believe having 2 `.service` files in the examples directory would be confusing for folks that aren't familiar with this systemd feature.

PR #369, opened 2021-01-27
IPv6 support for outside (udp)

nbrownus:
Enables the UDP side and the lighthouse code base to work with IPv6 addresses.

Current areas needing more thought
- Should we have a main config control to enable/disable IPv6 support? How would that interact with `lighthouse.remote_allow_list` and `lighthouse.local_allow_list`? A simple boolean would make all ipv6 invalid to the allow lists but it is duplicative.
- We should likely have a config option to prioritize v4 or v6 addresses, this is in addition to the preference we have on local over public ips
- We do not currently respect v6 scope and we likely should.
- Ideally we would not use deterministic and trackable v6 addresses, or at least default a config option to ignore them.

This is not yet performance tuned.


rawdigits (approved):
I approve of this product or service.

udp_linux.go
```diff
@@ -148,12 +110,24 @@ func (u *udpConn) LocalAddr() (*udpAddr, error) {
 
 	addr := &udpAddr{}
 	if rsa.Addr.Family == unix.AF_INET {
+		pp := (*unix.RawSockaddrInet4)(unsafe.Pointer(&rsa))
+		addr.Port = uint16(rsa.Addr.Data[0])<<8 + uint16(rsa.Addr.Data[1])
+		copy(addr.IP, pp.Addr[:])
+
+	} else if rsa.Addr.Family == unix.AF_INET6 {
+		//TODO: this cast sucks and we can do better
```
wadey, line 118: I think we could use `unix.Getsockname` in this method to get rid of these casts (although the unix library is just doing the same casts under the hood, but it would clean this function up).

rawdigits (approved):

rawdigits (approved):

rawdigits (approved):

wadey (approved):
We are allocating a bit extra in LightHouseHandler but we can come back in fix that in a follow up.

PR #370, opened 2021-02-02
Teardown tunnel automatically if peer's certificate expired

ton31337:
Handle this with periodic keepalive ticks. This is needed to avoid
hanging a connection even peer's certificate expired.

In case you use short-lived certificates (let's say 2 hours), current
behavior is a bit wrong, because the tunnel stays UP and RUNNING fine
unless restarted nebula service.


ton31337:
An optional feature on/off would be great as well to avoid breaking existing setups.

virtadpt:
Can we please also have a notice "Node X's certificate has expired, tearing down connection" in the logs?

ton31337:
Gonna update that, also adding a test for this case.

ton31337:
@virtadpt added more logging regarding disconnection.

virtadpt:
So, monitor output for "Invalid certificate status", and then suss out the value of `vpnIp` for the host that needs remediated.

Works for me.  I'll give it a try this weekend.

ton31337:
@virtadpt any chance you had tested it?

ton31337:
@rawdigits @nbrownus any chance to review this?

virtadpt:
@ton31337 Not yet.  I'm still working 90 hour weeks.

ton31337:
Still, no?

virtadpt:
I finally went off call yesterday.  I'll give it a try when I have a moment.

ton31337:
Any updates? :)

connection_manager.go
```diff
@@ -153,6 +154,26 @@ func (n *connectionManager) Run() {
 	}
 }
 
+// Check if peer's certificate is not expired or invalid.
+func (n *connectionManager) checkToDisconnect(now time.Time, hostinfo *HostInfo) (bool, *cert.NebulaCertificate) {
+	if !n.intf.disconnectInvalid {
+		return false, nil
+	}
+
+	if hostinfo == nil {
```
nbrownus, line 163: Since `hostinfo` will have a nil check before this function, we can remove this re-check.

connection_manager.go
```diff
@@ -153,6 +154,26 @@ func (n *connectionManager) Run() {
...
+		return false, nil
+	}
+
+	if remoteCert := hostinfo.GetCert(); remoteCert != nil {
+		valid, _ := remoteCert.Verify(now, n.intf.caPool)
```
nbrownus, line 168: Passing the error back to the caller would allow the `disconnect` log to differentiate why (cert expired, root expired, could not find ca, etc)

nbrownus (commented):
Made a few comments, using `connectionManager` for this is nice as it won't really punish the hot path and should give a fairly quick reaction to changes in CA trust/certificate expiry.

examples/config.yml
```diff
@@ -10,6 +10,8 @@ pki:
   #blocklist is a list of certificate fingerprints that we will refuse to talk to
   #blocklist:
   #  - c99d4e650533b92061b09918e838a5a0a6aaee21eed1d12fd937682865936c72
+  #disconnect_invalid is a toggle to force a client to be disconnected if the certificate is expired or invalid.
```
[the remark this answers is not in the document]
ton31337: Fixed.

PR #373, opened 2021-02-04
Check certificate banner during nebula-cert print

brad-defined:
Before this change, if you were to accidentally try to print a file that is not a certificate, you would get a nebulous error message:
`Error: error while unmarshaling cert: proto: can't skip unknown wire type X`

Added the banner check, which results in the clearer error:
`Error: error while unmarshaling cert: bytes did not contain a proper nebula certificate banner`

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -127,6 +127,9 @@ func UnmarshalNebulaCertificateFromPEM(b []byte) (*NebulaCertificate, []byte, er
 	if p == nil {
 		return nil, r, fmt.Errorf("input did not contain a valid PEM encoded block")
 	}
+	if p.Type != CertBanner {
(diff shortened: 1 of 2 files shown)


nbrownus:
This is great, can you also add a test to cover it? We've tried to keep the `cert` package test coverage rather high.

brad-defined:
I looked for existing tests to ~cut & paste from~ inspire me, but didn't find any. So I included tests for the other Unmarshal* funcs in cert.go.

nbrownus (approved):
Thanks for increasing the test coverage!

wadey (approved):
Looks good, my only concern was if it would break comments in the certificate file and it looks like you have tests for that. :+1:

Issue #314, opened 2020-10-11
Unable To Start Prometheus Stats Endpoint On Nebula IP Unless `0.0.0.0` Address Is Used


bonedaddy:
Attempting to start a nebula node that has enabled prometheus statistics to accept connections on the nebula node IP causes startup to fail with the error message `bind: cannot assign requested address`.  If you want to be able to have the prometheus endpoint accept connections on the nebula IP, then you need to set the listening address to `0.0.0.0` which is not desirable since this would then start a listener on all attached interfaces.

Example config:

```yaml
  type: prometheus
  listen: 0.0.0.0:8081
  path: /metrics
  namespace: prometheusns
  subsystem: nebula
  interval: 10s
```

Issue #243, opened 2020-06-20
administrative sshd refused connection from nebula vpn network


ghostoy:
I enabled sshd on lighthouse by listening on nebula internal network ip: 172.31.1.1:2222 and added incoming firewall rule for vpn ips. **On the first boot of nebula service**, the connection coming from vpn ips, like 172.31.1.2, still got refused by the firewall. The workaround is just to **signal the nebula process to reload the configuration file without changing anything**. And everything works just fine.

ghostoy:
Config for the lighthouse:
```yaml
# ... parts before sshd
# sshd can expose informational and administrative functions via ssh this is a
sshd:
  # Toggles the feature
  enabled: true
  # Host and port to listen on, port 22 is not allowed for your safety
  listen: 127.0.0.1:2222
  # A file containing the ssh host private key to use
  # A decent way to generate one: ssh-keygen -t ed25519 -f ssh_host_ed25519_key -N "" < /dev/null
  host_key: /etc/nebula/ssh_host_ed25519_key
  # A file containing a list of authorized public keys
  authorized_users:
    - user: lighthouse
      # keys can be an array of strings or single string
      keys:
        - "..."
# ... parts before firewall -> inbound
  inbound:
    # Allow icmp between any nebula hosts
    - port: any
      proto: icmp
      host: any

    # Allow tcp/443 from any host with BOTH laptop and home group
    - port: 443
      proto: tcp
      groups:
        - laptop
        - home

    - port: 2222
      proto: tcp
      groups:
        - ssh
```

brad-defined:
Thanks for reporting this issue!
I've also encountered this behavior. I believe what's happening is the Nebula SSH server attempts to start before the Nebula interface starts. Since the SSH server is configured to use only the Nebula IP, and since the Nebula device isn't available yet, it'll fail at startup. But a reconfig successfully starts the SSH server because during the reconfig, the Nebula interface is available.

PR #375, opened 2021-02-05
Ensure the Nebula device exists before attempting to bind to the Nebula IP

brad-defined:
As reported in https://github.com/slackhq/nebula/issues/243 and https://github.com/slackhq/nebula/issues/314, internal services that are configured to bind only to the Nebula IP sometimes fail because they attempt to bind to the Nebula interface before it's created. This diff moves the start of the SSH, DNS, and Stats internal services until after we're certain that the Nebula interface is created.


brad-defined:
I rebased against master, and re-tested. I discovered that I never actually fixed the race in the stats case, so that code moved around a little bit. (I didn't wrap the Listen call in the delayed function, so it still failed when bound only to the Nebula IP. Oops.)

I also updated the Control struct to have named variables for the delayed startup functions, instead of the more generic array.

sshd/server.go
```diff
@@ -93,6 +93,7 @@ func (s *SSHServer) Run(addr string) error {
 	var err error
 	s.listener, err = net.Listen("tcp", addr)
 	if err != nil {
+		s.l.WithField("err", err).Warn("Error starting SSH server listener, shutting down")
```
nbrownus, line 96: Caller should be the one to log

stats.go
```diff
@@ -17,24 +17,35 @@ import (
...
 
 	interval := c.GetDuration("stats.interval", 0)
 	if interval == 0 {
-		return fmt.Errorf("stats.interval was an invalid duration: %s", c.GetString("stats.interval", ""))
+		return nil, fmt.Errorf("stats.interval was an invalid duration: %s", c.GetString("stats.interval", ""))
 	}
 
+	var startFn func()
 	switch mType {
 	case "graphite":
```
nbrownus, line 36: I recall an old debate regarding this, I think, will need to noodle

sshd/server.go
brad-defined: Moved the log line, and verified that I saw the error log when misconfiguring the ssh listener.

nbrownus (approved):
Looks good, also closes #393

PR #378, opened 2021-02-11
Update README.md

stangri:
Add missing period.


stangri:
@nbrownus @AndrewChubatiuk sorry, just realized I never followed up on that and for some reason GitHub doesn't offer me a list of reviewers to choose from, so I'm tagging you gentlemen manually.

Trivial fix, not sure why it slipped thru. ;)

nbrownus (approved):

PR #382, opened 2021-02-17
Proper multiqueue support for tun devices

wadey:
This change is for Linux only.

Previously, when running with multiple `tun.routines`, we would only have one file descriptor. This change instead sets IFF_MULTI_QUEUE and opens a file descriptor for each routine. This allows us to process with multiple threads while preventing out of order packet reception issues.

To attempt to distribute the flows across the queues, we try to write to the tun/UDP queue that corresponds with the one we read from. So if we read a packet from tun queue "2", we will write the outgoing encrypted packet to UDP queue "2". Because of the nature of how multi queue works with flows, a given host tunnel will be sticky to a given routine (so if you try to performance benchmark by only using one tunnel between two hosts, you are only going to be using a max of one thread for each direction).

Because this system works much better when we can correlate flows between the tun and udp routines, we are deprecating the undocumented "tun.routines" and "listen.routines" parameters and introducing a new "routines" parameter that sets the value for both. If you use the old undocumented parameters, the max of the values will be used and a warning logged.

Co-author: @nbrownus

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -86,13 +86,11 @@ listen:
   #read_buffer: 10485760
   #write_buffer: 10485760
 
+# EXPERIMENTAL
 # Routines is the number of thread pairs to run that consume from the tun and UDP queues.
 # Currently, this defaults to 1 which means we have 1 tun queue reader and 1
 # UDP queue reader. Setting this above one will set IFF_MULTI_QUEUE on the tun
 # device and SO_REUSEPORT on the UDP socket to allow multiple queues.
-#
-# You can also set this to 0 (or the special string "auto") to pick a good default for
-# the number of CPUs you have. The current logic is Max((NumCPU - 1) / 2, 1).
 #routines: 1
 
 punchy:
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -86,6 +86,15 @@ listen:
   #read_buffer: 10485760
   #write_buffer: 10485760
 
+# Routines is the number of thread pairs to run that consume from the tun and UDP queues.
+# Currently, this defaults to 1 which means we have 1 tun queue reader and 1
+# UDP queue reader. Setting this above one will set IFF_MULTI_QUEUE on the tun
+# device and SO_REUSEPORT on the UDP socket to allow multiple queues.
+#
+# You can also set this to the special string "auto" to pick a good default for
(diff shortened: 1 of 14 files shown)


examples/config.yml
```diff
@@ -86,6 +86,15 @@ listen:
   #read_buffer: 10485760
   #write_buffer: 10485760
 
+# Routines is the number of thread pairs to run that consume from the tun and UDP queues.
+# Currently, this defaults to 1 which means we have 1 tun queue reader and 1
+# UDP queue reader. Setting this above one will set IFF_MULTI_QUEUE on the tun
+# device and SO_REUSEPORT on the UDP socket to allow multiple queues.
+#
+# You can also set this to the special string "auto" to pick a good default for
+# the number of CPUs you have. The current logic is (NumCPU - 1) / 2.
```
nbrownus, line 95: We should point out that the minimum value is 1 with `auto`. Also maybe `0` == `auto`?

interface.go
```diff
@@ -97,6 +102,8 @@ func NewInterface(c *InterfaceConfig) (*Interface, error) {
 		udpQueues:          c.udpQueues,
 		tunQueues:          c.tunQueues,
 		version:            c.version,
+		writers:            make([]*udpConn, c.udpQueues),
```
nbrownus, line 105: We can simplify the config and tun structs to only have `routines` now instead of a distinct `tun` and `udp` queue counts

examples/config.yml
```diff
@@ -86,6 +86,13 @@ listen:
   #read_buffer: 10485760
   #write_buffer: 10485760
 
+# EXPERIMENTAL
```
nbrownus, line 89: We should note this only applies to linux

main.go
```diff
@@ -93,6 +93,30 @@ func Main(config *Config, configTest bool, buildVersion string, logger *logrus.L
...
 
+	var routines int
+
+	// If `routines` is set, use that and ignore the specific values
+	if routines = config.GetInt("routines", 0); routines != 0 {
+		if routines < 1 {
+			routines = 1
+		}
+		if routines > 1 {
+			l.WithField("routines", routines).Info("using multiple routines")
```
nbrownus, line 104: Convention has been to uppercase the first letter `Using`, another one below

nbrownus (approved):

rawdigits (approved):

PR #383, opened 2021-02-17
Add a check in the makefile to ensure a minimum version of go is installed

nbrownus:
This adds a check to the `Makefile` that informs folks that are trying to build nebula if they are using a version of go that is inappropriate.

I'm no `make` expert but this is working on OSX (gnu make 3.81) and Linux (gnu make 4.3)


rawdigits (approved):
I was prob wrong that this would affect many folks, but either way thanks. :)

wadey (approved):
LGTM :+1:

PR #387, opened 2021-02-22
Increase MaxRemotes and warn if reaching that limit

krombel:
While running nebula on a docker host (which dozens of host ips as you might guess) we ran into the issue that the MaxRemotes of 10 weren't enough.
As we wanted to be at least warned in that case (as it would have reduced our debugging time enormously) I added a log as well so people may find it in their logs if a host reports more than MaxRemotes hosts to the lighthouse.

Alternative approach:
Add a Filter in addRemotes to only include an entry in Remotes if it may be addressed via preferred_range or isPrivate (optionally only in case we are reaching the limit)

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -15,7 +15,7 @@ import (
 
 //const ProbeLen = 100
 const PromoteEvery = 1000
-const MaxRemotes = 10
+const MaxRemotes = 25
 
 // How long we should prevent roaming back to the previous IP.
 // This helps prevent flapping due to packets already in flight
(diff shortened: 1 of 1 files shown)


wadey:
I think it would be a better idea to make MaxRemotes a config setting instead of increasing the constant value a bit.

Did you see that you can filter out all docker interfaces using the lighthouse local_allow_list?

```
lighthouse:
  local_allow_list:
    interfaces:
      'docker.*': false
```

more details in the example config: https://github.com/slackhq/nebula/blob/v1.3.0/examples/config.yml#L59-L72

PR #388, opened 2021-02-24
Minimalist config examples

rawdigits:
*WIP*

The existing config.yml has the kitchen sink and makes users aware of every possible option, but is daunting for first time users. These configurations can help someone get started with just the settings they need to run nebula.


jdoss (approved):
LGTM

PR #391, opened 2021-02-25
Routine-local conntrack cache

wadey:
Previously, every packet we see gets a lock on the conntrack table and updates it. When running with multiple routines, this can cause heavy lock contention and limit our ability for the threads to run independently. This change caches reads from the conntrack table for a very short period of time to reduce this lock contention. This cache will currently default to disabled unless you are running with multiple routines, in which case the default cache delay will be 1 second. This means that entries in the conntrack table may be up to 1 second out of date and remain in a routine local cache for up to 1 second longer than the global table.

Instead of calling `time.Now()` for every packet, this cache system relies on a tick thread that updates the current cache "version" each tick. Every packet we check if the cache version is out of date, and reset the cache if so.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -6,7 +6,6 @@ import (
 	"net"
 	"os"
 	"runtime"
-	"sync/atomic"
 	"time"
 
 	"github.com/rcrowley/go-metrics"
@@ -39,6 +38,7 @@ type InterfaceConfig struct {
 	DropMulticast           bool
 	UDPBatchSize            int
 	routines                int
+	ConntrackCache          time.Duration
 	MessageMetrics          *MessageMetrics
 	version                 string
 }
@@ -60,6 +60,7 @@ type Interface struct {
 	dropMulticast      bool
 	udpBatchSize       int
 	routines           int
+	conntrackCache     time.Duration
 	version            string
 
 	writers []*udpConn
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -117,6 +117,18 @@ func Main(config *Config, configTest bool, buildVersion string, logger *logrus.L
 		}
 	}
 
+	// EXPERIMENTAL
+	// Intentionally not documented yet while we do more testing and determine
+	// a good default value.
+	conntrackCache := config.GetDuration("firewall.conntrack.routine_cache", 0)
```
nbrownus, line 123: Feels like we should call this `routine_cache_timeout` or similar

interface.go
```diff
@@ -38,6 +38,7 @@ type InterfaceConfig struct {
 	DropMulticast           bool
 	UDPBatchSize            int
 	routines                int
+	ConntrackCache          time.Duration
```
nbrownus, line 41: The names feel like they collide here, `conntrackCache` being a type and an object while `ConntrackCache` is a duration. Move this one and the one below to `ConntrackCacheTimeout`?

nbrownus (dismissed):

nbrownus (approved):

PR #394, opened 2021-03-01
Add a way to trigger punch backs via lighthouse

nbrownus:
`Rebind` is called by the controlling program when a network interface change is detected (iOS only for now, soon to be others). This code adds a way to signal through the lighthouse that a remote should punch back on a possible new IP for the local, greatly improving the roaming story.

Before this change, roaming on a routable ip generally meant the tunnel would have to die and be stood back up.


wadey (approved):
This looks good to me. The race detector will complain about the Interface.rebindCount, but we know there isn't actually a race issue there so I'm okay with it :+1:.

Issue #316, opened 2020-10-16
fatal error: concurrent map writes


bonedaddy:
# Overview

* When running a node that reports back to more than one lighthouse there is a race condition that appears to be triggered during the handshaking process
* It also looks like this only happens when prometheus stats is enabled but im not certain about that
* Possible source of race is the [packetStore field](https://github.com/slackhq/nebula/blob/f6d0b4b893c620345d9c1c71502a21fe4887cc08/hostmap.go#L46)
# Version

* 1.3.0

# Configuration

* 2 lighthouses
* prometheus listening on `0.0.0.0:8081`

# Error Log
```
         github.com/slackhq/nebula/lighthouse.go:401 +0xc6c
 created by github.com/slackhq/nebula.(*LightHouse).HandleRequest
         github.com/slackhq/nebula/lighthouse.go:402 +0x30
 github.com/slackhq/nebula.(*LightHouse).HandleRequest.func2(0x40001eb710, 0x658920, 0x40000ce000)
         runtime/time.go:188 +0xc0
 time.Sleep(0x12a05f200)
 goroutine 113 [sleep]:
         net/http/server.go:2969 +0x2e8
 created by net/http.(*Server).Serve
         net/http/server.go:1851 +0x5ac
 net/http.(*conn).serve(0x400023a000, 0x65eca0, 0x4000207b00)
         net/http/server.go:984 +0x150
 net/http.(*conn).readRequest(0x400023a000, 0x65eca0, 0x4000207b00, 0x0, 0x0, 0x0)
         net/http/request.go:1012 +0x74
 net/http.readRequest(0x400005b140, 0x0, 0x4000240100, 0x0, 0x0)
         net/textproto/reader.go:39
 net/textproto.(*Reader).ReadLine(...)
         net/textproto/reader.go:58 +0x60
 net/textproto.(*Reader).readLineSlice(0x4000222a80, 0x4000240100, 0x4000020180, 0x7f5f256000, 0x0, 0x0)
         bufio/bufio.go:389 +0x30
 bufio.(*Reader).ReadLine(0x400005b140, 0x3daf9c, 0x4000193800, 0x7f85f127d0, 0x0, 0x4000222a80, 0x7f5f256090)
         bufio/bufio.go:360 +0x38
 bufio.(*Reader).ReadSlice(0x400005b140, 0x1c80a, 0x4000240100, 0x100, 0xf8, 0x589f00, 0x1)
         bufio/bufio.go:101 +0xf8
[...]

Issue #226, opened 2020-05-05
Race condition:  handshake finishing while packet is processing


fm407:
```
ERRO[0080] Failed to call noise.WriteMessage             error="noise: unexpected call to WriteMessage should be ReadMessage" handshake="map[stage:0 style:ix_psk0]" vpnIp=10.10.70.82
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x989fde]
goroutine 29 [running]:
github.com/slackhq/nebula.ixHandshakeStage0(0xc0000f0480, 0xc00a3c5a52, 0xc0000f0300)
        github.com/slackhq/nebula/handshake_ix.go:41 +0x50e
github.com/slackhq/nebula.(*Interface).getOrHandshake(0xc0000f0480, 0xa0a1952, 0x2329)
        github.com/slackhq/nebula/inside.go:91 +0x132
github.com/slackhq/nebula.(*Interface).consumeInsidePacket(0xc0000f0480, 0xc000246500, 0xc8, 0x2329, 0xc0002382e0, 0xc0002382f      0, 0xc, 0xc, 0xc000248a00, 0x2329, ...)
        github.com/slackhq/nebula/inside.go:32 +0xe6
github.com/slackhq/nebula.(*Interface).listenIn(0xc0000f0480, 0x0)
        github.com/slackhq/nebula/interface.go:155 +0x13a
created by github.com/slackhq/nebula.(*Interface).Run
        github.com/slackhq/nebula/interface.go:116 +0x59f
 ```

Operating System: Kali GNU/Linux Rolling
            Kernel: Linux 5.4.0-kali3-amd64
             Architecture: x86-64

Nebula Version: 1.2.0

wadey:
I'm pretty sure the issue here is certState getting set to nil here while the new packet is processing: https://github.com/slackhq/nebula/blob/v1.2.0/hostmap.go#L568

Issue #147, opened 2020-01-01
crash due to concurrent map writes


missdeer:
```
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 30 [running]:
runtime.throw(0x613713, 0x15)
	F:/go/src/runtime/panic.go:774 +0x5c fp=0x1ec5244 sp=0x1ec5230 pc=0x404e0
runtime.mapassign(0x580468, 0x2a9fcc0, 0x6f62b9, 0xf0)
	F:/go/src/runtime/map.go:680 +0x538 fp=0x1ec5284 sp=0x1ec5244 pc=0x1d7e8
github.com/slackhq/nebula.ixHandshakeStage0(0x1c6b200, 0xc0a86464, 0x20364d0)
	G:/sourcecode/nebula/handshake_ix.go:67 +0xaf8 fp=0x1ec575c sp=0x1ec5284 pc=0x4f6c30
github.com/slackhq/nebula.(*Interface).getOrHandshake(0x1c6b200, 0xc0a86464, 0x200d640)
	G:/sourcecode/nebula/inside.go:83 +0x98 fp=0x1ec57ac sp=0x1ec575c pc=0x504e20
github.com/slackhq/nebula.(*Interface).SendMessageToVpnIp(0x1c6b200, 0x1c00003, 0xc0a86464, 0x1c70ec0, 0xa, 0xa, 0x1c70ed0, 0xc, 0xc, 0x1e31400, ...)
	G:/sourcecode/nebula/inside.go:113 +0x24 fp=0x1ec57ec sp=0x1ec57ac pc=0x505234
github.com/slackhq/nebula.(*LightHouse).QueryServer(0x1d7c000, 0xc0a86450, 0x701cb0, 0x1c6b200)
	G:/sourcecode/nebula/lighthouse.go:82 +0x18c fp=0x1ec591c sp=0x1ec57ec pc=0x508348
github.com/slackhq/nebula.(*LightHouse).Query(0x1d7c000, 0xc0a86450, 0x701cb0, 0x1c6b200, 0x1, 0x490fc, 0x0, 0x0, 0x0)
	G:/sourcecode/nebula/lighthouse.go:58 +0x174 fp=0x1ec5950 sp=0x1ec591c pc=0x508198
github.com/slackhq/nebula.ixHandshakeStage0(0x1c6b200, 0xc0a86450, 0x20360e0)
[...]

PR #396, opened 2021-03-05
Fix most known data races

wadey:
This change fixes all of the known data races that `make smoke-docker-race` finds, except for one.

Most of these races are around the handshake phase for a hostinfo, so we add a RWLock to the hostinfo and Lock during each of the handshake stages.

Some of the other races are around consistently using `atomic` around the `messageCounter` field. To make this harder to mess up, I have renamed the field to `atomicMessageCounter` (I also removed the unnecessary extra pointer deference as we can just point directly to the struct field).

The last remaining data race is around reading `ConnectionInfo.ready`, which is a boolean that is only written to once when the handshake has finished. Due to it being in the hot path for packets and the rare case that this could actually be an issue, holding off on fixing that one for now.

here is the results of `make smoke-docker-race`:

before:

```
lighthouse1: Found 2 data race(s)
host2:       Found 36 data race(s)
host3:       Found 17 data race(s)
host4:       Found 31 data race(s)
```

after:

```
host2: Found 1 data race(s)
host4: Found 1 data race(s)
```

Fixes: #147
Fixes: #226
Fixes: #283
Fixes: #316

### DIFF
--- a/handshake_manager.go
+++ b/handshake_manager.go
@@ -103,6 +103,8 @@ func (c *HandshakeManager) handleOutbound(vpnIP uint32, f EncWriter, lighthouseT
 	if err != nil {
 		return
 	}
+	hostinfo.Lock()
+	defer hostinfo.Unlock()
 
 	// If we haven't finished the handshake and we haven't hit max retries, query
 	// lighthouse and then send the handshake packet again.
--- a/inside.go
+++ b/inside.go
@@ -106,16 +106,18 @@ func (f *Interface) getOrHandshake(vpnIp uint32) *HostInfo {
 		hostinfo.Lock()
 		defer hostinfo.Unlock()
 
-		ixHandshakeStage0(f, vpnIp, hostinfo)
-		// FIXME: Maybe make XX selectable, but probably not since psk makes it nearly pointless for us.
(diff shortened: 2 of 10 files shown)


inside.go
```diff
@@ -223,7 +228,7 @@ func (f *Interface) sendNoMetrics(t NebulaMessageType, st NebulaMessageSubType,
 	var err error
 	//TODO: enable if we do more than 1 tun queue
```
forfuncsake, line 229: Is this TODO needing to be auctioned now with multi queue?

wadey: @forfuncsake No because we use sync/atomic for the messageCounter now. We should remove that outdated TODO.

handshake_manager.go
```diff
@@ -103,6 +103,8 @@ func (c *HandshakeManager) handleOutbound(vpnIP uint32, f EncWriter, lighthouseT
 	if err != nil {
 		return
 	}
+	hostinfo.Lock()
```
nbrownus, line 106: It could get confusing but at this point we are close to optimizing away the hostmap lock here

nbrownus (approved):

PR #399, opened 2021-03-08
Do not allow someone to run a nebula lighthouse with an ephemeral port

rawdigits:
Running a lighthouse with an ephemeral port doesn't make sense, so we should not allow nebula to be started this way.


rcrowley (dismissed):
LGTM :)

rawdigits:
@rcrowley You Have No Power Here!

nbrownus (approved):

nbrownus (approved):

PR #400, opened 2021-03-09
fix race in getOrHandshake

wadey:
We missed this race with #396 (and I think this is also the true crash in
issue #226). We need to lock a little higher in the getOrHandshake
method, before we reset hostinfo.ConnectionInfo. Previously, two
routines could enter this section and confuse the handshake process.

This could result in the other side sending a recv_error that also has
a race with setting hostinfo.ConnectionInfo back to nil. So we make sure
to grab the lock in handleRecvError as well.

Neither of these code paths are in the hot path (handling packets
between two hosts over an active tunnel) so there should be no
performance concerns.

### DIFF
--- a/inside.go
+++ b/inside.go
@@ -91,6 +91,17 @@ func (f *Interface) getOrHandshake(vpnIp uint32) *HostInfo {
 		return hostinfo
 	}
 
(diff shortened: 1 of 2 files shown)


forfuncsake (approved):
LGTM

PR #401, opened 2021-03-10
Refactor handshake_ix

wadey:
There are some subtle race conditions with the previous handshake_ix implementation, mostly around collisions with localIndexId. This change refactors it so that we have  a "commit" phase during the handshake where we grab the lock for the hostmap and ensure that we have a unique local index before storing it. We also now avoid using the pending hostmap at all for receiving stage1 packets, since we have everything we need to just store the completed handshake.

### DIFF
--- a/handshake_ix.go
+++ b/handshake_ix.go
@@ -154,8 +154,6 @@ func ixHandshakeStage1(f *Interface, addr *udpAddr, hostinfo *HostInfo, packet [
 			hostId:          vpnIP,
 			HandshakePacket: make(map[uint8][]byte, 0),
 		}
-		hostinfo.Lock()
-		defer hostinfo.Unlock()
 
 		l.WithField("vpnIp", IntIp(vpnIP)).WithField("udpAddr", addr).
 			WithField("certName", certName).
@@ -238,6 +236,7 @@ func ixHandshakeStage1(f *Interface, addr *udpAddr, hostinfo *HostInfo, packet [
 
 			//hostinfo.ClearRemotes()
 			hostinfo.AddRemote(*addr)
+			hostinfo.ForcePromoteBest(f.hostMap.preferredRanges)
 			hostinfo.CreateRemoteCIDR(remoteCert)
 			f.lightHouse.AddRemoteAndReset(ip, addr)
 			if f.serveDns {
@@ -256,9 +255,9 @@ func ixHandshakeStage1(f *Interface, addr *udpAddr, hostinfo *HostInfo, packet [
 				f.hostMap.DeleteHostInfo(ho)
 			}
 
-			f.hostMap.AddVpnIPHostInfo(vpnIP, hostinfo)
-
 			hostinfo.handshakeComplete()
+			f.hostMap.AddVpnIPHostInfo(vpnIP, hostinfo)
+			return false
 		} else {
 			l.WithField("vpnIp", IntIp(hostinfo.hostId)).WithField("udpAddr", addr).
 				WithField("certName", certName).
--- a/handshake_ix.go
+++ b/handshake_ix.go
@@ -98,34 +98,6 @@ func ixHandshakeStage1(f *Interface, addr *udpAddr, hostinfo *HostInfo, packet [
 			return true
 		}
 
-		hostinfo, _ := f.handshakeManager.pendingHostMap.QueryReverseIndex(hs.Details.InitiatorIndex)
-		if hostinfo != nil {
-			hostinfo.RLock()
-			defer hostinfo.RUnlock()
-
-			if bytes.Equal(hostinfo.HandshakePacket[0], packet[HeaderLen:]) {
-				if msg, ok := hostinfo.HandshakePacket[2]; ok {
-					f.messageMetrics.Tx(handshake, NebulaMessageSubType(msg[1]), 1)
-					err := f.outside.WriteTo(msg, addr)
-					if err != nil {
-						l.WithField("vpnIp", IntIp(hostinfo.hostId)).WithField("udpAddr", addr).
-							WithField("handshake", m{"stage": 2, "style": "ix_psk0"}).WithField("cached", true).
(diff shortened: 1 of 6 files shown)


hostmap.go
```diff
@@ -234,19 +234,21 @@ func (hm *HostMap) DeleteIndex(index uint32) {
...
+	hm.Unlock()
 
-		// Check if we have an entry under hostId that matches the same hostinfo
-		// instance. Clean it up as well if we do.
-		var hostinfo2 *HostInfo
-		hostinfo2, ok = hm.Hosts[hostinfo.hostId]
-		if ok && hostinfo2 == hostinfo {
-			delete(hm.Hosts, hostinfo.hostId)
-		}
+	hostinfo.RLock()
```
forfuncsake, line 242: hostinfo could be `nil` here in a `!ok` case, yeah?

wadey: Yes good catch! Will fix

hostmap.go
```diff
@@ -234,16 +234,18 @@ func (hm *HostMap) DeleteIndex(index uint32) {
...
+		hostinfo.RUnlock()
 
 		// Check if we have an entry under hostId that matches the same hostinfo
 		// instance. Clean it up as well if we do.
-		var hostinfo2 *HostInfo
-		hostinfo2, ok = hm.Hosts[hostinfo.hostId]
+		hm.Lock()
+		hostinfo2, ok := hm.Hosts[hostId]
 		if ok && hostinfo2 == hostinfo {
 			delete(hm.Hosts, hostinfo.hostId)
```
forfuncsake, line 250: This line should use the local `hostId` var

forfuncsake (dismissed):

handshake_ix.go
```diff
@@ -146,7 +146,7 @@ func ixHandshakeStage1(f *Interface, addr *udpAddr, hostinfo *HostInfo, packet [
 			return true
 		}
 
-		hostinfo, err = f.handshakeManager.AddIndex(myIndex, ci)
+		hostinfo, err = f.handshakeManager.AddIndex(vpnIP, myIndex, hs.Details.InitiatorIndex, ci)
```
forfuncsake, line 149: There's a lock call 9 lines below here that can probably be moved down above line 203:

```
203: hostinfo.HandshakePacket[0] = make([]byte, len(packet[HeaderLen:]))
```

forfuncsake (approved):
LGTM, tiny optimization available

nbrownus (approved):
😅 I think this is correct, thank you for suffering through it with me. will test on my networks too. tentative approval.

handshake_ix.go
```diff
@@ -73,206 +73,174 @@ func ixHandshakeStage0(f *Interface, vpnIp uint32, hostinfo *HostInfo) {
...
+// to tear down.
 func ixHandshakeStage1(f *Interface, addr *udpAddr, hostinfo *HostInfo, packet []byte, h *Header) bool {
 	var ip uint32
-	if h.RemoteIndex == 0 {
-		ci := f.newConnectionState(false, noise.HandshakeIX, []byte{}, 0)
-		// Mark packet 1 as seen so it doesn't show up as missed
-		ci.window.Update(1)
+	if h.RemoteIndex != 0 {
+		// We already have this index, add the new remote addr
+		f.hostMap.AddRemote(ip, addr)
```
forfuncsake, line 82: I'm a bit concerned by this zero value `ip`, is that right?

wadey: hmmm yeah good point, looking at the blame this has been wrong for a long time! This must be dead code then that never actually triggers.

handshake_ix.go
```diff
@@ -73,206 +73,167 @@ func ixHandshakeStage0(f *Interface, vpnIp uint32, hostinfo *HostInfo) {
 
 }
 
+// Note we always return `false` in ixHandshakeStage1 because there is nothing
+// to tear down.
 func ixHandshakeStage1(f *Interface, addr *udpAddr, hostinfo *HostInfo, packet []byte, h *Header) bool {
```
forfuncsake, line 78: While we're cleaning up.... this hostinfo input value is never used (we always overwrite it now, but before your change we always shadowed it!), can we remove it?

handshake_manager.go
```diff
@@ -196,18 +197,96 @@ func (c *HandshakeManager) AddVpnIP(vpnIP uint32) *HostInfo {
...
+func (c *HandshakeManager) AddIndexHostInfo(h *HostInfo) error {
+	c.pendingHostMap.Lock()
+	defer c.pendingHostMap.Unlock()
+	c.mainHostMap.RLock()
+	defer c.mainHostMap.RUnlock()
+
+	for i := 0; i < 32; i++ {
+		index, err := generateIndex()
+		if err != nil {
+			return err
```
forfuncsake, line 276: Given this is not an infinite loop anymore, it might be worth a log and `continue` here.
Although I have no idea if an error from the internal `rand.Read()` call would be non-fatal.

hostmap.go
```diff
@@ -400,36 +362,26 @@ func (hm *HostMap) queryUnsafeRoute(ip uint32) uint32 {
...
-			return false
-		}
-		complete := i.HandshakeComplete
-		hm.RUnlock()
-		return complete
+	hm.Hosts[hostinfo.hostId] = hostinfo
+	hm.Indexes[hostinfo.localIndexId] = hostinfo
+	hm.RemoteIndexes[hostinfo.remoteIndexId] = hostinfo
 
+	if l.Level > logrus.DebugLevel {
```
forfuncsake, line 380:
```suggestion
	if l.Level >= logrus.DebugLevel {
```

handshake_manager.go
wadey: Yeah I think the error from rand.Read would be so unexpected its okay to return the error from here.

rawdigits (approved):
tired

Issue #313, opened 2020-10-10
Trouble connecting between two nodes


brianluong:
Hey all!

I had a weird issue where Host A could not `ping` Host B over Nebula. I happened to resolve it but am not sure
1. Why it happened in the first place 
2. Why my "fix" worked

(For some additional context, we have thousands of nodes connecting happily over Nebula and this is the first time we've observed this issue)

## Debugging 
Host A: datalake-queue4-app.ca1.qprod.net, vpnIP: `100.70.39.63`, mainIP: `10.112.39.63`
Host B: secdev-canary1-app.ca1.qprod.net, vpnIP: `100.70.38.110`, mainIP: `10.112.38.110`

Host A  logs:
```
Oct 10 10:34:24 datalake-queue4-app.ca1.qprod.net nebula[17345] {"certName":"secdev-canary1-app.ca1.qprod.net","handshake":{"stage":1,"style":"ix_psk0"},"initiatorIndex":1059072191,"level":"info","msg":"Handshake message received","remoteIndex":0,"responderIndex":0,"time":"2020-10-10T10:34:24-06:00","udpAddr":{"ip":"10.112.38.110","port":4333},"vpnIp":"100.70.38.110"}
Oct 10 10:34:24 datalake-queue4-app.ca1.qprod.net nebula[17345]: {"certName":"secdev-canary1-app.ca1.qprod.net","handshake":{"stage":1,"style":"ix_psk0"},"initiatorIndex":1059072191,"level":"info","msg":"Prevented a handshake race","remoteIndex":0,"responderIndex":3462305623,"time":"2020-10-10T10:34:24-06:00","udpAddr":{"ip":"10.112.38.110","port":4333},"vpnIp":"100.70.38.110"}
Oct 10 10:34:24 datalake-queue4-app.ca1.qprod.net nebula[17345]: {"certName":"secdev-canary1-app.ca1.qprod.net","handshake":{"stage":1,"style":"ix_psk0"},"initiatorIndex":1059072191,"level":"info","msg":"Handshake message received","remoteIndex":0,"responderIndex":0,"time":"2020-10-10T10:34:24-06:00","udpAddr":{"ip":"10.112.38.110","port":4333},"vpnIp":"100.70.38.110"}
[...]

PR #402, opened 2021-03-10
don't allow a useless handshake with yourself

rawdigits:
This prevents a silly condition noticed in #313 where a host can handshake with itself. 

Notes:
* This is NOT a fix for that issue, and is unlikely to be the cause of the issue, but something we should also stop from happening, because this is a useless entry in the hostMap.
* This will not cleanly merge with #401 and that PR should take precedence. I will update this one once it is merged.

To reproduce this issue you can simply create a `static_host_map` entry pointing a nebula ip besides your own at your local nebula port and initiating traffic to it.

Example:

Host1:
  Real IP: 10.1.1.1
  Nebula listener port: 1234
  Nebula IP: 192.168.1.1
 
```
static_host_map:
  "192.168.1.222": ["10.1.1.1:1234"]
```

```shell
ping 192.168.1.222
(fails)
```

A hostmap entry for yourself will appear and the nebula logs will show a successful handshake.

### DIFF
--- a/handshake_ix.go
+++ b/handshake_ix.go
@@ -107,6 +107,14 @@ func ixHandshakeStage1(f *Interface, addr *udpAddr, packet []byte, h *Header) {
 	certName := remoteCert.Details.Name
 	fingerprint, _ := remoteCert.Sha256Sum()
 
+	if vpnIP == f.GetVpnIP() {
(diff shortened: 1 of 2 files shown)


wadey (approved):
LGTM :+1:

PR #404, opened 2021-03-13
Fix selection of UDP remote to use during stage2

wadey:
The change for #401 incorrectly called HostInfo.ForcePromoteBest in
stage2, when we really we want to pick the remote that we received the
response from.


rawdigits (approved):
This looks more correcter to me. Going to test but I believe this is right.

PR #406, opened 2021-03-16
Fix remote allow list lighthouse lock

tsroten:
Currently, if you use the remote allow list config, as soon as you attempt to create a tunnel to a node that has a blocked IP address, a mutex is locked and never unlocked. This happens even if the node has an allowed remote IP address in addition to the blocked remote IP address.

This pull request ensures that the lighthouse mutex is unlocked whenever we attempt to add a remote IP.


wadey (approved):
Great find, and thank you for the test as well. :raised_hands:

PR #410, opened 2021-03-17
*WIP* Make nebula on windows much faster by buffering on udp

rawdigits:
*WIP* - do not merge!

The default windows UDP receive buffer is 8192KB which is .... way too small.

This increases the buffer to an arbitrary value at the moment, but results in a 10x speed increase on my test. I'll clean it up and make it use the listen.read_buffer and write_buffer in the near future.


nbrownus:
Closing in favor of #905

PR #416, opened 2021-03-22
Don't craft buffers if we don't need them

nbrownus:


hostmap.go
```diff
@@ -599,12 +599,19 @@ func (i *HostInfo) handshakeComplete() {
...
 	atomic.StoreUint64(&i.ConnectionState.atomicMessageCounter, 2)
-	i.logger().Debugf("Sending %d stored packets", len(i.packetStore))
-	nb := make([]byte, 12, 12)
-	out := make([]byte, mtu)
-	for _, cp := range i.packetStore {
-		cp.callback(cp.messageType, cp.messageSubType, i, cp.packet, nb, out)
+
+	if l.Level >= logrus.DebugLevel {
+		i.logger().Debugf("Sending %d stored packets", len(i.packetStore))
+	}
```
rawdigits, line 605: Should we move the log message inside of your new `if` check? It is conceivable that we will eventually trigger handshakes by something other than a packet being sent to a particular address, in which case we would needlessly log `Sending 0 stored packets`.

rawdigits (changes requested):

hostmap.go
nbrownus: EBRAINFART

rawdigits (approved):

hostmap.go
nbrownus: Best I can come up with right now, if you are in debug trying to see why initial packets are doing weird things, seeing `Sending 0 stored packets` could help indicate it what is going on.

But I haven't needed this message in a few years. I can take it or leave it.

PR #417, opened 2021-03-22
Allow more callers to select queues

nbrownus:
Does our best to use the queue we received on to send on


wadey (commented):
My main concern is hacking lightHouse.Query using the queue for the current packet. This means that every lighthouse packet send is going to switch which queue we send/receive from, and cause us to have multiple routines processing inbound packets from the lighthouses. I think it would be much better to leave the lighthouse queries pegged to queue 0 for now (until we decide to store the q to use in the HostInfo)

nbrownus:
After testing at Slack this appears to do nothing of consequence.

PR #418, opened 2021-03-22
Lighthouse performance pass

nbrownus:
With the addition of ipv6 and some bit rot, lighthouse needed a performance pass.

This introduces the gogo protobuf library as well as some enforcement of stored ip limits that should exist.

A few open questions marked as `TODO` in `lighthouse.go`


wadey (dismissed):
LGTM, a few small comments but I am fine with merging and fixing in a follow up since this PR is getting so big. In general this looks safer and better than what we had before (and more tests!) :+1:

wadey (approved):

wadey (approved):

Issue #338, opened 2020-11-19
openwrtx64 can't run nebula


zgcbug:
I downloaded the file nebula（linux-amd64.tar.gz）to run，not found errors。
root@OpenWrt:~# ./etc/nebula/nebula -config /etc/nebula/config.yml
-ash: ./etc/nebula/nebula: not found
root@OpenWrt:~# bash ./etc/nebula/nebula -config /etc/nebula/config.yml
bash: ./etc/nebula/nebula: No such file or directory
root@OpenWrt:~# /etc/nebula/nebula -config /etc/nebula/config.yml
-ash: /etc/nebula/nebula: not found
root@OpenWrt:~# bash /etc/nebula/nebula -config /etc/nebula/config.yml
/etc/nebula/nebula: /etc/nebula/nebula: cannot execute binary file

wadey:
It looks like you are trying to run the binary with `bash`. Assuming you downloaded the binary to `/etc/nebula/nebula`, trying executing it like this (remove the `bash`):

    /etc/nebula/nebula -config /etc/nebula/config.yml

zgcbug:
root@OpenWrt:~# /etc/nebula/nebula -config /etc/nebula/config.yml
-ash: /etc/nebula/nebula: not found
same error

wadey:
You should redownload the binary and put it in either /usr/bin or /usr/local/bin and run it from there

zgcbug:
I've tried，same errors

actronzs:
> I've tried，same errors

on openwrt router:

copy nebula to /usr/bin/
copy config.yml to /etc/nebula/
run it like this: nebula -config /etc/nebula/config.yml

I also suspect that you want to start the wrong type of nebula. for openwrt router architecture (arm, mips, softfloat) must be the right one.

ftong2020:
Interestingly, I encountered this issue as well. I was running Openwrt targeting x86_64(on Vmware Esxi with Intel chip) Version 19.07. The binary works fine on Ubuntu 64bit VMs, but will not work on openwrt.

Error message is exactly the same,
root@OpenWrt:~# ./nebula
-ash: ./nebula: not found

If I switch it to nebula binary targeting darwin(MacOs), it will produce the same error as if it is running on ubuntu
root@OpenWrt:~# ./nebula
�: not foundne 1: ����
./nebula: line 2: H__PAGEZEROx__TEXT����__text__TEXT: not found
./nebula: line 3: syntax error: unexpected "("

[...]

Issue #295, opened 2020-09-06
static link binary for Linux


hronro:
I just found nebula for Linux is not static linked (it's weird since nebula is written in Go), and I can not run it on Alpine Linux.

Shipping a static link binary would helped a lot.

MikePadge:
Copypasta from Stack

By default, if using the net package a build will likely produce a binary with some dynamic linking, e.g. to libc. You can inspect dynamically vs. statically link by viewing the result of ldd output.bin

There are two solutions I've come across:

Disable CGO, via CGO_ENABLED=0
Force the use of the Go implementation of net dependencies, netgo via go build -tags netgo -a -v, this is implemented for a certain platforms
From https://golang.org/doc/go1.2:

The net package requires cgo by default because the host operating system must in general mediate network call setup. On some systems, though, it is possible to use the network without cgo, and useful to do so, for instance to avoid dynamic linking. The new build tag netgo (off by default) allows the construction of a net package in pure Go on those systems where it is possible.

The above assumes that the only CGO dependency is the standard library's net package.

RealOrangeOne:
FYI looks like Nebula is now available in Alpine's package repo in v3.13

https://pkgs.alpinelinux.org/package/v3.13/community/x86_64/nebula

PR #421, opened 2021-03-24
Use CGO_ENABLED=0

RealOrangeOne:
Fixes https://github.com/slackhq/nebula/issues/295

_Might_ fix https://github.com/slackhq/nebula/issues/338

Also impacts https://github.com/slackhq/nebula/pull/48 and https://github.com/slackhq/nebula/pull/68/

Naive attempt to create statically linked binaries, suitable for dstros like Alpine. Unlike previously, binaries now execute on Alpine.

Binary still appears to execute and connect to tunnels fine on Arch.

```
$ file nebula
nebula: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=-XXX, not stripped
```


wadey:
I agree with this change, but only for `linux` builds. I think cgo does some things on darwin, windows that might be necessary there. The only thing that CGO does for us on linux is hostname lookups, but I don't think that should be an issue for most people. The power of having repeatable builds by disabling CGO would be very nice.

Do you think you could change this to only disable CGO on linux builds?

RealOrangeOne:
A quick google search yields no really useful references other than "Where Go calls C". [This](https://stackoverflow.com/questions/61515186/when-using-cgo-enabled-is-must-and-what-happens) is a bit of an interesting read.

I don't have access to a windows or macOS machine to try this patch out, but perhaps instead we should test this on windows and macos to see if it's better / no worse? Using the same compiler flags for each OS is probably a good thing, especially if that 1 OS is the one most of CI runs against.

Alternatively, perhaps instead specifically using `netgo` is an option (`go build -tags osusergo,netgo`). I'll give that a try this evening to see if that also solves the issue without completely disabling CGo.

RealOrangeOne:
@wadey I found a simpler way. By explicitly using `netgo` explicitly instead of the cgo implementation, we keep the benefits of using `cgo` for some things, but with the cross-platform benefits.

I _think_ I prefer this approach?

RealOrangeOne:
I know little about Go's internals, but yeah I do kinda prefer the cgo approach, that's also how [Hugo](https://github.com/gohugoio/hugo/blob/master/goreleaser.yml#L14) does it.

I'll leave the branch as-is for now. Think and I can revert if needed.

I'm really keen to get this merged and shipped ASAP, as it'll help out a lot of platforms!

wadey:
I've updated this PR to just set CGO_ENABLED=0, which I think is the better option. I have been testing with this and see no issues.

nbrownus (approved):

PR #423, opened 2021-03-26
Don't use a global logger

nbrownus:
This is a start to an end to an end testing vision where we can easily control packet flow. The underlying point is that a global logger can't be segmented per process, confusing the debug process when things break.

We can also do fun things around increasingly verbose `make test` targets.

`logrus.FieldLogger` would be a better type to target but it does not expose the log level, useful in avoiding debug logs in the hot path.

### DIFF
--- a/bits.go
+++ b/bits.go
@@ -26,7 +26,7 @@ func NewBits(bits uint64) *Bits {
 	}
 }
 
-func (b *Bits) Check(i uint64) bool {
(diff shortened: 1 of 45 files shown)


wadey (approved):
Getting rid of the global logger is a good thing, even if this PR is huge. Looking over it, it looks safe. We may want to use our own interface for the logger in the future to give more flexibility, but we can think about that when the time comes.

PR #425, opened 2021-03-26
Start of end to end testing with a good handshake between two nodes

nbrownus:
This change makes it possible to boot multiple `nebula`s in a single process, control their message flow, and gain deep access to their memory.

The initial use case is for robust end to end testing of handshakes.

### DIFF
--- a/tun_tester.go
+++ b/tun_tester.go
@@ -1,0 +1,103 @@
+// +build e2e_testing
+
+package nebula
+
+import (
+	"fmt"
+	"io"
+	"net"
+
+	"github.com/sirupsen/logrus"
+)
+
+type Tun struct {
+	Device       string
+	Cidr         *net.IPNet
+	MTU          int
+	UnsafeRoutes []route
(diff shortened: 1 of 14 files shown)


rawdigits (dismissed):

tun_tester.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (c *Tun) Read(b []byte) (int, error) {
+	p := <-c.rxPackets
+	copy(b, p)
+	return len(p), nil
+}
+
+func (c *Tun) NewMultiQueueReader() (io.ReadWriteCloser, error) {
+	return nil, fmt.Errorf("TODO: multiqueue not implemented for darwin")
```
wadey, line 102:
```suggestion
	return nil, fmt.Errorf("TODO: multiqueue not implemented")
```

wadey (dismissed):
LGTM except a few minor wording things

wadey (approved):

PR #426, opened 2021-03-26
Don't use a global ca pool

nbrownus:


rawdigits (dismissed):

wadey (approved):
LGTM, yay one less global.

PR #429, opened 2021-03-31
More LH cleanup

nbrownus:
This addresses a few concerns from #418 

In addition:
- Adds methods to benefit the lighthouse ip structs to avoid creating garbage
- Changes to using 2 uint64 in the pb for ipv6
- LH will now ignore any ip within the running nebula range.


wadey (dismissed):
LGTM, just left one small nit comment

wadey (approved):

PR #433, opened 2021-04-02
Only set serveDns if the host is also configured to be a lighthouse.

brad-defined:
Nebula attempts an Add() call against the global dnsR variable if serveDns is true.
Unfortunately, the global dnsR is only initialized in main.go if the dns config is set to true AND the system is configured to be a lighthouse.
If the host is not configured to be a lighthouse but is also configured to serve DNS, the Add call will panic against the nil dnsR.

### DIFF
--- a/dns_server.go
+++ b/dns_server.go
@@ -133,7 +133,7 @@ func getDnsServerAddr(c *Config) string {
 func startDns(l *logrus.Logger, c *Config) {
 	dnsAddr = getDnsServerAddr(c)
 	dnsServer = &dns.Server{Addr: dnsAddr, Net: "udp"}
-	l.Infof("Starting DNS responder at %s\n", dnsAddr)
+	l.WithField("dnsListener", dnsAddr).Infof("Starting DNS responder")
 	err := dnsServer.ListenAndServe()
 	defer dnsServer.Shutdown()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


dns_server.go
```diff
@@ -130,7 +130,7 @@ func getDnsServerAddr(c *Config) string {
 func startDns(l *logrus.Logger, c *Config) {
 	dnsAddr = getDnsServerAddr(c)
 	dnsServer = &dns.Server{Addr: dnsAddr, Net: "udp"}
-	l.Debugf("Starting DNS responder at %s\n", dnsAddr)
+	l.Infof("Starting DNS responder at %s\n", dnsAddr)
```
nbrownus, line 133: We should use a `WithField` here, either `udpAddr` or `dnsAddr` will work.

main.go
```diff
@@ -341,7 +341,7 @@ func Main(config *Config, configTest bool, buildVersion string, logger *logrus.L
 	//handshakeMACKey := config.GetString("handshake_mac.key", "")
 	//handshakeAcceptedMACKeys := config.GetStringSlice("handshake_mac.accepted_keys", []string{})
 
-	serveDns := config.GetBool("lighthouse.serve_dns", false)
+	serveDns := config.GetBool("lighthouse.am_lighthouse", false) && config.GetBool("lighthouse.serve_dns", false)
```
nbrownus, line 344: I think a warning log in the event `serve_dns` is true while `am_lighthouse` is false would be helpful

nbrownus (changes requested):

dns_server.go
brad-defined: Updated to dnsListener, as this tracks with the "sshListener" field name logged for the same purpose.

main.go
brad-defined: done!

nbrownus (approved):

PR #437, opened 2021-04-06
Refactor remotes and handshaking to give every address a fair shot

nbrownus:
This change unifies the idea of the lighthouse cache and the hostinfo remotes address list. The motivation is:
- To enable a quicker handshake by attempting to transmit the handshake packet to all known remotes on each attempt
- Give every lighthouse an equal opportunity to provide answers, in the event of a buggy or compromised lighthouse, other lighthouses should have their shot at getting the right answer to the client.

Limits are 10 ipv4 and 10 ipv6 per lighthouse (clients) or per client (lighthouse).

Learned addresses changed to only store 1 of each v4 and v6 address. This should benefit hosts that get new public facing ports through a NAT frequently.

This also adds a check to ensure the responder is who we expected at stage 2. There is an open question on whether we keep the result of the work there (less disruptive in the event a lie disrupted a tunnel to a host you were communicating with) or discard the work. There is another PR nearly ready that speaks to this specific problem in a more elegant way.

Given the scope of the change, a few other items were included:
- #244 is folded in since hostinfo marshaling needed rework.
- Lighthouses will not attempt to roam clients, this keeps a lighthouse on the ip proto the client was configured with.
- Listen address calculation is fixed on linux, sometimes `nil:4242` would be reported


hostmap.go
```diff
@@ -607,82 +475,26 @@ func (i *HostInfo) handshakeComplete(l *logrus.Logger) {
...
-}
-
-func (i *HostInfo) unlockedBlockRemote(remote *udpAddr) {
-	if !i.unlockedIsBadRemote(remote) {
-		// We copy here because we are taking something else's memory and we can't trust everything
-		i.badRemotes = append(i.badRemotes, remote.Copy())
+	// We copy here because we likely got this remote from a source that reuses the object
+	if !i.remote.Equals(remote) {
+		i.remote = remote.Copy()
+		i.remotes.LearnRemote(i.hostId, remote.Copy())
```
wadey, line 496: do we need to call .Copy() twice here though? We could probably just reuse the `i.remote` copy in LearnRemote.

lighthouse.go
```diff
@@ -203,70 +211,47 @@ func (lh *LightHouse) DeleteVpnIP(vpnIP uint32) {
...
+		to := NewIp6AndPort(toAddr.IP, uint32(toAddr.Port))
+		if !lh.unlockedShouldAddV6(to) {
 			return
 		}
-		am.v4 = prependAndLimitV4(am.v4, to)
+		am.unlockedPrependV6(lh.myVpnIp, to)
 	}
+
+	// Mark it as static
+	lh.staticList[vpnIp] = struct{}{}
```
wadey, line 240: we don't have the lh lock when we call this, I think we probably want it.

lighthouse.go
```diff
@@ -591,27 +519,39 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, vpnIp uint32, addr
...
+	lhh.lh.Lock()
+	am := lhh.lh.unlockedGetRemoteList(n.Details.VpnIp)
+	am.RLock()
+	lhh.lh.Unlock()
 
-	for _, to := range n.Details.Ip6AndPorts {
-		lhh.lh.addRemoteV6(n.Details.VpnIp, to, false, false)
-	}
+	am.unlockedSetV4(vpnIp, n.Details.Ip4AndPorts, lhh.lh.unlockedShouldAddV4)
+	am.unlockedSetV6(vpnIp, n.Details.Ip6AndPorts, lhh.lh.unlockedShouldAddV6)
```
wadey, line 553: We grab the RLock, but then we call unlockedSetv4 which modifies the remote list, so we should have the write Lock here.

lighthouse.go
```diff
@@ -637,35 +577,13 @@ func (lhh *LightHouseHandler) handleHostUpdateNotification(n *NebulaMeta, vpnIp
...
-	// We prefer the first n addresses if we got too big
-	if len(am.v4) > MaxRemotes {
-		am.v4 = am.v4[:MaxRemotes]
-	}
-
-	if len(am.v6) > MaxRemotes {
-		am.v6 = am.v6[:MaxRemotes]
-	}
+	am.unlockedSetV4(vpnIp, n.Details.Ip4AndPorts, lhh.lh.unlockedShouldAddV4)
+	am.unlockedSetV6(vpnIp, n.Details.Ip6AndPorts, lhh.lh.unlockedShouldAddV6)
```
wadey, line 585: Same as above, we should have the Write Lock here.

examples/config.yml
```diff
@@ -202,16 +202,16 @@ logging:
...
-  # Total time to try a handshake = sequence of `try_interval * retries`
-  # With 100ms interval and 20 retries it is 23.5 seconds
+  # Handshakes are sent to all known addresses at each interval with a linear backoff,
+  # Wait try_interval after the 1st attempt, 2 * try_interval after the 2nd, etc, until the handshake is older than timeout
+  # A 100ms interval with the default 10 retries will give a handshake 5.5 seconds to resolve before timing out
   #try_interval: 100ms
-  #retries: 20
-  # wait_rotation is the number of handshake attempts to do before starting to try non-local IP addresses
-  #wait_rotation: 5
+  #retries: 20s
```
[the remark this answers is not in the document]
nbrownus: Nope, just a typo. It's still an int and not a duration.

handshake_ix.go
```diff
@@ -15,13 +15,8 @@ import (
...
 	if hostinfo.remote == nil {
-		ips, err := f.lightHouse.Query(vpnIp, f)
-		if err != nil {
-			//l.Debugln(err)
-		}
-		for _, ip := range ips {
-			hostinfo.AddRemote(ip)
-		}
+		//TODO: do we even need this here???
+		f.lightHouse.QueryServer(vpnIp, f)
```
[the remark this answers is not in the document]
nbrownus: I haven't tested with this removed but handshake manager has some specific logic that handles requery, this is a bit of a hammer in comparison. Will need more noodles.

handshake_manager.go
```diff
@@ -252,7 +243,22 @@ func (c *HandshakeManager) CheckAndComplete(hostinfo *HostInfo, handshakePacket
...
+	// Check if we are also handshaking with this vpn ip
+	pendingHostInfo, found := c.pendingHostMap.Hosts[hostinfo.hostId]
+	if found && pendingHostInfo != nil {
+		if !overwrite {
+			// We won, let our pending handshake win
+			return pendingHostInfo, ErrExistingHandshake
+		}
+
+		// We lost, take this handshake and move any cached packets over so they get sent
+		hostinfo.packetStore = append(hostinfo.packetStore, pendingHostInfo.packetStore...)
```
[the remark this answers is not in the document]
nbrownus: Yes but then we also need other places to lock and this story becomes pretty nasty. `HostInfo.cachePacket` does not lock itself, one place uses `HostInfo.ci.queueLock` and the other does not. Feels like untangling in another PR.

hostmap.go
nbrownus: If remote list doesn't have a discrete copy then any modification to `i.remote` will leak into the remote list. It's probably fine but I wanted to take the easy way out of the thought process here and keep them separated.

lighthouse.go
nbrownus: It's a bit of a hand wave, the only time this function is called is on boot pre-start (currently)

handshake_manager.go
```diff
@@ -109,91 +94,91 @@ func (c *HandshakeManager) handleOutbound(vpnIP uint32, f EncWriter, lighthouseT
...
 	}
-}
 
-func (c *HandshakeManager) NextInboundHandshakeTimerTick(now time.Time) {
-	c.InboundHandshakeTimer.advance(now)
-	for {
-		ep := c.InboundHandshakeTimer.Purge()
-		if ep == nil {
-			break
+	//TODO: this will generate a load of queries for hosts with only 1 ip (i'm not using a lighthouse, static mapped)
```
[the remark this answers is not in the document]
nbrownus: I'll leave this as a TODO for now since the semantics are unchanged, just noting that we can maybe do better.

lighthouse.go
nbrownus: The better way to deal with this is to move this into RemoteList as a bool

wadey (dismissed):
All of my issues have been resolved, this is a big change but I think it looks good (and probably safer than what we have now) :+1:

wadey (approved):

PR #438, opened 2021-04-06
Remove obsolete systemd unit settings (take 2)

JohnMaguire:
Similar to PR #412, except this removes the remaining references.

`dist/arch/nebula.service` is used by the [Arch `nebula`](https://github.com/archlinux/svntogit-community/blob/packages/nebula/trunk/PKGBUILD#L40) package.


nbrownus:
@coderobe since they are the package maintainer in Arch.

coderobe:
lgtm, shipit

nbrownus (approved):

PR #441, opened 2021-04-14
PSK Support

nbrownus:
This implements PSK at stage 0 for noise.

Using PSK gets you 2 main things
1. The first handshake packet is no longer sent in plaintext, which masks the identity of the initiator
2. Makes handshaking with unintended recipients far more difficult

Co-author: @wadey 
Co-author: @rawdigits

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -227,10 +227,10 @@ handshakes:
   # after receiving the response for lighthouse queries
   #trigger_buffer: 64
 
-  # pki can be used to mask the contents of handshakes and makes handshaking with unintended recipients more difficult
+  # psk can be used to mask the contents of handshakes and makes handshaking with unintended recipients more difficult
   # all settings respond to a reload
   psk:
-    # mode defines the how pre shared keys can be used in a handshake
+    # mode defines how the pre shared keys can be used in a handshake
     # `none` (the default) does not send or receive using a psk. Ideally `enforced` is used
     # `transitional-accepting` will send handshakes without using a psk and can receive handshakes using a psk we know about
     # `transitional-sending` will send handshakes using a psk but will still accept handshakes without them
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -228,19 +228,22 @@ handshakes:
   #trigger_buffer: 64
 
   # pki can be used to mask the contents of handshakes and makes handshaking with unintended recipients more difficult
+  # all settings respond to a reload
   psk:
     # mode defines the how pre shared keys can be used in a handshake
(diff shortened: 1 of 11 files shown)


examples/config.yml
```diff
@@ -201,16 +201,45 @@ logging:
...
+
   # trigger_buffer is the size of the buffer channel for quickly sending handshakes
   # after receiving the response for lighthouse queries
   #trigger_buffer: 64
 
+  # pki can be used to mask the contents of handshakes and makes handshaking with unintended recipients more difficult
+  # all settings respond to a reload
+  psk:
+    # mode defines the how pre shared keys can be used in a handshake
+    # `none` (the default) does not send or receive using a psk. Ideally `enforced` is used
```
JohnMaguire, line 219: I wonder how much value the `none` state provides versus using `transitional-accepting` with no PSKs as the default. This would reduce what needs to be modified in the config to enable the feature - and reduces the cognitive overhead of 4 possible states to 3.

handshake_ix.go
```diff
@@ -68,28 +67,51 @@ func ixHandshakeStage0(f *Interface, vpnIp uint32, hostinfo *HostInfo) {
...
+		}
+
+		msg, _, _, err = ci.H.ReadMessage(nil, packet[HeaderLen:])
+		if err != nil {
+			// Calls to ReadMessage with an incorrect psk should fail, try the next one if we have one
+			continue
+		}
+
+		// Sometimes ReadMessage returns fine with a nil psk even if the handshake is using a psk, ensure our protobuf
+		// comes out clean as well
```
JohnMaguire, line 94: I'm curious to understand more about what conditions would cause this to occur - if you know, expanding on the code comment might help future readers.

JohnMaguire (commented):
I need to read over the noise protocol to fully understand how this works, but I like it!

handshake_ix.go
JohnMaguire: After spending a few hours reading about Noise, I think this code comment applies to when PSK is set to `none` or otherwise set to an empty value. Since `k` won't be set to anything yet, we'll simply return out whatever data we get here: https://github.com/flynn/noise/blob/master/state.go#L134-L137

When we have PSK turned on, our `k` will be seeded, and we won't hit the conditional. If we have PSK turned off, and the other side has PSK turned on, I believe we would hit this condition. Or if we have PSK turned off and the other side is just sending garbage.

nbrownus:
There doesn't appear to be a lot of interest in PSK support so I will close this for now. We can bring it back when there is a stronger desire.

PR #443, opened 2021-04-15
Minor windows focused improvements

nbrownus:
This change does the following
- Adjusts the build number for non releases to provide more useful information `$lastReleaseTag-$currentBranch-$commitsAheadOfLastRelease-$currentSha`, if the current directory has uncommitted changes `-dirty` is appended. An example here `1.3.0-win-make-51-gd8fe8ad-dirty` (thanks @wadey for putting this together)
- `Makefile` should work on Windows powershell and command prompt assuming you have `gnu make`, `awk`, `cut`, and `git` installed.
- `make bin` will output binaries with the `.exe` suffix
- Service mode logs are routed to the windows event log system instead of nowhere.

For the `make` dependencies I used `choco` to install `awk`, `make`, and `gnuwin32-coreutils.install`. Also added `%PROGRAMFILES(X86)%\GnuWin32\bin` to `%PATH%`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 GOMINVERSION = 1.16
 NEBULA_CMD_PATH = "./cmd/nebula"
 NEBULA_CMD_SUFFIX =
-BUILD_NUMBER ?= $(shell git describe --abbrev=0 --match "v*" | cut -dv -f2)-$(shell git branch --show-current)-$(shell git describe --dirty | cut -d- -f2-)
+BUILD_NUMBER ?= $(shell git describe --abbrev=0 --match "v*" | cut -dv -f2)-$(shell git branch --show-current)-$(shell git describe --long --dirty | cut -d- -f2-)
 
 ifeq ($(OS),Windows_NT)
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,12 +1,22 @@
 GOMINVERSION = 1.16
 NEBULA_CMD_PATH = "./cmd/nebula"
-BUILD_NUMBER ?= dev+$(shell date -u '+%Y%m%d%H%M%S')
+NEBULA_CMD_SUFFIX =
+BUILD_NUMBER ?= $(shell git describe --abbrev=0 --match "v*" | cut -dv -f2)-$(shell git branch --show-current)-$(shell git describe --dirty | cut -d- -f2-)
```
wadey, line 4:
```suggestion
BUILD_NUMBER ?= $(shell git describe --abbrev=0 --match "v*" | cut -dv -f2)-$(shell git branch --show-current)-$(shell git describe --long --dirty | cut -d- -f2-)
```

`--long` on the last describe makes it work better when you are building on the same git commit with the tag

wadey (approved):

PR #444, opened 2021-04-16
Replace promoteCounter with atomicMessageCounter

nbrownus:
`HostInfo.promoteCounter` and `HostInfo.ConnectionState.atomicMessageCounter` have merged into meaning basically the same thing.

`Interface.getOrHandshake` increments `HostInfo.promoteCounter` via `f.hostMap.PromoteBestQueryVpnIP`, which happens every time we try to send a message.

`Interface.sendNoMetrics` increments `HostInfo.ConnectionState.atomicMessageCounter`, which is actually sending a message.

This means we will attempt requery and promotion on the next message send instead of before, a minor difference.

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -40,7 +40,6 @@ type HostInfo struct {
 
 	remote            *udpAddr
 	remotes           *RemoteList
-	promoteCounter    uint32
 	ConnectionState   *ConnectionState
 	handshakeStart    time.Time        //todo: this an entry in the handshake manager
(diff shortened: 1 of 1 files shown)


nbrownus:
Going to put this back on the back burner for now. A side effect I hadn't considered is when we tun rx multiple packets on multiple queues, it creates a race. If more than 1 packet gets into the `TryPromoteBest` before `sendNoMetrics` increments the counter, and the counter is at a position to hit either promote or requery, multiple attempts will be made very quickly.

PR #447, opened 2021-04-23
Add locking around ssh conns to avoid concurrent map access on reload

nbrownus:
@briankelly hit this earlier today, I have been unable to reproduce but the panic doesn't lie.

### DIFF
--- a/sshd/server.go
+++ b/sshd/server.go
@@ -3,6 +3,7 @@ package sshd
 import (
 	"fmt"
 	"net"
+	"sync"
 
 	"github.com/armon/go-radix"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


rawdigits (dismissed):

sshd/server.go
```diff
@@ -151,11 +159,11 @@ func (s *SSHServer) Stop() {
 	}
 
 	// Force close all existing connections.
-	// TODO I believe this has a slight race if the listener has just accepted
```
brad-defined, line 154: FWIW, the 'slight' race I was referring to in this comment is different than the panic-inducing race this diff fixes.
- Accepting goroutine (ie, Run() method) accepts a new connection, and starts doing work.
- Concurrently, closing goroutine closes the listener.
- Closing goroutine next attempts to close all connections, obtains lock.
- Closing goroutine closes all current connections, releases lock
- Accepting goroutine continues doing work, and adds the new connection to existing connections s.conns map
- Accepting goroutine detects error on listener, and exits.
I think the worst outcome of this is that a connection that we intended to close stays open on the old SSHServer.

brad-defined (approved):

rawdigits (approved):

PR #448, opened 2021-04-23
Fully close tunnels when CloseAllTunnels is called

nbrownus:
@JohnMaguire noticed that when his mobile device sleeps, and all tunnels are closed (a beta feature), the remote end will try to handshake and enter race avoidance because we haven't properly torn down the tunnels.

To be clear, this only affects mobile clients.


wadey (approved):
Looks good, and seems like a safe change to me :+1:

PR #449, opened 2021-04-26
Get out faster on nil udpAddr

nbrownus:
We had a crash in mobile land related to calling `udpAddr.String()` on a nil address, this addresses other possible crashes.

### DIFF
--- a/udp_all.go
+++ b/udp_all.go
@@ -33,14 +33,26 @@ func (ua *udpAddr) Equals(t *udpAddr) bool {
 }
 
 func (ua *udpAddr) String() string {
+	if ua == nil {
+		return "<nil>"
+	}
+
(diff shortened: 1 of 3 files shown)


udp_all.go
```diff
@@ -33,14 +33,26 @@ func (ua *udpAddr) Equals(t *udpAddr) bool {
 }
 
 func (ua *udpAddr) String() string {
+	if ua == nil {
+		return ""
+	}
```
wadey, line 38:
```suggestion
	if ua == nil {
		return "<nil>"
	}
```

I think we should match what the stdlib does for similar things here, so it isn't confusing having a blank string in logs for this error case: https://github.com/golang/go/blob/go1.16.3/src/net/ip.go#L294-L296

wadey (approved):
This looks sane and good to me :+1:

PR #450, opened 2021-04-27
Add more metrics

wadey:
This change adds the following counter metrics:

Metrics to track packets dropped at the firewall:

    firewall.{incoming,outgoing}.dropped.local_ip
    firewall.{incoming,outgoing}.dropped.remote_ip
    firewall.{incoming,outgoing}.dropped.no_rule

Metrics to track handshakes attempts that have been initiated and ones
that have timed out (ones that have completed are tracked by the
existing "handshakes" histogram).

    handshake_manager.initiated
    handshake_manager.timed_out

Metrics to track when cached_packets are dropped because we run out of
buffer space, and how many are sent once the handshake completes.

    hostinfo.cached_packets.dropped
    hostinfo.cached_packets.sent

This change also notes how many cached packets we have when we log the
final "Handshake received" message for either stage1 for stage2.

### DIFF
--- a/firewall.go
+++ b/firewall.go
@@ -68,13 +68,20 @@ type Firewall struct {
 	rules        string
 	rulesVersion uint16
 
-	trackTCPRTT  bool
-	metricTCPRTT metrics.Histogram
-	l            *logrus.Logger
+	trackTCPRTT     bool
+	metricTCPRTT    metrics.Histogram
+	incomingMetrics *firewallMetrics
+	outgoingMetrics *firewallMetrics
 
-	metricDroppedRemoteIP metrics.Counter
-	metricDroppedLocalIP  metrics.Counter
-	metricDroppedNoRule   metrics.Counter
+	l *logrus.Logger
+}
+
+type firewallMetrics struct {
+	droppedLocalIP   metrics.Counter
+	droppedRemoteIP  metrics.Counter
+	droppedNoRule    metrics.Counter
+	allowedConntrack metrics.Counter
+	allowedRule      metrics.Counter
 }
 
 type FirewallConntrack struct {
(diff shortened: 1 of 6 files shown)


firewall.go
```diff
@@ -195,8 +199,12 @@ func NewFirewall(l *logrus.Logger, tcpTimeout, UDPTimeout, defaultTimeout time.D
 		UDPTimeout:     UDPTimeout,
 		DefaultTimeout: defaultTimeout,
 		localIps:       localIps,
-		metricTCPRTT:   metrics.GetOrRegisterHistogram("network.tcp.rtt", nil, metrics.NewExpDecaySample(1028, 0.015)),
 		l:              l,
+
+		metricTCPRTT:          metrics.GetOrRegisterHistogram("network.tcp.rtt", nil, metrics.NewExpDecaySample(1028, 0.015)),
+		metricDroppedLocalIP:  metrics.GetOrRegisterCounter("firewall.dropped.local_ip", nil),
```
nbrownus, line 205: I think it would make more sense to orient these on direction then dropped reason `firewall.{incoming, outgoing}.dropped.local_ip`, `firewall.{incoming, outgoing}.dropped.remote_ip`, `firewall.{incoming, outgoing}.dropped.no_rule`

Also seems like it opens us up to `firewall.{incoming, outgoing}.passed.conntrack` and `firewall.{incoming, outgoing}.passed.rule`

firewall.go
```diff
@@ -71,6 +71,10 @@ type Firewall struct {
 	trackTCPRTT  bool
 	metricTCPRTT metrics.Histogram
```
nbrownus, line 72: minor nit, move this one next to its friends :)

nbrownus (approved):
LGTM

PR #451, opened 2021-04-27
Dont apply race avoidance to existing handshakes, use the handshake time to determine who wins

nbrownus:
We used to check the main hostmap at stage 1 for an existing tunnel and apply race avoidance logic to it. This causes a lot of pain for hosts that were unable to transmit a `close_tunnel` packet but declared the tunnel dead locally (through roaming, unclean shutdown, network loss).

What happens now is that we track the `Time` field transmitted in handshake packet on the hostinfo object, this field has always been present and populated. If we get into the situation described above we will make sure the new handshake is _actually_ newer than the one we used before. If it is, we replace the tunnel. If it wasn't, we enter the same old flow of testing the tunnel. I believe the only real world case here are an incredibly high latency link or an attacker cached a handshake packet and retransmitted it at a later time, both which are handled.

Most notably, any client pre v1.4 will transmit the time value it received at stage 1 back on stage 2. If that responders clock is way off from the initiator and they get into the flow described at the top _as an initiator_ they may enter the test tunnel flow, which is identical to how the issue would be handled in v1.3 and before.

TLDR: This makes handshakes *much* faster in some error conditions

### DIFF
--- a/handshake_ix.go
+++ b/handshake_ix.go
@@ -211,7 +211,7 @@ func ixHandshakeStage1(f *Interface, addr *udpAddr, packet []byte, h *Header) {
 			f.l.WithField("vpnIp", IntIp(vpnIP)).WithField("udpAddr", addr).
 				WithField("certName", certName).
 				WithField("oldHandshakeTime", existing.lastHandshakeTime).
-				WithField("newHandshakeTime", existing.lastHandshakeTime).
+				WithField("newHandshakeTime", hostinfo.lastHandshakeTime).
 				WithField("fingerprint", fingerprint).
 				WithField("initiatorIndex", hs.Details.InitiatorIndex).WithField("responderIndex", hs.Details.ResponderIndex).
 				WithField("remoteIndex", h.RemoteIndex).WithField("handshake", m{"stage": 1, "style": "ix_psk0"}).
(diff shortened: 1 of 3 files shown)


handshake_ix.go
```diff
@@ -204,18 +207,15 @@ func ixHandshakeStage1(f *Interface, addr *udpAddr, packet []byte, h *Header) {
...
-			// handshake avoidance
-
-			//TODO: sprinkle the new protobuf stuff in here, send a reply to get the recv_errors flowing
-			//TODO: if not new send a test packet like old
-
+			// This means there was an existing tunnel and this handshake was older than the one we are currently based on
 			f.l.WithField("vpnIp", IntIp(vpnIP)).WithField("udpAddr", addr).
 				WithField("certName", certName).
+				WithField("oldHandshakeTime", existing.lastHandshakeTime).
+				WithField("newHandshakeTime", existing.lastHandshakeTime).
```
wadey, line 214:
```suggestion
				WithField("oldHandshakeTime", existing.lastHandshakeTime).
				WithField("newHandshakeTime", hostinfo.lastHandshakeTime).
```

wadey (commented):
A small issue with the logging, but otherwise I like this change. I think your analysis is correct about the different scenarios, and I think this change will help a bunch to avoid cases that would previously get stuck in handshake avoidance. :+1:

handshake_ix.go
nbrownus: Ah crap, nice catch!

wadey (approved):

PR #453, opened 2021-05-01
Increase the timestamp resolution for handshakes

nbrownus:
#451 introduced a way to more quickly resolve tunnel replacement but the end 2 end tests end up being racy since the current handshake resolution is 1 second and the communication speed there is very fast.

The real world impact of this change is to make tunnel replacement faster on fast networks. Pre v1.4 vs a v1.4 is basically the v1.4 host will always win, even if it shouldn't, but the v1.3 host will always lose in some very specific situations. The fallback mechanism is the same for v1.3 and before, race avoidance.


wadey (approved):
After thinking about this a bit, I'm pretty sure this will be backwards compatible with 1.3 hosts in the same network.

PR #456, opened 2021-05-02
Added generic systemd units supporting multiple networks

gboddin:
I think `network-online.target` is supported on most distributions, not sure about `network.target`.


gboddin:
Probably needs to be moved to https://github.com/slackhq/nebula/tree/master/examples/service_scripts , I'll followup on this :thinking:

tarrenj:
Also see #583.

nbrownus:
Thanks for the PR, looks like the most recent version of our example systemd unit has `network-online.target` currently.

PR #459, opened 2021-05-05
Unlock the hostmap quickly, lock hostinfo instead

nbrownus:


wadey (approved):
This looks good and makes sense to me.

PR #460, opened 2021-05-07
iputil package

nbrownus:
This makes a new `iputil` package with a type that speaks to both storage and logging of uint32 based ip addresses. It is also the start of making a v4/v6 `vpnIp` type. Some movement from `vpnIP`/`VpnIP` -> `vpnIp`/`VpnIp` as well.

This is part of moving `tun` and `udp` files into their own packages.

The target is `master-next` so that I can avoid a squash merge/rebase nonsense.


wadey (approved):
I'm not a huge fan of the name "iputil", but I don't have a better name right now. So I guess lets not bikeshed it too much and move forward (we can always refactor and rename it again if we want).

nbrownus:
Discussed in slack, we may change `iputil.VpnIp` to `overlay.IP` in a future PR pending a discussion about udp and vpn ip unification.

PR #461, opened 2021-05-07
Rename vpn ip

nbrownus:
Builds upon #460 to further normalize vpn ip names


wadey (approved):
I prefer the stdlib style of "IP" instead of "Ip" but I don't want to be a blocker here.

nbrownus:
😅 since we had a mix, I picked the logging convention. But also for the exported struct, `VPNIP`? I don't have a razor for why `VpnIP` makes logical sense but it looks fine to me.

PR #472, opened 2021-05-16
Updated vagrant env, more automation

aarzhanov:
- Added new box
- Auto install vagrant plugins
- Ansible deployment during the execution of the vagrant up.
- Improved README (new workflow, check go version)
- new .gitignore


nbrownus:
Thanks for the PR but we have actually removed the vagrant quick start guide in this repo and instead point folks to a docs website for this. https://nebula.defined.net/docs/guides/quick-start/

PR #473, opened 2021-05-16
Filter DNS queries by zone

aaronkollasch:
Adds config options `lighthouse.dns.zones`, a list of DNS zones the server has an answer for, and `lighthouse.dns.drop_filtered`, which specifies whether to reply to queries outside of those zones.


nbrownus:

PR #474, opened 2021-05-18
Add an ability to specify metric for unsafe routes

ton31337:
This is needed when using multiple routes to the same destination.

My use case is that I'm connected via OpenVPN and Nebula and both are
using for a default route (kinda HA in case).

```
% grep -A2 0.0.0.0/1 /opt/nebula/config.yml
    - route: 0.0.0.0/1
      via: 172.16.14.184
      metric: 12312321
```

```
% ip route show 0.0.0.0/1
0.0.0.0/1 dev nebula1 scope link metric 12312321 mtu 1300
```


wadey (dismissed):
Approved, but we need to write a CHANGELOG entry.

ton31337:
Will update.

ton31337:
> Approved, but we need to write a CHANGELOG entry.

Updated.

nbrownus (approved):

PR #481, opened 2021-05-31
Create SECURITY.md

mhowell11:
I am a security researcher on HackerOne and when I was testing some vulnerabilities according to the HackerOne guidelines for Slack, I came here and noticed that Nebula does not have a Security Policy. I am aware that this is not a Tier 0 vulnerability but I wanted to contribute a draft of one anyways to help the security of Nebula. Thank you for your consideration.


hlovdal:
Notice there is also [this pull request](https://github.com/slackhq/nebula/pull/263).

wadey:
Thanks for your pull request! We are going to merge a simpler one instead that defers to the HackerOne page which will be updated more frequently.

Duplicate of #864

PR #482, opened 2021-05-31
fixed recv_errors spoofing condition

AndrewChubatiuk:
Hi @nbrownus 
Fixed a small bug that was introduced in 
https://github.com/slackhq/nebula/commit/df7c7eec4a39bd0eb73a66a4e9d38e70efae721e#diff-5d05d02296a1953fd5fbcb3f4ab486bc5f7c34b14c3bdedb068008ec8ff5beb4
having problems due to it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


wadey (dismissed):
Good catch, and thanks for the PR :+1:

wadey (approved):

nbrownus:
🤦 nice catch!

PR #483, opened 2021-06-03
Fix single command ssh exec, add a `raw` flag to `print-cert`

nbrownus:
Before, if you tried to use an ssh client to execute a single command (not on the command line) it would hang as the session would not close properly, this fixes that and adds a way to get a tunnels cert as a PEM.


wadey (approved):
This looks correct to me 👍

PR #497, opened 2021-07-15
WIP: multi source port

demmer:
## Description
WIP: send outbound packets using one of N bound source ports

## Motivation
The goal of this work is to send packets between two hosts using more than one 5-tuple. When running on networks like AWS where the underlying network driver and overlay fabric makes routing, load balancing, and failover decisions based on the flow hash, this enables more than one flow between pairs of hosts.

## Proposal
With this change, Nebula will optionally bind to multiple additional UDP sockets that are only used for sending, not for receiving. 

Add a new optional configuration "send.ports" which specifies how many ports to use per thread for sending (default is 1). Notably, the destination port is always fixed for a given host.

If this is set to a value > 1, then nebula binds to additional sockets and then picks the one to send based on a "hash" of source of the src and dst ports. This is deliberately just a random scrambling of the bits, I'm sure there's something better we could do.

~After some internal testing revealed that the original commit triggered the roaming behavior, I also added a modification to `hostDidRoam` so that it ignores the ports and only pays attention to the IP addresses.~

After some _more_ discussion I thought of a different approach: Tag packets sent from the "non-canonical" interface, i.e. using a source port that the sender is not listening on, with a subtype that indicates to the receiver that it should ignore  packet for roaming detection.

## Caveats
This PR is the first time I've ever touched the nebula code, I have only a cursory understanding of the implications of this change on the handshaking / roaming / etc. 

~In particular, I have absolutely 0 understanding of roaming or what my change does or would do.~
I now kinda maybe get this a bit.

### DIFF
--- a/firewall.go
+++ b/firewall.go
@@ -165,6 +165,21 @@ func (fp FirewallPacket) MarshalJSON() ([]byte, error) {
 	})
 }
 
+const (
+	prime1 uint16 = 173
+	prime2 uint16 = 3313
+	prime3 uint16 = 5387
+)
+
+func (fp FirewallPacket) HashPorts(numBuckets int) int {
+	if numBuckets <= 1 {
+		return 0
+	}
+
+	// super lame but fast mixing
+	return int((fp.LocalPort^prime1)+(fp.RemotePort^prime2)+prime3) % numBuckets
+}
+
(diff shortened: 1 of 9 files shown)


wadey:
We reworked this PR into #768, thus I'm closing this older one.

Issue #493, opened 2021-07-01
Connecting to device's local nebula IP


gilby:
I installed nebula on 3 FreeBSD servers using the port and also on MacBook, and can reach every server from all them, but can't ping or connect to the IP address on the device that IP is on. Is that the expected behavior, just a *BSD thing, or do I have my configuration wrong?

I originally thought using nebula would open up the local network interfaces to the devices on the nebula network, but I can't access those from another server. Was I incorrect on that being possible?

The main issue is that many services only bind to one IP address or all IP addresses, so binding it to the nebula IP means I can't access the server from the server it's running on. I'll get around it by firewalling the access to it so it's only accessible on the nebula and intranet.

caguiclajmg:
> I originally thought using nebula would open up the local network interfaces to the devices on the nebula network, but I can't access those from another server. Was I incorrect on that being possible?

Are you referring to the IP addresses of the hosts that are on a different network interface? No, this isn't a common feature in VPN software

Technically, it is possible, but I would highly suggest against it. If you really want to do this, you can set up static routes on every client that you need to connect from and then use the target's nebula IP as the gateway to reach other interfaces.

> The main issue is that many services only bind to one IP address or all IP addresses, so binding it to the nebula IP means I can't access the server from the server it's running on. I'll get around it by firewalling the access to it so it's only accessible on the nebula and intranet.

Unless you're using software from the stone age, this shouldn't be a problem.

A lot of services allow binding to multiple specific addresses (e.g. in OpenSSH you can specify multiple `ListenAddress` directives).

gilby:
Yes, I was referring to the machine's private address in that datacenter.

The software from the stone age is MariaDB. Looks like MySQL got out of the stone age a couple years ago and allows it, though we're not switching to that.

The main issue remains though and that is of connecting to services on the host's nebula IP.

caguiclajmg:
You still have the option of binding to all interfaces and just firewalling off the addresses you don't want people to see.

If you don't want that, there still are a couple of sane options.

1. NAT a port on your nebula IP to the port on the private interface
2. Set up a tunnel (e.g. frp, ssh)

ALDemonPit:
Nothing

brad-defined:
[...]

PR #501, opened 2021-07-20
Immediately forward packets received on the nebula TUN device from self to self

brad-defined:
PR #192 updated Nebula to immediately drop packets read by the Nebula TUN device and destined to the local Nebula VPN IP.

On Linux machines, the local route table will send packets destined to the local nebula VPN IP through the loopback device, bypassing Nebula all together and permitting the packets to flow. On my mac, the route table instead sends these packets through the Nebula TUN device, where they are immediately dropped.

On my linux lighthouse - VPN IP routes to loopback device:
```
~# ip route get 172.16.200.1
local 172.16.200.1 dev lo src 172.16.200.1 uid 0 
    cache <local> 
```
On my mac - VPN IP routes to utun0, which is my Nebula TUN device:
```
nebula % route get 172.16.200.3
   route to: 172.16.200.3
destination: 172.16.200.3
  interface: utun0
      flags: <UP,HOST,DONE,LOCAL>
 recvpipe  sendpipe  ssthresh  rtt,msec    rttvar  hopcount      mtu     expire
       0         0         0         0         0         0      1300         0 
```

This diff updates Nebula to immediately forward these types of packets right back out the TUN device. This change will make my Mac (and presumably every other machine that routes the way Mac does) behave the same as Linux machines.

I also considered an alternative approach of having the locally destined packet be processed by the rest of the Nebula stack. This alternative approach would ensure the Nebula firewall rules are applied, for instance. However, this alternative approach would not behave the same as Linux (which does not apply the Nebula firewall rules to these local connections, as they bypass Nebula entirely), and would also require Nebula to handshake with itself locally - something which is not permitted elsewhere in the code for different reasons.

I also tried to address the issue by manually adding a route for my Nebula IP to route through the loopback device. This approach didn't reliably fix the issue for me - sometimes it worked, sometimes my manual route rule was applied after Nebula's route rules, or for some reason still sent my packets to the Nebula TUN device.

This PR fixes issue #493


inside.go
```diff
@@ -19,8 +19,12 @@ func (f *Interface) consumeInsidePacket(packet []byte, fwPacket *FirewallPacket,
 		return
 	}
 
-	// Ignore packets from self to self
+	// Immediately forward packets from self to self
 	if fwPacket.RemoteIP == f.myVpnIp {
+		_, err := f.readers[q].Write(packet)
```
nbrownus, line 24: In group discussion, we think this should be limited to macos only and leave the old behavior for all other oss

brad-defined: Wrote it up with a const, in the hopes that a const gives the compiler the best opportunity to optimize the code path as it sees fit. Happy to implement the o/s specific path other ways if desired.

brad-defined:
Also fixes issue #38

PR #503, opened 2021-07-22
Check CA cert and key match in nebula-cert sign

JohnMaguire:
`func (nc *NebulaCertificate) VerifyPrivateKey(key []byte) error` would
previously return an error even if passed the correct private key for a
CA certificate `nc`.

That function has been updated to support CA certificates, and
nebula-cert now calls it before signing a new certificate. Previously,
it would perform all constraint checks against the CA certificate
provided, take a SHA256 fingerprint of the provided certificate, insert
it into the new node certificate, and then finally sign it with the
mismatching private key provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - SSH server handles single `exec` requests correctly. (#483)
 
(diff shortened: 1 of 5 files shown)


wadey (approved):
Just ran into this same issue, and I'm glad to see there is already a PR!

wadey (approved):

Issue #142, opened 2019-12-28
In the example config.yml it mentions defining subnets in a node's certificate - how is this done?


bmullan:
I read a line in the **example config.yml** file that says:

> \# Unsafe routes allows you to route traffic over nebula to non-nebula nodes
> \# Unsafe routes should be avoided unless you have hosts/services that cannot run nebula
> \# **NOTE: The nebula certificate of the "via" node *MUST* have the "route" defined as a subnet in its certificate**
>   unsafe_routes:
>     \#- route: 172.16.1.0/24
>     \#  via: 192.168.100.99
>     \#  mtu: 1300 #mtu will default to tun mtu if this option is not sepcified

How is this done?

harpchad:
With nebula-cert when you create the cert for that node, i.e.:
```
nebula-cert sign -name mynode -ip 192.168.255.1/24 -out-crt ./mynode.crt -out-key ./mynode.key -subnets 192.168.254.1/24,192.168.255.1/24 -groups red,blue
```

lightngn:
Comment,
Should your command read as follows in the subnet section?  -subnets 192.168.254.0/24,192.168.255.0/24
***Like this***

> nebula-cert sign -name mynode -ip 192.168.255.1/24 -out-crt ./mynode.crt -out-key ./mynode.key -subnets 192.168.254.0/24,192.168.255.0/24 -groups red,blue

bmullan:
@harpchad 
@lightngn asked if the command example you gave was correct or not?  Can you look at his question and say which is correct?  
Thanks

lightngn:
What are your subnets that you are trying to add to unsafe_routes?

bmullan:
yes
but his question was

> Should your command read as follows in the subnet section? -subnets 192.168.254.0/24,192.168.255.0/24
> Like this
> 
>     nebula-cert sign -name mynode -ip 192.168.255.1/24 -out-crt ./mynode.crt -out-key ./mynode.key -subnets 192.168.254.0/24,192.168.255.0/24 -groups red,blue

which differs from your example by leaving the last octet "0" instead of "1"

harpchad:
@lightngn 
Yes, the subnets should be the network number .0, typed out an example too quickly, thanks for catching that.

JonTheNiceGuy:
Follow-up question.

Given this `nebula-cert` command:

[...]

PR #507, opened 2021-08-03
Add destination CIDR checking

jilyaluk:
Currently, if a node has unsafe_routes behind it, they can't be firewalled granularly, e.g. opening tcp/443 port will open it for all hosts.
This PR adds new config option, `local_cidr` which is interpreted the same way as `cidr`, but packet's local IP is checked instead of remote IP.

I know that you are not planning to implement general-purpose firewall as per [this comment](https://github.com/slackhq/nebula/issues/142#issuecomment-570250934), however offloading unsafe_route firewalling to iptables will inevitably lose some perks like checking node groups (or require some additional automation to keep them in sync). This commit however does not add any significant performance penalties (local_cidr is checked last, hence it will only impact packets that will be dropped otherwise) and has pretty simple implementation.

cc @nbrownus @wadey

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -246,7 +246,8 @@ firewall:
   #   host: `any` or a literal hostname, ie `test-host`
   #   group: `any` or a literal group name, ie `default-group`
   #   groups: Same as group but accepts a list of values. Multiple values are AND'd together and a certificate would have to contain all groups to pass
-  #   cidr: a CIDR, `0.0.0.0/0` is any.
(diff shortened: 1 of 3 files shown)


jilyaluk:
Rebased onto latest master

jstasiak:
Hey, this is *exactly* what I was looking for and an important feature for me (I'd like to avoid dealing with OS-specific firewalls on devices where I run Nebula).

The PR looks really nice and it would be great to see it merged.

wadey (approved):

PR #535, opened 2021-09-24
create ConnectionState before adding to HostMap

wadey:
We have a few small race conditions with creating the HostInfo.ConnectionState
since we add the host info to the pendingHostMap before we set this
field. We can make everything a lot easier if we just add an "init"
function so that we can set this field in the hostinfo before we add it
to the hostmap.

We have seen this issue cause crashes in v1.4.0.


handshake_manager.go
```diff
@@ -191,13 +191,13 @@ func (c *HandshakeManager) handleOutbound(vpnIp iputil.VpnIp, f udp.EncWriter, l
 	}
 }
 
-func (c *HandshakeManager) AddVpnIp(vpnIp iputil.VpnIp) *HostInfo {
-	hostinfo := c.pendingHostMap.AddVpnIp(vpnIp)
-	// We lock here and use an array to insert items to prevent locking the
```
nbrownus, line 196: I think if we remove this then we should add the locking or rename the function to `UnlockedAddVpnIp`

handshake_manager_test.go
```diff
@@ -71,7 +71,7 @@ func Test_NewHandshakeManagerTrigger(t *testing.T) {
 
 	assert.Equal(t, 0, testCountTimerWheelEntries(blah.OutboundHandshakeTimer))
 
-	hi := blah.AddVpnIp(ip)
+	hi := blah.AddVpnIp(ip, nil)
```
nbrownus, line 74: Lets add a test case to ensure our init func is called and the created return bool is correct

handshake_manager.go
wadey: This comment is outdated and from before the lock was moved to inside of the OutboundHandshakeTimer itself. Now we can finally delete this old and wrong comment!

handshake_manager_test.go
wadey: Test added higher up in the file in latest commit

nbrownus (approved):
Do we want to note this one in the changelog?

wadey:
Added CHANGELOG

nbrownus (approved):

PR #537, opened 2021-09-25
Allow embedding nodes into other services/applications

Zilog8:
In certain cases, it may not desirable or even possible to install a tun device to connect to a Nebula network. A neat alternative would be to allow embedding of the Nebula service into applications or other services, which would then have individual access to resources on a Nebula network. This changeset includes:
-A fake Tun to handle packets in/out.
-Minor modifications to allow providing a Tun when starting Nebula.
-An example application that sends and receives packets directly through Nebula, without a real Tun device.

### DIFF
--- a/cert.go
+++ b/cert.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


johnmaguire:
Closing as this is stale, and I believe the goal is achieved by #965 which merged and is available since v1.8.0.

PR #547, opened 2021-10-12
Convert uint32 to netaddr.IP in most of the code

ton31337:
Basically, just a preparation for IPv6 overlay support.


ton31337:
Good to know there is a progress towards that. Thanks.

PR #550, opened 2021-10-20
Add a context object in nebula.Main to clean up on error

brad-defined:
This PR makes nebula's Main() method re-callable should it return an error on first invocation.

* Create a context object in nebula.Main, and pass it around to various goroutines, so that they can shut
down on failed Main invocations, as well as when control.Stop() is called.
* Plus cleanup from john@defined.net to ensure Punchy and LhWorker run immediately on startup.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,6 +1,7 @@
 package nebula
 
 import (
+	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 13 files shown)


wadey (approved):
LGTM :+1:

PR #552, opened 2021-10-21
Break apart file loading from Config struct

kriive:
I'm embedding nebula in my binary and I'm passing the configuration through LoadString(), so I'm not passing through the usual config file. When using LoadString() there's no way to Reload a certificate, this PR tries to correct that separating the actual Reload() from ReloadConfig(). Function names may not be the clearest, so I'm super open to suggestions.

### DIFF
--- a/cmd/nebula-service/main.go
+++ b/cmd/nebula-service/main.go
@@ -52,9 +52,16 @@ func main() {
 	l.Out = os.Stdout
 
 	c := config.NewC(l)
-	err := c.Load(*configPath)
+
+	// read files from config path and store read files in configFiles string slice
+	configFiles, err := config.ReadConfigFiles(*configPath)
 	if err != nil {
-		fmt.Printf("failed to load config: %s", err)
+		fmt.Printf("failed to read config(s): %s", err)
+		os.Exit(1)
+	}
+
+	if err := c.Load(configFiles...); err != nil {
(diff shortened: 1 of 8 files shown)


kriive:
I finally found some time (I got the flu, so plenty of time haha) to implement this. I broke apart file reading and loading from the Config struct, so Load and Reload are now variadic and take one or more config as strings. 

I added some tests (which I now see I'll have to fix, since they're failing) to cover sorting and some simple cases.

To handle SIGHUP I implemented what I believe is an ugly hack I'm not really fond of. I believe we could implement this better if `Main` accepted a `ctx` object initialized in the various `cmd/` `main()`s. 

For example:
```go
func Main(ctx context.Context, c *config.C, configTest bool, buildVersion string, logger *logrus.Logger, tunFd *int) (retcon *Control, reterr error) {
	ctx, cancel := context.WithCancel(ctx)
	// Automatically cancel the context if Main returns an error, to signal all created goroutines to quit.
	defer func() {
		if reterr != nil {
			cancel()
		}
	}()
	// ...
```
At this point we could handle SIGHUPs directly in the package main. What do you think?

kriive:
Any news on this?

nbrownus:
This is an area we are focused on currently and hope to have a complete answer for by the v1.10 release. Our current approach is taking a different path though so I am closing this PR.

PR #561, opened 2021-10-28
Add an ability to specify allow from list

ton31337:
Allow from `allow_from` instead of only static hosts. Also, 127.0.0.1 is the wrong
source, because 127.0.0.0/8 MUST be allowed. It's a local range.


ton31337:
@nbrownus @wadey could you take a look here? If not, it can be closed.

ton31337:
ping

PR #563, opened 2021-11-02
Add link to further documentation

nbrownus:


wadey (approved):
More docs are never a bad thing :+1:

PR #566, opened 2021-11-04
Fix race between punchback and lighthouse handler reset

nbrownus:
There is a race between `lhh.HandlerRequest` doing a `lhh.resetMeta` and `lhh.handleHostPunchNotification` when punch back is enabled. This fixes the immediate issue but the note about using a timer wheel still applies.

This also fixes an opportunity for punching back to the wrong host, depending on how many `NebulaMeta_HostPunchNotification` messages a node is receiving.


wadey (approved):
LGTM 👍

PR #571, opened 2021-11-08
Darwin signing, notarizing, and universal binaries

### LINKED ISSUE #24: macOS Release Binaries Aren't Signed

nbrownus:
This PR does a few things:
1. Creates universal binaries that work on both `amd64` and `arm64` for darwin.
2. Signs the binaries using key material owned by defined.net
3. Notarizes the resulting zip file using an Apple account owned by defined.net
4. Modernizes the action plugins used in the github release action

This will break existing automations as we no longer publish architecture specific binaries for darwin, only a single `nebula-darwin.zip` will be available going forward.

Interestingly, you still can not execute these binaries in `Finder.app` but it should not complain if you execute them in a terminal.

Closes #24, closes #29

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,7 @@ jobs:
           mv build/*.tar.gz release
 
       - name: Upload artifacts
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v2
         with:
           name: linux-latest
           path: release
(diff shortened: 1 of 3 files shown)


wadey (approved):
This is will break some scripts for downloading the latest version. But I think that is worth it to get signed fat binaries :+1:

jilyaluk:
Hey there,

Thanks for adding notarisation support!

I worked on a similar issue of signing Mac binaries. Even though for some reason zipping is required before uploading the binary for notarisation, it could actually be distributed as a binary (not archived). It seems that Apple stores all binaries from .zip separately and Gatekeeper correctly validates signature in this case.

nbrownus:
That's good to know but since we ship 2 binaries per os and architecture (`nebula` and `nebula-cert`) we need some way to package them. Previously we used `tar.gz` and it felt silly to move between that and `zip`, certainly given that the use of universal binaries changed the name of the package we ship for Darwin (`nebula-darwin-$arch.tar.gz` -> `nebula-darwin.zip`).

PR #575, opened 2021-11-11
Move util to test, contextual errors to util

nbrownus:
This is part of a larger restructuring that will give us an `overlay` package containing tun and routing logic

### DIFF
--- a/allow_list_test.go
+++ b/allow_list_test.go
@@ -7,12 +7,12 @@ import (
 
 	"github.com/slackhq/nebula/cidr"
 	"github.com/slackhq/nebula/config"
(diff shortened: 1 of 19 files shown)


test/main.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func NewTestLogger() *logrus.Logger {
```
wadey, line 10: should this file be `test/logger.go` instead?

nbrownus: weird, it is `test/logger.go` 😉

wadey (approved):

Issue #148, opened 2020-01-02
Integrate PCP and NAT-PMP 


rawdigits:
I can't find evidence this is in wide use, but worth supporting these, as it may help folks behind NAT.

SlyngDK:
I have created a PR on this, you can try testing it https://github.com/slackhq/nebula/pull/579

I am running a test setup with it and works fine with miniupnpd on the gateway.

Config needed:
```
punchy:
  port_mappings.pcp:
    enabled: true
```

comex:
For what it's worth, my Xfinity-provided home router:
- Has a 'port restricted cone' NAT, i.e. sending a UDP packet to a given IP and port only allows responses from that specific IP and port.  Thus UDP hole punching is not possible.
- Does not support PCP or NAT-PMP, as far as I can tell.
- *Does* support UPnP IGD.

So there are common setups where hole punching does not work but automated port forwarding does work (though not specifically PCP or NAT-PMP).  I can make do with manual port forwarding, but it would be nice if Nebula 'just worked' in this situation.

PR #579, opened 2021-11-12
Port mapping PCP

SlyngDK:
I have been working on support port mapping/forward using PCP.

It will be nice with a review, of what is missing? (I am aware of test is missing, and will do if your considering this PR)

Your are also welcome to change and/or copy and do your own implementation.


tarrenj:
PCP is a new concept to me so this might be an irrelevant question, but how does this play with UPnP environments (i.e., running a Nebula node behind a router with UPnP configured)

SlyngDK:
@tarrenj 
PCP is support by miniupnpd, which is used in many routers, but it depends on the manifactor to enable PCP.

I have only tested this with miniupnpd on a linux server running as firewall/gateway.

For info check this https://en.wikipedia.org/wiki/Port_Control_Protocol

SlyngDK:
@nbrownus Do you have time to discourse/look at this?

thejinx0r:
@SlyngDK 
Does it log if it's enabled?
I tried it out, and it doesn't seem to enable it on my mikrotik router.

Maybe I understand it wrong, but is pcp a subset of upnp?

Anyways, here's abit more info:
I don't use go, so I'm not sure if I built it correctly. It says that the version was `build=1.4.0-port-mapping-pcp-51-ga8d36fb`

I enabled it in the config following this comment:
https://github.com/slackhq/nebula/issues/148#issuecomment-979140490

SlyngDK:
@thejinx0r 
Yes, it's logging at this line https://github.com/slackhq/nebula/pull/579/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R401
You could try change logging to debug, and see what going on.
I have only test with miniupnp with `enable_natpmp=yes`, which enables natpmp and pcp.

thejinx0r:
@SlyngDK 
I managed to enable it in nebula. For some reason,  it didn't recognise ` port_mappings.pcp`
I had to separate it in to it's dictionary.

Afterwards, it logged that PCP was enabled. The debug logs shows that it was trying to enable it on my router.
My Mikrotik router does not support PCP, only UPnP. So while PCP is a subset of UPnP, I guess not everyone implements PCP.

PR #580, opened 2021-11-12
Update nebula.service

joshaspinall:
Add a delay to avoid startup issues on some systems.

### DIFF
--- a/examples/service_scripts/nebula.service
+++ b/examples/service_scripts/nebula.service
@@ -5,6 +5,7 @@ After=basic.target network.target
 Before=sshd.service
 
 [Service]
+ExecStartPre=/bin/sleep 5
 SyslogIdentifier=nebula
 ExecReload=/bin/kill -HUP $MAINPID
 ExecStart=/usr/local/bin/nebula -config /etc/nebula/config.yml


ton31337:
This sounds like a masking of a real problem. In what circumstances does it happen? I would say we should try to handle this in the code because it's only for systems using `systemd`.

joshaspinall:
I have a mix of systems across a number of locations running Ubuntu.  For reasons I am unsure of, sometimes the service does not start due to a 'start request repeated too quickly' error in the systemd service logs.  I can confirm this has resolved the issue wherever I have seen it.  Happy to propose this as a temporary workaround until such time there is a more elegant solution.

nbrownus:
Strong agree with @ton31337. `journalctl -u nebula` should give some hints as to what the problem is. You will likely find the correct fix is adjust your `Wants` and/or `After`. 

Another approach, which I would not recommend to do on its own, would be to handle this with `RestartSec`, and/or `StartLimitIntervalSec`, `StartLimitBurst` https://www.freedesktop.org/software/systemd/man/systemd.unit.html#StartLimitIntervalSec=interval

joshaspinall:
I will investigate the further options to fix 'properly' as suggested.  As the file currently here causes me an issue, would you wish for me to report back with a preferred change?

nbrownus:
I'm interested to see the target that makes it work for you. Feel free to re-open this PR with the changes if you feel they are appropriate for the community at large!

PR #581, opened 2021-11-12
Push route handling into overlay, a few more nits fixed

nbrownus:


overlay/tun_darwin.go
```diff
@@ -279,7 +282,12 @@ func (t *tun) Activate() error {
 	}
 
 	// Unsafe path routes
-	for _, r := range t.UnsafeRoutes {
+	for _, r := range t.Routes {
+		if r.Via == nil {
+			// We don't allow route MTUs so only install routes with a via
+			continue
```
wadey, line 288: should we log a warning for this case? That we are skipping an entry because `via` is missing?

wadey (approved):

overlay/tun_darwin.go
wadey: nevermind I see that newTun handles this.

PR #583, opened 2021-11-12
Add README and restart burst limit to service file

tarrenj:
Add a 30 second delay before restarting the service after a failure and
prevents additional restarts after 5 attempts.

Add some basic usage/installation instructions for the Systemd service
file to the README.md within the example service file directory.

Change `After=` to nss-lookup for when hostnames are in the nebula config file.

### DIFF
--- a/examples/service_scripts/README.md
+++ b/examples/service_scripts/README.md
@@ -1,0 +1,10 @@
+## Systemd:
+1. `# cp nebula.service /lib/systemd/system/`
+1. `# systemctl daemon-reload`
+1. Automatically start at boot (if desired):
+`# systemctl enable nebula.service`
+1. Start immediatly:
+`# systemctl start nebula.service`
+
+## Initd:
+##todo##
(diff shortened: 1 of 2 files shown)


tarrenj:
This is a continuation/resolution of #580.

tarrenj:
Also see #522, waiting for `nss-lookup.target` will ensure hostnames resolve before Nebula starts up.

johnmaguire:
I don't think we should recommend copying the systemd unit to `/lib` - that's where system-installed scripts go. User-installed ones belong in `/etc`. :)

I think also we do not need the TODO bit about initd in the README. I'd maybe eschew the README altogether as we really prefer people install from their system package manager if possible. For users who have other goals, we kind of expect they know how to install service files, or can figure it out.

The other systemd unit file changes are more interesting, and possibly worth considering. Maybe you can update the PR to just include those changes?

johnmaguire:
The most important changes seem to have already landed: https://github.com/slackhq/nebula/blob/master/examples/service_scripts/nebula.service

Issue #585, opened 2021-11-15
Darwin nebula client not finding/receiving udpAddrs for other hosts.


turtlewax:
My lighthouse most definitely knows and has a udpAddr for a remote host (Linux) and my mac host. Trouble is, my mac nebula host doesn't seem to be able to find out about the udpAddr for the other remote host (Linux).

Here is what I see on my lighthouse:
`Nov 15 10:26:14 hostname nebula[1478]: time="2021-11-15T10:26:14-05:00" level=info msg="Handshake message sent" certName=pc1 fingerprint=2733666f6007c2b508475b269fdb332a240343c49ba36b3973c5ffc31f999151 handshake="map[stage:2 style:ix_psk0]" initiatorIndex=1797597823 issuer=298e1986b9d3e1f20316a9b3389ebb037dcfe465932453e2b25c4f949e3e0217 remoteIndex=0 responderIndex=1842167964 sentCachedPackets=0 udpAddr="<redacted public IP>:48415" vpnIp=192.168.200.3`

And here is what my mac nebula client reports:
`INFO[0868] Handshake timed out durationNs=9382057076 handshake="map[stage:1 style:ix_psk0]" initiatorIndex=788921434 remoteIndex=0 udpAddrs="[]" vpnIp=192.168.200.3`

Now, the lighthouse log absolutely shows both of these separate hosts properly, with different, correct public udpAddrs assigned for each. The other remote host (192.168.200.3) is correctly reporting the udpAddr for the mac client, but the mac client never seems to get or find out about a udpAddr for the remote host (udpAddrs="[]", as seen above).

Why is this? Is this a bug in the Darwin nebula client that no one has reported yet?

nbrownus:
Have you confirmed that the mac client is connected to the same lighthouse?

turtlewax:
Yes, I only have a single lighthouse running.  The lighthouse itself has correct udpAddr and vpnIp values for both of the clients (Linux and Mac).  The Linux client does have a correct udpAddr value for the macOS client, but the macOS client has a perpetually empty udpAddr value for the Linux client.

Client configurations are basically identical, with the exception of certs and keys of course.

nbrownus:
Do you have either of these setup on any of your hosts? https://github.com/slackhq/nebula/blob/78d0d46baea578c9d6e02c2f0f137068bc269839/examples/config.yml#L52
[...]

PR #587, opened 2021-11-15
Warn if no lighthouses were configured on a non lighthouse node

nbrownus:
Would help in situations like #585


tarrenj:
This has also tripped me up by mistake. Maybe also add a check to ignore `static_host_map` if `amLighthouse`?

nbrownus:
> This has also tripped me up by mistake. Maybe also add a check to ignore `static_host_map` if `amLighthouse`?

It's ok to have static host entries on a lighthouse node, you can run an entire mesh without a lighthouse if you desire.

wadey (approved):

PR #590, opened 2021-11-18
don't set ConnectionState to nil

wadey:
We might have packets processing in another thread, so we can't safely
just set this to nil. Since we removed it from the hostmaps, the next
packets to process should start the handshake over again.

I believe this comment is outdated or incorrect, since the next
handshake will start over with a new HostInfo, I don't think there is
any way a counter reuse could happen:

> We must null the connectionstate or a counter reuse may happen

Here is a panic we saw that I think is related:

    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x93a037]
    goroutine 59 [running, locked to thread]:
    github.com/slackhq/nebula.(*Firewall).Drop(...)
            github.com/slackhq/nebula/firewall.go:380
    github.com/slackhq/nebula.(*Interface).consumeInsidePacket(...)
            github.com/slackhq/nebula/inside.go:59
    github.com/slackhq/nebula.(*Interface).listenIn(...)
            github.com/slackhq/nebula/interface.go:233
    created by github.com/slackhq/nebula.(*Interface).run
            github.com/slackhq/nebula/interface.go:191

### DIFF
--- a/outside.go
+++ b/outside.go
@@ -349,9 +349,8 @@ func (f *Interface) handleRecvError(addr *udp.Addr, h *header.H) {
 		return
 	}
 
-	// We delete this host from the main hostmap
-	f.hostMap.DeleteHostInfo(hostinfo)
-	// We also delete it from pending to allow for
+	f.closeTunnel(hostinfo, false)
+	// We also delete it from pending hostmap to allow for
 	// fast reconnect.
 	f.handshakeManager.DeleteHostInfo(hostinfo)
 }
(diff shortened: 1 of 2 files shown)


forfuncsake (dismissed):
I don't have a full mental model of all uses of the `ConnectionState` or `copyHostInfo`, but at a quick scan I didn't see anything obvious that could lead to counter reuse either.

Looks like it should be safer to not nil this out.

outside.go
```diff
@@ -352,9 +352,7 @@ func (f *Interface) handleRecvError(addr *udp.Addr, h *header.H) {
 	// We delete this host from the main hostmap
 	f.hostMap.DeleteHostInfo(hostinfo)
```
nbrownus, line 353: I think it would be more correct to call `f.closeTunnel(hostinfo, true)` since this is missing a few parts of removing a tunnel currently.

wadey: Good call, PR updated.

nbrownus (approved):

PR #592, opened 2021-11-23
Read MaxRemotes and roaming timer values from config file.

tarrenj:
Read `PromoteEvery`, `ReQueryEvery`, and RoamingSuppressSeconds`
from a new `dynamic_routing` block in the config file instead of
hardcoded constants in the `hostmap.go` file.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -128,6 +128,15 @@ punchy:
 # and has been deprecated for "preferred_ranges"
 #preferred_ranges: ["172.16.0.0/24"]
 
+dynamic_roaming:
+  #promote_every: 1000
+  #requery_every: 5000
+  # How long we should prevent roaminb back to the previous IP.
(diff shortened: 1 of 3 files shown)


tarrenj:
I would also like to move `MaxRemotes` (defined in hostmap.go, used in remote_list.go`) to a config element but I was unsure the best way to do this. Any suggestions?

tarrenj:
```Go
// Must type cast instead of using c.GetDuration because the value is multiplied by time.Second everywhere else.
RoamingSuppressSeconds = time.Duration(c.GetInt("dynamic_roaming.roaming_suppress_seconds", 2))
```
This is messy, but was less impactful than removing the `*time.Seconds` everywhere and renaming the removing the `Seconds` from the var name...

johnmaguire:
@tarrenj Could you share the motivation behind this change? Is there a reason you need to configure these values?

PR #596, opened 2021-11-24
Add clean/uninstall and expand install Makefile targets

tarrenj:
Adds a Makefile target for `clean` and `uninstall`, and expands the existing `install` target. Also implement some standard Makefile conventions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,10 @@ ALL = $(ALL_LINUX) \
 	freebsd-amd64 \
 	windows-amd64
 
+CONFIG_DIR = ~/.config/nebula/
+CONFIG_FILE_PATH = "$(CONFIG_DIR)config.yml"
+PKI_DIR = ~/.config/nebula/certificates/
+
 e2e:
 	$(TEST_ENV) go test -tags=e2e_testing -count=1 $(TEST_FLAGS) ./e2e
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -91,6 +95,15 @@ bin:
 install:
 	go install $(BUILD_ARGS) -ldflags "$(LDFLAGS)" ${NEBULA_CMD_PATH}
 	go install $(BUILD_ARGS) -ldflags "$(LDFLAGS)" ./cmd/nebula-cert
+	mkdir -p "$(CONFIG_DIR)";\
+	mkdir -p "$(PKI_DIR)";\
+	cp --update --backup --verbose examples/config "$(CONFIG_FILE)"
```
hlovdal, line 100: [cp on FreeBSD](https://www.freebsd.org/cgi/man.cgi?cp) does not support those command line options.

tarrenj: @hlovdal Thanks for the feedback! I've opened https://github.com/slackhq/nebula/pull/607 as another iteration of this, but I still haven't resolved this issue. Do you know what the convention on handling config files is? The config file should only be copied in if the user has not modified the existing one, I know I could use an `if` to achieve this but I'd like to keep things simple. Maybe the solution is just to always copy the config to `example_config.yml` and make it the users responsibility to make updates in another file?

hlovdal: In rpm based systems, behaviour for overwriting config files is determined both by the specification used to generate the rpm file as well as if the file is modified or not.

A simple hello-world package would typically have something like the following in its spec file:

```rpmspec
%files
%defattr(-,root,root,-)
/usr/bin/hello
%config /etc/hello-message.txt
```

The key thing here is `%config` macro that marks a file specifically as a configuration file. The wast majority of files under /etc are marked as config files, but I did some querying and the first example of something that is not was `/etc/dconf/profile/ibus`. So that means that when ever the `ibus` package is updated, that file will be overwritten regardless of modifications ([Etckeeper](https://etckeeper.branchable.com/) for the win in those situations). But this is quite rare, I had to search quite a bit to find an example.

Since rpm keeps a database of all installed files along with their nominal md5sum it can easily detect if a config file is modified or not when doing updates, and if the config file is not modified it will be updated regardless of the `%config` macro. If not either the new file is either added with an extra `.rpmnew` file extension or the existing version is replaced but saved with an extra `.rpmsave` file extension. Jon Warbrick has an [excellent summary](https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html).

----

But for a Makefile target that lacks rpm's infrastructure I think the only sensible approach is to not overwrite the config file but instead copy it to an additional file (with displayed information/warning). The risk and effect of losing a user's configuration is more severe than the inconvenience of having to do some post install updates.

And specifically in nebula's case, *"if the user has not modified the existing one"* will virtually never be true since > 99.999% of nebula users will modify at least `static_host_map` and `lighthouse.hosts` thus making that point moot.

PR #607, opened 2021-12-13
Add clean/uninstall and expand install Makefile targets

tarrenj:
Add a `clean`, and `uninstall` target.
Expand existing `install` target
Use (some) existing Makefile conventions like specifying the shell and `prefix` directory specification.


tarrenj:
This is a continuation of https://github.com/slackhq/nebula/pull/596

Issue #608, opened 2021-12-13
Report on certificate lifetime in prometheus metrics


benyanke:
It would be very useful to get the certificate expiry date/time in the prometheus metrics, so that administrators could write an prometheus alert to query which hosts need their certs renewed soon.

Is this functionality planned, or would it be accepted as a PR?

nbrownus:
That makes a lot of sense and I don't think I had considered it before just now. Pretty easy to do and would certainly accept a PR for it.

benyanke:
Thanks! I started the work here.

https://github.com/benyanke/nebula/tree/nebula-issue-608

benyanke:
@nbrownus - I put up a PR (#613) which isn't quite ready for merge, but is 80% of the way there and seems to basically working.

How does it look at a high level? Is this the right general approach within the nebula code base, or should it be architected in a different way?

johnmaguire:
Closed by #782.

PR #613, opened 2021-12-14
WIP - Add Prometheus Metrics

benyanke:
This is an open work in progress, looking for comment, not yet ready for review.

This implements the metrics discussed in #608 , improving the metrics provided in the included prometheus exporter.

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -126,7 +126,6 @@ func startPrometheusStats(l *logrus.Logger, i time.Duration, c *config.C, buildV
 	pr.MustRegister(start_time)
 	start_time.Set(float64(time.Now().Unix()))
 
-	// TODO : flesh this out with real data
 	// Export our certificate information as labels on a static gauge
 	for _, ca := range caCert.CAs {
 
--- a/stats.go
+++ b/stats.go
@@ -112,6 +112,49 @@ func startPrometheusStats(l *logrus.Logger, i time.Duration, c *config.C, buildV
 	pr.MustRegister(g)
 	g.Set(1)
 
+	// Export the start time of the current binary execution
+	start_time := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   namespace,
+		Subsystem:   subsystem,
+		Name:        "start_time",
+		Help:        "Time the nebula client started, in unixtime",
+		ConstLabels: prometheus.Labels{},
+	})
+	pr.MustRegister(start_time)
+	start_time.Set(float64(time.Now().Unix()))
+
+	// TODO : flesh this out with real data
+	// Export our certificate information as labels on a static gauge
+	ca_info := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "ca_cert_expiry",
+		Help:      "Returns CA cert time expiry in unixtime",
+		ConstLabels: prometheus.Labels{
+			// TODO : fill these in with actual data, not hardcoded fields
+			"name":       "ca-name-here",
+			"public_key": "1234567890abcdef1234567890abcdef1234567890abcdef",
+		},
+	})
+	pr.MustRegister(ca_info)
+	ca_info.Set(1234534)
+
+	// TODO : flesh this out with real data
+	// Export our certificate information as labels on a static gauge
+	cert_info := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: namespace,
+		Subsystem: subsystem,
+		Name:      "client_cert_expiry",
+		Help:      "Returns client cert time expiry in unixtime",
+		ConstLabels: prometheus.Labels{
+			// TODO : fill these in with actual data, not hardcoded fields
+			"name":       "host-name-here",
+			"public_key": "1234567890abcdef1234567890abcdef1234567890abcdef",
+		},
+	})
+	pr.MustRegister(cert_info)
+	cert_info.Set(987654321)
+
 	var startFn func()
 	if !configTest {
 		startFn = func() {
--- a/stats.go
+++ b/stats.go
@@ -1,6 +1,7 @@
 package nebula
 
 import (
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"log"
@@ -15,13 +16,15 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/rcrowley/go-metrics"
 	"github.com/sirupsen/logrus"
+	"github.com/slackhq/nebula/cert"
 	"github.com/slackhq/nebula/config"
 )
 
 // startStats initializes stats from config. On success, if any futher work
 // is needed to serve stats, it returns a func to handle that work. If no
 // work is needed, it'll return nil. On failure, it returns nil, error.
-func startStats(l *logrus.Logger, c *config.C, buildVersion string, configTest bool) (func(), error) {
+func startStats(l *logrus.Logger, c *config.C, buildVersion string, clientCert *CertState, caCert *cert.NebulaCAPool, configTest bool) (func(), error) {
+
 	mType := c.GetString("stats.type", "")
 	if mType == "" || mType == "none" {
 		return nil, nil
@@ -41,7 +44,7 @@ func startStats(l *logrus.Logger, c *config.C, buildVersion string, configTest b
 		}
 	case "prometheus":
 		var err error
-		startFn, err = startPrometheusStats(l, interval, c, buildVersion, configTest)
+		startFn, err = startPrometheusStats(l, interval, c, buildVersion, clientCert, caCert, configTest)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


stats.go
```diff
@@ -112,6 +115,51 @@ func startPrometheusStats(l *logrus.Logger, i time.Duration, c *config.C, buildV
...
+				// TODO : fill these in with actual data, not hardcoded fields
+				"name":       ca.Details.Name,
+				"public_key": hex.EncodeToString(ca.Details.PublicKey),
+			},
+		})
+
+		pr.MustRegister(ca_info)
+		ca_info.Set(float64(ca.Details.NotAfter.Unix()))
+
+	}
```
benyanke, line 147: I'd love comment on the `CAs` slice - what sort of situations would involve multiple CAs? I'll admit I don't understand fully in what situations you might have multiples here, so this section lines 130-147 might be poorly implemented.

benyanke:
Hi @[nbrownus](https://github.com/nbrownus) - based on our discussion in the issue we had a few months ago, what do you think about this PR?

stats.go
nbrownus: A few main reasons:
- Disaster recovery, trusting multiple roots and keeping all but the currently active one offline allows you to more easily rotate to a new CA in the event the active CA is compromised.
- Ability to rotate out an expiring CA by trusting the current one and the new one being transitioned to.
- CAs can be limited in scope (ips, groups, and subnets they can issue for). You can use these properties to build parts of your mesh with different levels of trust, while still allowing some or all machines to communicate in some way.
- Use individual CAs to build islands and only allow those islands to communicate with specific machines.

benyanke: Thanks, this is very helpful context. I'll do some more testing and potentially refactoring to account for this new info.

stats.go
```diff
@@ -112,6 +115,51 @@ func startPrometheusStats(l *logrus.Logger, i time.Duration, c *config.C, buildV
...
+	// Export the start time of the current binary execution
+	start_time := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace:   namespace,
+		Subsystem:   subsystem,
+		Name:        "start_time",
+		Help:        "Time the nebula client started, in unixtime",
+		ConstLabels: prometheus.Labels{},
+	})
+	pr.MustRegister(start_time)
+	start_time.Set(float64(time.Now().Unix()))
```
nbrownus, line 127: I think this would be better as `uptime` making it effectively a heartbeat that way we can determine when a node stops reporting it.

stats.go
```diff
@@ -15,13 +16,15 @@ import (
...
 	"github.com/sirupsen/logrus"
+	"github.com/slackhq/nebula/cert"
 	"github.com/slackhq/nebula/config"
 )
 
 // startStats initializes stats from config. On success, if any futher work
 // is needed to serve stats, it returns a func to handle that work. If no
 // work is needed, it'll return nil. On failure, it returns nil, error.
-func startStats(l *logrus.Logger, c *config.C, buildVersion string, configTest bool) (func(), error) {
+func startStats(l *logrus.Logger, c *config.C, buildVersion string, clientCert *CertState, caCert *cert.NebulaCAPool, configTest bool) (func(), error) {
```
nbrownus, line 26: We will lose the references to `clientCert` and `caCert` on reload as it stands. We should refactor to use `f.emitStats` instead https://github.com/slackhq/nebula/blob/b5b9d33ee753b510b316f15b503ecb11209fdf9b/interface.go#L309

Using the `Interface`s reference to `f.certState` and `f.caPool`.

We also support both prometheus and graphite. I think many of these stats would work well in a graphite world where labels aren't readily supported, so we would need to either make these work as best as possible with `gometrics` or teach `Interface` to know which stats system is being used and export the appropriate information for graphite or prometheus.

stats.go
```diff
@@ -112,6 +115,51 @@ func startPrometheusStats(l *logrus.Logger, i time.Duration, c *config.C, buildV
...
+
+		ca_info := prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: namespace,
+			Subsystem: subsystem,
+			Name:      "ca_cert_expiry",
+			Help:      "Returns CA cert time expiry in unixtime",
+			ConstLabels: prometheus.Labels{
+				// TODO : fill these in with actual data, not hardcoded fields
+				"name":       ca.Details.Name,
+				"public_key": hex.EncodeToString(ca.Details.PublicKey),
```
nbrownus, line 140: We should use `clientCert.Sha256Sum()` instead since this is how the `pki.blocklist` works what is printed in many of the cert related logs.

stats.go
```diff
@@ -112,6 +115,51 @@ func startPrometheusStats(l *logrus.Logger, i time.Duration, c *config.C, buildV
...
+			},
+		})
+
+		pr.MustRegister(ca_info)
+		ca_info.Set(float64(ca.Details.NotAfter.Unix()))
```
nbrownus, line 145: I think it would be easier to write alerts if this tracked the remaining seconds until expiry: `remaining time < 14 days`.

It could also be interesting to report how long the cert was issued for. An example would be to warn at the half life: `remaining time / valid lifetime <= 0.5`

stats.go
```diff
@@ -112,6 +115,51 @@ func startPrometheusStats(l *logrus.Logger, i time.Duration, c *config.C, buildV
...
+
+	}
+
+	// Export our certificate information as labels on a static gauge
+	cert_info := prometheus.NewGauge(prometheus.GaugeOpts{
```
nbrownus, line 150: I would give this the same treatment as the notes in `ca_info`

stats.go
```diff
@@ -112,6 +115,51 @@ func startPrometheusStats(l *logrus.Logger, i time.Duration, c *config.C, buildV
 	pr.MustRegister(g)
 	g.Set(1)
 
+	// Export the start time of the current binary execution
+	start_time := prometheus.NewGauge(prometheus.GaugeOpts{
```
nbrownus, line 119: `startTime` to follow camel case convention, a few others like this below.

stats.go
benyanke: I considered that, but this sort of thing seems to be a bit of a prometheus anti-pattern. The generally accepted practice around prometheus metrics is when something is static, it should be reported as static and write the alert around that. I can update it to your method if you insist, but it would be breaking the typical way people use prometheus.

Thoughts?

benyanke: Examples of this are `node_boot_time_seconds` in the prometheus node exporter, and `ssl_earliest_cert_expiry` in the blackbox exporter, both of which are first-party prometheus exporters maintained by the core maintainers of prometheus.

stats.go
nbrownus: I am thinking of process uptime - number of seconds elapsed since the `nebula` process started.

Backing up, what is your use case for this stat?

benyanke: It's my understanding that's what I have here - as that function is only resolved initially, not on every metrics request.

My primary usecase for this stat is to allow someone to write an alert to detect how often the nebula agent is being restarted. It could also be used for other things like alerting after a config change is pushed by config management, to allow admins to see which agents haven't yet restarted. It's a common prometheus metrics pattern to provide this data. for example, Traefik's `traefik_config_last_reload_success` metric which does a similar thing.

stats.go
nbrownus: There are a few complicating factors. Certificates can be issued with different lifespans. You may want some nodes with certs that live for hours while others live for months or years. We would at least need to add `NotBefore` so that we can determine the lifespan.

Exporting only the static information is likely the most widely applicable but also the most frustrating to implement alerts for.

Maybe @wadey has some thoughts

benyanke: Adding `NotBefore` makes sense to me. I can do that.

However, on the alerting point, I'd argue it would be far more user-hostile to go against prometheus typical usage patterns. Anyone already using prometheus for other things (like the Blackbox exporter to monitor) web certificates will already be deeply familiar with this pattern. The key thing is that anyone using prometheus is unlikely to be using it *only* to monitor nebula, it's a broader ecosystem.

Again, I'd be happy to adopt your suggestion if you insist, but I think it will be user-hostile for prometheus users.

stats.go
nbrownus: Wouldn't using `changes` on `up` achieve a similar result?

benyanke: Not sure I understand what you mean.

stats.go
nbrownus: No I think we can get by with `NotBefore` and `NotAfter`. 

Alerting based on a percentage of consumed lifespan could be less burdensome than `(nebula_client_client_cert_not_after - time()) / (nebula_client_client_cert_not_after - nebula_client_client_cert_not_before) <= 0.5`, so maybe we want an to expose both styles (possibly optionally).

A sampling of the exporters that live out there show a mix of approaches
- Duration values https://github.com/joe-elliott/cert-exporter (grafana advises the use of this one https://grafana.com/blog/2020/03/04/pro-tip-how-to-monitor-client-certificate-expirations-with-prometheus/)
- Static values https://github.com/ribbybibby/ssl_exporter
- Duration values https://github.com/amimof/node-cert-exporter
- Static values https://github.com/enix/x509-certificate-exporter

benyanke: For what it's worth, the alert query string you noted above seems pretty reasonable to me, but I've also been using prometheus for a while. I have many alert queries even more complex than that, and most other non-trivial prometheus users would too.

I don't deny that some smaller exporters (all the ones linked are under 300 stars) might do different things, but it does still seem worth trying to follow the standard created and recommended by the prometheus maintainers where possible, unless there's a really strong reason not to do so.

Is there a use-case I'm missing where a substantial amount of people will be using prometheus to alert only on nebula and nothing else?

benyanke: Alternatively, what if I update the PR to provide a documentation section on suggested alerts? This is something I would like to do anyways, I could roll it in with this one, and get rid of the need of anyone to figure out the alert string. I'd also like to get the alert strings put here too (https://awesome-prometheus-alerts.grep.to) if they accept it.

JohnMaguire:
Closing this out as #782 merged.

PR #615, opened 2021-12-19
Add dns wildcarding

aaronstillwell:
This PR is an attempt at implementing DNS wildcard support. The use case in mind is for those of us who wish to use nebula's DNS in conjunction with virtual hosts.

At the moment, say we have a server `my-server.nebula`, it is not possible to use virtual hosts to have, for example multiple apps served at `app-1.my-server.nebula`, `app-2.my-server.nebula` etc.

This initial (naive) implementation can be enabled with the new config options as seen below.

```
# ...
lighthouse:
  # ...
  dns:
    # ...
    wildcard: true
    wildcard_limit: 5 # recursion limit used when searching for a matching host
```

This config is global and doesn't support per-host or per-group enabling. I welcome discussion on the proposed feature and its implementation as-is as well as commentary around the desired configuration of the feature.

Relevant slack discussion: https://nebulaoss.slack.com/archives/CRWJJM52B/p1639478617050700

### DIFF
--- a/dns_server.go
+++ b/dns_server.go
@@ -38,22 +38,19 @@ func newDnsRecords(hostMap *HostMap) *dnsRecords {
 
 func (d *dnsRecords) Query(data string) string {
 	d.RLock()
-	if r, ok := d.dnsMap[data]; ok {
-		d.RUnlock()
-		return r
-	}
 
 	if d.dnsWildcardEnabled {
-		limitCounter := 0
-		parts := strings.Split(data, ".")
-		for len(parts) > 1 && limitCounter < d.dnsWildcardLimit {
-			host := strings.Join(parts[1:], ".")
-			if r, ok := d.dnsMap[host]; ok {
+		hostParts := strings.SplitN(data, ".", d.dnsWildcardLimit+1)
+		for i := range hostParts {
+			if r, ok := d.dnsMap[strings.Join(hostParts[i:], ".")]; ok {
+				d.RUnlock()
 				return r
 			}
-			parts = strings.Split(host, ".")
-
-			limitCounter++
+		}
+	} else {
+		if r, ok := d.dnsMap[data]; ok {
+			d.RUnlock()
+			return r
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


dns_server.go
```diff
@@ -37,6 +42,21 @@ func (d *dnsRecords) Query(data string) string {
 		d.RUnlock()
 		return r
 	}
+
+	if d.dnsWildcardEnabled {
+		limitCounter := 0
+		parts := strings.Split(data, ".")
```
wadey, line 48: I think you can just use strings.SplitN here instead of the limit counter.

https://pkg.go.dev/strings#SplitN

aaronstillwell: Will refactor this and also fix the missing call to `d.RUnlock()`

wadey:
Can you update examples/config.yml with the new config options? (They can be commented out by default)

dns_server.go
```diff
@@ -113,6 +130,8 @@ func handleDnsRequest(l *logrus.Logger, w dns.ResponseWriter, r *dns.Msg) {
 
 func dnsMain(l *logrus.Logger, hostMap *HostMap, c *config.C) func() {
 	dnsR = newDnsRecords(hostMap)
+	dnsR.dnsWildcardEnabled = c.GetBool("dns.wildcard", false)
+	dnsR.dnsWildcardLimit = c.GetInt("dns.wildcard_limit", 5)
```
aaronstillwell, line 134: @wadey I've just realised these should be `lighthouse.dns.wildcard` right?

wadey: Yes that is correct https://github.com/slackhq/nebula/blob/v1.5.2/examples/config.yml#L32

tarrenj:
Very excited for this feature!

examples/config.yml
```diff
@@ -33,6 +33,8 @@ lighthouse:
     # The DNS host defines the IP to bind the dns listener to. This also allows binding to the nebula node IP.
     #host: 0.0.0.0
     #port: 53
+    #wildcard: true # Enable wildcarding on hosts e.g allowing my-app.my-host.nebula to resolve to the IP for my-host.nebula
+    #wildcard_limit: 5 # Limit of number of possible subdomains
```
jstasiak, line 37: I'd rephrase this to something like `# The maximum supported subdomain depth` or something. "number of possible" subdomains sounds to me like it would limit the number of any level subdomains

jstasiak (approved):
Not a Go person and not a member of the team but I had a quick look and it looks good to me. I would very much like to see this land as it's a feature I'd like to (be able to) use.

nbrownus:
Closing as stale

PR #616, opened 2021-12-21
Add some minor documentation to the example config

tarrenj:


nbrownus:
Thanks for the PR, the note about port 0 is (unfortunately) 2 lines above this.

PR #617, opened 2021-12-27
hostname adding to DNS handles multiple hostnames

stewart123579:
Multiple hostnames can be specified in nebula-cert by separating them by
a space and then each string will now be added to the DNS server with
the same details.


LarsHaalck:
This feature would be really nice for hosts running something like a (nginx) reverse proxy and would allow to expose different (or all if also merging the wildcard PR) virtual hosts. 

I would really appreciate a merge.

nbrownus:

PR #618, opened 2021-12-27
Extend dns to allow wildcards

stewart123579:
This extends PR https://github.com/slackhq/nebula/pull/617 (which allows for multiple hostnames for a single nebula node) by allowing wildcards in the hostname.


nbrownus:

PR #627, opened 2022-01-24
Fix errant capitalisation in DNS TXT response

rktjmp:
The DNS TXT record response mislabels the cert expiry as `NotAFter`, instead of `NotAfter`,  I don't *think* this is intentional.

This patch could technically be a breaking change to peoples infrastructure.


johnmaguire:
Thanks for the PR - rebased in #1127.

PR #649, opened 2022-03-08
Lighthouse reload support

nbrownus:
Adds support for reloading many of the `lighthouse` and `punchy` config stanzas

Replaces #149

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -26,6 +27,7 @@ type C struct {
 	oldSettings map[interface{}]interface{}
 	callbacks   []func(*C)
 	l           *logrus.Logger
+	reloadLock  sync.Mutex
 }
 
 func NewC(l *logrus.Logger) *C {
@@ -133,6 +135,9 @@ func (c *C) CatchHUP(ctx context.Context) {
 }
 
 func (c *C) ReloadConfig() {
+	c.reloadLock.Lock()
+	defer c.reloadLock.Unlock()
+
 	c.oldSettings = make(map[interface{}]interface{})
 	for k, v := range c.Settings {
 		c.oldSettings[k] = v
@@ -150,6 +155,9 @@ func (c *C) ReloadConfig() {
 }
 
 func (c *C) ReloadConfigString(raw string) error {
+	c.reloadLock.Lock()
+	defer c.reloadLock.Unlock()
+
 	c.oldSettings = make(map[interface{}]interface{})
 	for k, v := range c.Settings {
 		c.oldSettings[k] = v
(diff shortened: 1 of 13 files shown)


lighthouse.go
```diff
@@ -39,80 +45,225 @@ type LightHouse struct {
...
+	return (*RemoteAllowList)(atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&lh.atomicRemoteAllowList))))
+}
+
+func (lh *LightHouse) GetLocalAllowList() *LocalAllowList {
+	return (*LocalAllowList)(atomic.LoadPointer((*unsafe.Pointer)(unsafe.Pointer(&lh.atomicLocalAllowList))))
+}
+
+func (lh *LightHouse) reload(c *config.C, initial bool) error {
+	if initial || c.HasChanged("lighthouse.interval") {
+		atomic.StoreInt64(&lh.interval, int64(c.GetInt("lighthouse.interval", 10)))
```
wadey, line 135: We use `atomic.StoreInt64` to write the value for interval, but we aren't using `LoadInt64` for the place that reads interval (LhUpdateWorker). We should probably make this consistent. Should we also rename to `atomicInterval`?

lighthouse.go
```diff
@@ -39,80 +45,225 @@ type LightHouse struct {
...
+	if initial || c.HasChanged("lighthouse.remote_allow_list") || c.HasChanged("lighthouse.remote_allow_ranges") {
+		ral, err := NewRemoteAllowListFromConfig(c, "lighthouse.remote_allow_list", "lighthouse.remote_allow_ranges")
+		if err != nil {
+			return util.NewContextualError("Invalid lighthouse.remote_allow_list", nil, err)
+		}
+
+		atomic.StorePointer((*unsafe.Pointer)(unsafe.Pointer(&lh.atomicRemoteAllowList)), unsafe.Pointer(ral))
+		if !initial {
+			//TODO: a diff will be annoyingly difficult
+			lh.l.Info("lighthouse.remote_allow_list and/or lighthouse.remote_allow_ranges has changed")
```
wadey, line 151: If we want to do it, [github.com/google/go-cmp](https://github.com/google/go-cmp) works well for this

wadey, line 151: If we want to do it, [github.com/google/go-cmp](https://github.com/google/go-cmp) works well for this (assuming we do it against config.oldSettings and not the CIDRTree itself).

config/config.go
```diff
@@ -149,6 +149,24 @@ func (c *C) ReloadConfig() {
 	}
 }
 
+func (c *C) ReloadConfigString(raw string) error {
```
wadey, line 152: Should we have a lock in this method to make sure something doesn't double-trigger the reload, and a second one starts while the first is still running? Could that happen?

wadey, line 152: Should we have a lock in this method to make sure something doesn't double-trigger the reload, and a second one starts while the first is still running? Could that happen?

wadey (approved):
A few nits and questions both otherwise looks good :+1:

wadey (approved):
LGTM

PR #653, opened 2022-03-15
Make nebula advertise its dynamic port to lighthouses

brad-defined:
Nebula is advertising port '0' to the lighthouse, when configured to dynamically choose a port.

This PR moves the dynamic port discovery code into NewLighthouseFromConfig, to fix the issue.


brad-defined:
Windows e2e deadlock, I don't believe it's related to the PR.

nbrownus (approved):

nbrownus:
Fixes a regression from #649

PR #655, opened 2022-03-17
Update link to nebula docs

jasikpark:
`defined.net/nebula/introduction` doesn't exist anymore; the README should point towards `defined.net/nebula/` instead


wadey (approved):
Thank you

PR #656, opened 2022-03-17
build and test with go1.18

wadey:
https://go.dev/doc/go1.18


JohnMaguire:
> A go directive indicates that a module was written assuming the semantics of a given version of Go. The version must be a valid Go release version: a positive integer followed by a dot and a non-negative integer (for example, 1.9, 1.14).

https://go.dev/ref/mod#go-mod-file-go

Given the new support for generics as well as changes to /x/sys, if developers & CI are running with go1.18, perhaps the go.mod should be bumped as well?

wadey:
> Given the new support for generics as well as changes to /x/sys, if developers & CI are running with go1.18, perhaps the go.mod should be bumped as well?

I don't see a good reason not to, so let's go ahead and do that. Updated the PR.

nbrownus (approved):

PR #659, opened 2022-03-29
Typo in README.md: ca.cert -> ca.crt

Osndok:
I noticed that one of the generated file names is not what the README describes. It is off by one character.


tarrenj:
In a similar vain to `cert`->`crt`, I'd really like to see command line arguments use `--` for long options versions (`--config`) and `-` for short options (`-c`), as described in the [GNU getopt](https://www.gnu.org/software/libc/manual/html_node/Getopt-Long-Options.html) and [IEEE](https://www.gnu.org/software/libc/manual/html_node/Getopt-Long-Options.html) specs.

PR #670, opened 2022-04-19
add listen.send_recv_error config option

wadey:
By default, Nebula replies to packets it has no tunnel for with a `recv_error` packet. This packet helps speed up reconnection
in the case that Nebula on either side did not shut down cleanly. This response can be abused as a way to discover if Nebula is running
on a host though. This option lets you configure if you want to send "recv_error" packets always, never, or only to private network remotes.
valid values: always, never, private

This setting is reloadable with SIGHUP.


interface.go
```diff
@@ -309,6 +348,29 @@ func (f *Interface) reloadFirewall(c *config.C) {
 		Info("New firewall has been installed")
 }
 
+func (f *Interface) reloadSendRecvError(c *config.C) {
+	stringValue := c.GetString("listen.send_recv_error", "")
+	if (c.InitialLoad() && stringValue != "") || c.HasChanged("listen.send_recv_error") {
+		switch stringValue {
+		case "always":
+			f.sendRecvErrorConfig = sendRecvErrorAlways
```
brad-defined, line 356: If I'm reading this correctly, the switch statement will only execute in the case of InitialLoad if the value is set to something. If the value is not set...go will implicitly set f.sendRecvErrorConfig to an initial value of 0, which happens to match sendRecvErrorAlways.

I think it'd be nice to explicitly set the default value, though. Maybe edit line 352 to say `stringValue := c.GetString("listen.send_recv_error", "always")` ?

interface.go
```diff
@@ -309,6 +348,29 @@ func (f *Interface) reloadFirewall(c *config.C) {
...
+		case "never":
+			f.sendRecvErrorConfig = sendRecvErrorNever
+		case "private":
+			f.sendRecvErrorConfig = sendRecvErrorPrivate
+		default:
+			if c.GetBool("listen.send_recv_error", true) {
+				f.sendRecvErrorConfig = sendRecvErrorAlways
+			} else {
```
brad-defined, line 364: The default, if unset, is sendRecvErrorAlways. But if the user set a value that doesn't match, it's sendRecvErrorNever. Is this intended?

I was thinking if an unknown value is set, it should be logged and the default value be applied.

interface.go
wadey: good point, I think I have cleaned this up with the latest commit if you can review again.

nbrownus (approved):

PR #675, opened 2022-05-02
fix panic in handleInvalidCertificate

donstephan:
When HandleMonitorTick fires, the hostmap can be nil which causes a panic to occur when trying to clean up the hostmap in handleInvalidCertificate. This fix just stops the invalidation from continuing if the hostmap doesn't exist.

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -286,6 +286,10 @@ func (n *connectionManager) handleInvalidCertificate(now time.Time, vpnIp iputil
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


brad-defined:
https://github.com/slackhq/nebula/pull/658 fixed this in HandleMonitorTick, I bet the same fix in HandleDeletionTick would work best.

(Meaning update HandleDeletionTick to remove the disconnectInvalid conditional...which I don't think belongs there.)

Issue #85, opened 2019-12-11
Nodes can see the Lighthouse but they cant see eachother


gerroon:
Hi

I setup a small network of 3+ nodes. Non LH nodes can ping the LH. LH can ping the nodes but the nodes cant ping eachoher.

This seems to work only for the nodes that are on the same wifi network. Anything from external to another external node or external to internal does not work, unless there is another form of VPN is active between the exteral nodes, like Wireguard.

The LH is behind a router so I port forwarded the default port, this seems to work given that any of the nodes can connect to the LH.

It is interesting that when I try to ping from one of the external nodes to a node in the home wifi, there is activity on the receiving internal node, but pings are all unsuccessful meaning that the ping just stalls.

```

time="2019-12-11T14:46:38-06:00" level=info msg="Handshake message sent" handshake="map[stage:1 style:ix_psk0]" initiatorIndex=688165208 remoteIndex=0 udpAddr="192.168.0.23:59683" vpnIp=10.x.0.12

time="2019-12-11T14:46:40-06:00" level=info msg="Handshake message sent" handshake="map[stage:1 style:ix_psk0]" initiatorIndex=688165208 remoteIndex=0 udpAddr="EXTERNAL-IP:59683" vpnIp=10.x.0.12

time="2019-12-11T14:46:43-06:00" level=info msg="Handshake message sent" handshake="map[stage:1 style:ix_psk0]" initiatorIndex=688165208 remoteIndex=0 udpAddr="10.3.0.2:59683" vpnIp=10.x.0.12


```

I have all the punch stuff enabled. Am I supposed to forward more ports or port ranges?

Please bear in mind that in the given situation WG works perfectly, and all the wg nodes can see eachother without issues, including all the traffic routing setup. I would like to setup Nebula as a fallback solution, in case one wonders why I am trying to use both.

wych42:
Same issue here.

I've got one lighthouse node with public ip addr, firewall open port 4242 for udp&tcp;
one laptop at home behind nat; one laptop at office(also behind nat).The connection status would be:

lighthouse <-both way connected-> home laptop
lighthouse <-both way connected-> office laptop
home laptop <-no connection-> office laptop

agreyfox:
I have same issues during setup stage and after  I change the config.yml in two settings, It works now.
1 "punch_back" should be true.
2 in the peer host section, lighthouse.host should be lighthouse address. and for lighthouse , I leave it blank. 

hope helps.

wych42:
[...]

PR #678, opened 2022-05-13
Relay

brad-defined:
All feedback is encouraged.

This PR enables a nebula node to relay traffic to a peer through an intermediary Nebula node.

Relay traffic is end-to-end encrypted, and hop-by-hop AD verified.

```mermaid
graph LR
    A[Peer HI <-> Terminal HI] -->|AD| B(Fwd HI <-> Fwd HI)
    B -->|AD| C(Terminal HI <-> Peer HI)
    A -->|End-to-End Encrypted| C
```

Relays are a feature of HostInfo objects. There are two different types of Relay objects - Forwarding relays, and Terminal relays. When a Terminal Relay receives a packet from an associated forwarding relay object, it authenticates the packet from the previous hop and then passes the packet through the regular packet processing routine, treating it the same as if it were received from the UDP socket.

A Forwarding Relay exists on the Relay node itself. It either processes a packet passed from another HostInfo, or processes a packet received from the UDP socket. It forwards the packet to a next hop, which will either be another local HostInfo that's either a Forwarding or Terminal Relay, or write the packet to the wire for a peer on a different host.

This adds a new message subtype: `MessageRelay`, on the `Message` type. On receipt of this message type, Nebula will look up the Relay information associated with the index indicated in the Nebula header. This header indexes to the HostInfo object of the previous hop. Relay messages are only forwarded over established Nebula connections, and are authenticated by validating the payload as AD info, avoiding double-encryption. When forwarding the Relay traffic, the packet payload is authenticated based on the HostInfo for the _next_ hop.

```mermaid
sequenceDiagram
    note left of Connector: Encrypt message for Peer, wrap in Neubula header
    Note left of Connector: AD payload for next hop `Relay`
    Connector->>+Relay: MessageRelay to RelayIdx 4
    Note right of Relay: Look up HostInfo of Connector by RelayIdx, verify AD
    Note right of Relay: Look up HostInfo of Relay PeerIP, AD payload for next hop `Peer`
    Relay->>+Peer: MessageRelay to RelayIdx 3
    note right of Peer: Look up HostInfo of Relay by RelayIdx, verify AD
    note right of Peer: Relay is Terminal, so process directly. Interpret payload as Nebula packet.
    note right of Peer: Look up HostInfo of `Connector` by Index, decrypt packet
```

So, if a Nebula node `Connector` wishes to relay to a peer `Peer` through an intermediary `Relay`, it will
1. Establish a Nebula connection to `Relay` with a handshake
2. Request a Relay tunnel to `Peer`
3. Create a handshake packet for `Peer`
4. Forward this handshake packet through the Relay tunnel to `Relay`, as AD traffic for `Relay`
5. Wait for its response handshake
6. Calculate symmetric keys
7. Forward all remaining traffic to `Peer` encrypted with its symmetric keys for `Peer` as AD for `Relay`

The Relay object holds local index, remote index, peer Ip, state, and type info. The type will either be `TerminalType`, meaning the host should process the payload directly, or `ForwardingType`, which means this host should forward the payload to the Relay.PeerIP.

For actual relaying, during handshake, the Nebula node will attempt to create Nebula tunnels to all discovered relays for its target peer. Once those tunnels are established, it'll send a new CreateRelayRequest message to the Relay, indicating its target VpnIP.
```mermaid
sequenceDiagram
    Connector->>+Relay: CreateRelayReuest, idx=1, from C to P
    Relay->>Peer: CreateRelayRequest, idx=2, from C to P
    note right of Peer: Relay{State:Established, PeerIP: C}
    Peer->>Relay: RelayCreated, idx=3
    Note right of Relay: Relay{localIdx 2, remoteIdx 3, State: Established, PeerIP: C }
    Note right of Relay: Relay{localIdx 4, remoteIdx 1, State: Established, PeerIP: P }
    Relay->>Connector: RelayCreated, idx=4
    Note left of Connector: Relay{localIdx 1, remoteIdx 4, State: Established, PeerIp: P }
```

The example config is updated with Relay config options.
A Nebula node is configured to act as a relay by settings its `relay.am_relay` to `true`
```
relay:
  am_relay: true
```

A Nebula node configures which hosts can be used as relays to access it by listing them in its `relay.relays` config.
```
relay:
  relays:
    -  192.168.100.1
```

Hosts so named must have their `am_relay` set to `true`, otherwise they'll ignore CreateRelayRequest messages. These IPs will be advertised to the LightHouse, and included in host query responses.

Cleanup depends on the existing connection manager logic. When the relays fail, the connection manager should detect the failed connection. When HostInfo objects are cleaned up due to these failures, the associated Relay state is also (locally) cleaned up.


handshake_manager.go
```diff
@@ -145,6 +145,8 @@ func (c *HandshakeManager) handleOutbound(vpnIp iputil.VpnIp, f udp.EncWriter, l
 
 	// Get a remotes object if we don't already have one.
 	// This is mainly to protect us as this should never be the case
+	// NB ^ This comment doesn't jive. It's how the thing gets intiailized.
```
brad-defined, line 148: NB = Nota Bene, a note I left for myself when reading the code.
NB != Nate Brown, that's just a fun coincidence.

jasikpark:
Is this meant to solve part of https://github.com/slackhq/nebula/issues/85#issuecomment-586622822 ?

nebula.proto
```diff
@@ -26,6 +25,7 @@ message NebulaMetaDetails {
   uint32 VpnIp = 1;
   repeated Ip4AndPort Ip4AndPorts = 2;
   repeated Ip6AndPort Ip6AndPorts = 4;
+  repeated uint32 RelayVpnIp = 6;
```
wadey, line 28: Why was field index `5` skipped in here? We should either fix this to be `5` or document why it was skipped.

hostmap.go
```diff
@@ -292,6 +418,17 @@ func (hm *HostMap) QueryIndex(index uint32) (*HostInfo, error) {
 		return nil, errors.New("unable to find index")
 	}
 }
+func (hm *HostMap) QueryRelayIndex(index uint32) (*HostInfo, error) {
+	//TODO: we probably just want ot return bool instead of error, or at least a static error
```
vbscott, line 422: Nit: s/ot/to

nebula.proto
brad-defined: thx - 5 was used for CIDR advertisements, which I removed for the new config scheme. Can move RelayVpnIp down a notch :)

outside.go
```diff
@@ -7,21 +7,21 @@ import (
...
+	"github.com/golang/protobuf/proto"
 	"github.com/sirupsen/logrus"
 	"github.com/slackhq/nebula/cert"
 	"github.com/slackhq/nebula/firewall"
 	"github.com/slackhq/nebula/header"
 	"github.com/slackhq/nebula/iputil"
 	"github.com/slackhq/nebula/udp"
 	"golang.org/x/net/ipv4"
-	"google.golang.org/protobuf/proto"
 )
```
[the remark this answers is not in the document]
brad-defined: I had a build error with the new protobuf import. I didn't realize the `google.golang.org` one was deprecated - this might be something wrong with my local protobuf compiler, will look into it.
```
./outside.go:89:31: cannot use &m (value of type *NebulaControl) as type protoreflect.ProtoMessage in argument to proto.Marshal:
        *NebulaControl does not implement protoreflect.ProtoMessage (missing ProtoReflect method)
```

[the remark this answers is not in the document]
brad-defined: I read up on this a little, and this is what I think is going on - 

The cert proto uses the new-and-improved protobuf generation code
https://github.com/slackhq/nebula/blob/master/cert/Makefile#L5
But the messages themselves use old-style protobufs
https://github.com/slackhq/nebula/blob/master/Makefile#L151

I tried to update the Nebula messages to the new style, but ran into a decision point due to the lack of MarshalTo() in the new style protobufs (which is called a couple of times in lighthouse.go) - at which point I stopped trying the migration, as it looks large enough to be a separate task.

I conclude that I can either move the Nebula protocol protobuf interaction out of this file (so that the existing Cert protobuf code can use the new imported module as-was), or import both protobufs and give one an alias, since this file handles both Nebula protocol protobufs as well as cert protobufs.

handshake_manager.go
```diff
@@ -181,6 +183,76 @@ func (c *HandshakeManager) handleOutbound(vpnIp iputil.VpnIp, f udp.EncWriter, l
...
+	hostinfo.logger(c.l).Infof("Look for relays...%v", hostinfo.remotes.relays)
+	// Send a RelayRequest to all known Relay IP's
+	for _, relay := range hostinfo.remotes.relays {
+		// Don't relay to myself, and don't relay through the host I'm trying to connect to
+		if *relay == vpnIp || *relay == c.lightHouse.myVpnIp {
+			continue
+		}
+		relayHostInfo, err := c.mainHostMap.QueryVpnIp(*relay)
+		if err != nil || relayHostInfo.GetRemote() == nil {
+			hostinfo.logger(c.l).WithError(err).WithField("relay", relay.String()).Info("Failed to find relay in main hostmap, or relay is not directly connected. Send test message.")
```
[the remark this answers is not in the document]
brad-defined: might look weird but doesn't break anything.

hostmap.go
```diff
@@ -245,9 +348,37 @@ func (hm *HostMap) DeleteReverseIndex(index uint32) {
 }
 
 func (hm *HostMap) DeleteHostInfo(hostinfo *HostInfo) {
```
brad-defined, line 350: I mostly punted on thinking through locking here. Looking at it now. I'm trying to come up with a locking strategy that avoids the possibility of deadlocks and is correct, but it's a bit challenging.

Suggestions here are welcome!

hostmap.go
```diff
@@ -397,6 +539,18 @@ func (hm *HostMap) Punchy(ctx context.Context, conn *udp.Conn) {
...
+func (i *HostInfo) AddRelay(idx uint32, relayIp iputil.VpnIp) {
+	i.Lock()
+	i.relays[relayIp] = struct{}{}
+	i.Unlock()
+}
+
+func (i *HostInfo) GetRemote() *udp.Addr {
+	i.RLock()
+	defer i.RUnlock()
+	return i.remote
```
brad-defined, line 551: :/ This may cause some problems. I think there are code paths where we hold the hostinfo lock and other paths where we don't, both of which read this value. If we call this function to get the remote, we may try to acquire a lock we've already got.

I think I'll roll this one back, and think deep thoughts in some other PR.

hostmap.go
brad-defined: I'm going to try sticking all the relay variables into a struct that has its own lock, so they can be managed independently from the HostInfo lock, but still be safe for concurrent access.

handshake_manager.go
```diff
@@ -181,6 +185,74 @@ func (c *HandshakeManager) handleOutbound(vpnIp iputil.VpnIp, f udp.EncWriter, l
 			Info("Handshake message sent")
 	}
 
+	if c.config.useRelays {
+		hostinfo.logger(c.l).Infof("Attempt to relay through hosts (%v)", hostinfo.remotes.relays)
```
wadey, line 189:
```suggestion
	if c.config.useRelays && len(hostinfo.remotes.relays) > 0 {
		hostinfo.logger(c.l).Infof("Attempt to relay through hosts (%v)", hostinfo.remotes.relays)
```

Change this log message to only trigger if there are relays for the hostinfo

outside.go
```diff
@@ -183,8 +272,12 @@ func (f *Interface) handleEncrypted(ci *ConnectionState, addr *udp.Addr, h *head
 	// If connectionstate exists and the replay protector allows, process packet
 	// Else, send recv errors for 300 seconds after a restart to allow fast reconnection.
 	if ci == nil || !ci.window.Check(f.l, h.MessageCounter) {
-		f.sendRecvError(addr, h.RemoteIndex)
-		return false
+		if addr != nil {
```
nbrownus, line 275: This is another spot where combining `udp.Addr` with the `ViaSender` could be helpful

relay_manager.go
```diff
@@ -0,0 +1,315 @@
...
+	relay, err := rm.EstablishRelay(h, m)
+	if err != nil {
+		rm.l.WithError(err).WithField("target", target.String()).Error("Failed to update relay for target")
+		return
+	}
+	// Do I need to complete the relays now?
+	if relay.Type == TerminalType {
+		return
+	}
+	// I'm the middle man. Let the initiator know that the I've established the relay they requested.
```
nbrownus, line 138: Nit: "know that I've established"

nbrownus (commented):
Great work on this! My comments are largely meant as notes to fix in the near future in separate pull requests. @wadey how does it look to you?

nebula.proto
```diff
@@ -63,3 +63,16 @@ message NebulaHandshakeDetails {
   uint64 Time = 5;
 }
 
+message NebulaControl {
+  enum MessageType {
+    None = 0;
```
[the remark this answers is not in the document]
brad-defined: Two things: first, it was inspired by the code above:
```
message NebulaMeta {
  enum MessageType {
    None = 0;
    HostQuery = 1;
    HostQueryReply = 2;
```

When in doubt, cut & paste.

Secondly, Go defaults unset int values to 0. Reserving the 0 value as 'none' helps ensure that developers specify the Type value explicitly when constructing Control messages.

Without reserving 0 (and setting CreateRelayRequest to enum value 0), a CreateRelayRequest could look like this:
```
			req := NebulaControl{
				Type:                NebulaControl_CreateRelayRequest,
				InitiatorRelayIndex: index,
				RelayFromIp:         uint32(h.vpnIp),
				RelayToIp:           uint32(target),
			}
```
or this (missing the Type value being set):
```
			// Send a CreateRelayRequest to the peer.
			req := NebulaControl{
				InitiatorRelayIndex: index,
				RelayFromIp:         uint32(h.vpnIp),
				RelayToIp:           uint32(target),
			}
```

Reserving '0' for None makes the second construction more detectable (because the peer will reject it.)

wadey (approved):
At this point I think this is safe to merge in and fix forward, looking forward to it!

nbrownus:
We are merging this PR as mostly complete and fixing issues/cleaning up in future PRs. Known items are tracked here #690

fly512:
How to choose an optimal path

tarrenj:
I know I'm quite late here, but this is an incredibly exciting feature add. Great work devs!!

windwalker78:
Can the lighthouse advertise itself a a relay host too?

joekale:
> Can the lighthouse advertise itself a a relay host too?

I do not believe it will "advertise" but it can be used as a lighthouse. You will need to configure the lighthouse to act as a relay and add it's nebula IP as a relay in the list of relays on node configs. I don't know what the minimal set of configs that need it added as a relay is though.

PR #687, opened 2022-06-01
Update protobuf

brad-defined:
Migrate to the latest google protobuf.

The gogo lib had a method to Marshal to a buffer that could avoid allocation. The new golang protobuf library also has a method for accomplishing this, included in this pr.


brad-defined:
Not much interest in this one I think. Generally happy with the current mix of protobuf's.

PR #688, opened 2022-06-06
Add a basic freebsd init script

gjedeer:


nbrownus:
Thanks for the PR! FreeBSD has a ports for nebula which includes the init script

https://cgit.freebsd.org/ports/tree/security/nebula

PR #693, opened 2022-06-23
Run `make vet` in CI

jasikpark:
currently `make vet` fails on the repo:

```
./firewall_test.go:140:7: github.com/slackhq/nebula/firewall.Packet composite literal uses unkeyed fields
./firewall_test.go:315:7: github.com/slackhq/nebula/firewall.Packet composite literal uses unkeyed fields
./firewall_test.go:374:7: github.com/slackhq/nebula/firewall.Packet composite literal uses unkeyed fields
./firewall_test.go:460:7: github.com/slackhq/nebula/firewall.Packet composite literal uses unkeyed fields
./control.go:75:2: misuse of unbuffered os.Signal channel as argument to signal.Notify
./control.go:76:2: misuse of unbuffered os.Signal channel as argument to signal.Notify
```

This PR adds a `make vet` step to each of the test jobs. I'll add fixups for the simple go vet errors.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,9 @@ jobs:
     - name: Build
       run: make all
 
+    - name: Vet
+      run: make vet
+
     - name: Test
       run: make test
 
@@ -79,8 +82,11 @@ jobs:
     - name: Build nebula-cert
       run: go build ./cmd/nebula-cert
 
+    - name: Vet
+      run: make vet
+
     - name: Test
-      run: go test -v ./...
+      run: make test
 
     - name: End 2 end
       run: make e2evv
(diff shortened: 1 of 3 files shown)


jasikpark:
It appears that the remaining go vet errors are in

https://github.com/slackhq/nebula/blob/1a7c57501140d326224872675dcd9193b266783e/control.go#L78-L80

It seems that updating it to ` sigChan := make(chan os.Signal, 1)` would cause the channel to be buffered, and remove the vet error. I'm not sure what value would be correct. From [reading](https://www.geeksforgeeks.org/buffered-channel-in-golang/) it seems that channels are essentially queues? So a value of `1` or `2` would probably be sufficient since we just want to catch one of those values and shutdown?

wadey (dismissed):

wadey (approved):

jasikpark:
@wadey mind merging this for me since I don't have merge perms atm?

PR #703, opened 2022-06-30
Enforce a minimum golang version with a nice error message.

brad-defined:
Only works for major versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-GOMINVERSION = 1.18
 NEBULA_CMD_PATH = "./cmd/nebula"
 GO111MODULE = on
 export GO111MODULE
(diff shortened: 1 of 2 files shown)


jasikpark (approved):
lgtm, tested this manually back when this was written

wadey:
FYI, once we upgrade to go1.21 we don't need this anymore:

> To improve forwards compatibility, Go 1.21 now reads the go line in a go.work or go.mod file as a strict minimum requirement: go 1.21.0 means that the workspace or module cannot be used with Go 1.20 or with Go 1.21rc1. This allows projects that depend on fixes made in later versions of Go to ensure that they are not used with earlier versions. It also gives better error reporting for projects that make use of new Go features: when the problem is that a newer Go version is needed, that problem is reported clearly, instead of attempting to build the code and printing errors about unresolved imports or syntax errors.

https://go.dev/doc/go1.21#tools

jasikpark:
is there a go 1.21 upgrade PR / path?

PR #721, opened 2022-07-18
add punchy.respond_delay config option

wadey:
Currently, the delay before starting the punchy.respond attempt is
hardcoded to 5 seconds. If you know your host is behind a problematic
NAT, you may want to configure this lower. This change lets you do this.

I also removed a comment from "punchy.delay" that was incorrect (the
punchy.delay option is not dependany on punchy.respond being set)


punchy.go
```diff
@@ -74,6 +75,12 @@ func (p *Punchy) reload(c *config.C, initial bool) {
 			p.l.Infof("punchy.delay changed to %s", p.GetDelay())
 		}
 	}
+	if initial || c.HasChanged("punchy.respond_delay") {
```
jasikpark, line 78: should this not be nested under the above punchy.respond configuration block?

jasikpark: hmm, fair ok

nbrownus (dismissed):

nbrownus (approved):

PR #722, opened 2022-07-18
Add nebula-cert.exe and cert files to .gitignore

JohnMaguire:
I'm not sure what, if anything, created `/cert/*.crt` and `/cert/*.key` but the `**` entries should help avoid accidental commits anywhere in the tree. (I find myself often creating test CAs, certs, etc.)

`make bin-windows` creates `nebula.exe` and `nebula-cert.exe` yet only the former previously appeared in the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,4 @@
 **.key
 **.pem
 !/examples/quickstart-vagrant/ansible/roles/nebula/files/vagrant-test-ca.key
-$/examples/quickstart-vagrant/ansible/roles/nebula/files/vagrant-test-ca.crt
+!/examples/quickstart-vagrant/ansible/roles/nebula/files/vagrant-test-ca.crt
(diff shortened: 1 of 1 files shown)


DasSkelett:
There is one .crt and one .key which would no longer be updateable after this change:
* https://github.com/slackhq/nebula/blob/master/examples/quickstart-vagrant/ansible/roles/nebula/files/vagrant-test-ca.key
* https://github.com/slackhq/nebula/blob/master/examples/quickstart-vagrant/ansible/roles/nebula/files/vagrant-test-ca.crt

JohnMaguire:
Sorry, thanks for that! I have now marked those files as explicitly allowed.

I verified there are no other files that will be impacted.

```
❯ find . -name '*.pem'
❯ find . -name '*.crt'
./examples/quickstart-vagrant/ansible/roles/nebula/files/vagrant-test-ca.crt
❯ find . -name '*.key'
./examples/quickstart-vagrant/ansible/roles/nebula/files/vagrant-test-ca.key
```

jasikpark (commented):
lgtm

nbrownus (approved):

Issue #694, opened 2022-06-24
windows: expose some method to reload config at runtime


kazzmir:
On unix systems nebula can be coerced to re-read its configuration file by sending SIGHUP to it. My understanding is Windows lacks the ability to send signals, and therefore nebula cannot take advantage of the SIGHUP method.

It would be convenient if there was some way to cause nebula to re-read its config without having to entirely kill the nebula process. Perhaps by writing to a file that nebula will frequently look for, or a named pipe: https://golangexample.com/a-windows-named-pipe-implementation-written-in-golang/

JohnMaguire:
@kazzmir While it's not as simple as SIGHUP on Linux, I did want to let you know that the built-in SSH server does have a "reload" command. You can find the configuration for the SSH server here: https://www.defined.net/nebula/config/#sshd

kazzmir:
I see, that is very cool! I will try it.

kazzmir:
The `reload` command is not supported on windows, because the ssh server simply sends itself a SIGHUP signal.
```
user@nebula > reload
not supported by windows
```
https://github.com/slackhq/nebula/blob/92dfccf01ac3019ff3e432c11654534be14c052f/ssh.go#L878

So this code would need to be able to interact with the main logic of nebula directly rather than a signal. Maybe there could be a global channel that listens for reloads, where sending SIGHUP would put a value into the channel, and ssh reload does the same.

Quick sketch:
```
reloadChannel := make(chan bool, 2)
signal.Notify(reloadChannel, syscall.HUP)
go func(){
  for {
    select {
      case <-reloadChannel:
          doReload()
    }
  }
}()
...
func sshReload(){
   reloadChannel <- true
}
```

PR #725, opened 2022-07-27
explicitly reload config from ssh command

kazzmir:
The `reload` command sends SIGHUP to the process to let it reload the config, but SIGHUP is not implemented in windows. Rather than indirectly reloading via SIGHUP, this PR explicitly invokes `c.ReloadConfig()` when the `reload` command is given in ssh.

Previous behavior on windows:
```
user@nebula > reload
not supported by windows
```
The new behavior immediately closes the ssh connection once `reload` is given because the sshd server closes, so the user never sees a message, which is probably ok but the ssh handler for `reload` could write a message back to the client just before calling `c.ReloadConfig()`

Resolves https://github.com/slackhq/nebula/issues/694

### DIFF
--- a/ssh.go
+++ b/ssh.go
@@ -877,6 +877,7 @@ func sshPrintTunnel(ifce *Interface, fs interface{}, a []string, w sshd.StringWr
 }
 
 func sshReload(c *config.C, w sshd.StringWriter) error {
+    err := w.WriteLine("Config reloaded")
     c.ReloadConfig()
-	return w.WriteLine("Config reloaded")
+	return err
 }
--- a/ssh.go
+++ b/ssh.go
@@ -877,7 +877,7 @@ func sshPrintTunnel(ifce *Interface, fs interface{}, a []string, w sshd.StringWr
 }
 
 func sshReload(c *config.C, w sshd.StringWriter) error {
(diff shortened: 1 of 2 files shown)


ssh.go
```diff
@@ -875,16 +876,7 @@ func sshPrintTunnel(ifce *Interface, fs interface{}, a []string, w sshd.StringWr
...
-	}
-	err = p.Signal(syscall.SIGHUP)
-	if err != nil {
-		return w.WriteLine(err.Error())
-		//TODO
-	}
-	return w.WriteLine("HUP sent")
+func sshReload(c *config.C, w sshd.StringWriter) error {
+    c.ReloadConfig()
+	return w.WriteLine("Config reloaded")
```
JohnMaguire, line 881: Per the PR description, IIUC this will never be seen because `c.ReloadConfig()` is a blocking call that restarts the SSH server.

Even if we change that line to `go c.ReloadConfig()` we'll be racing it. I agree then that this message should be sent prior to `c.ReloadConfig()`, e.g. "Reloading config"

nbrownus (approved):

JohnMaguire:
@kazzmir Thanks for this!

PR #726, opened 2022-07-27
Use issue templates to improve bug reports

nbrownus:
This change does 2 things:
1. When clicking `New Issue` on github, folks will be presented with a set of options. Report a bug, a link to to documentation, and the Slack invite link for the NebulaOSS team.
2. If `File a bug` is selected, folks are presented with a form to fill out.

I will keep the config at [this repo](https://github.com/nbrownus/testing/issues/new/choose) in sync so that we can experiment.

Heavily inspired by [Astro's](https://github.com/withastro/astro/issues/new/choose) setup.


wadey (approved):
I think this looks good to me, we can always improve more in the future

PR #736, opened 2022-08-27
Dockerfile support

SergeantSerk:
Hey all!

I felt the need to add a **Dockerfile** (and **.dockerignore**) to this repo seeing as how I'd want to run this on a Docker container where it can persist and restart with the host machine. Personally, this makes it so much easier than downloading the binary and creating a system service whereas a simple container can be loaded on the go.

Do note that with this **Dockerfile**, `nebula-cert` is not included, though that can be changed within the `COPY` command. Also, this uses the latest tag of `golang` for building and runtime, just to note for later if a stable release tag needs to be used if any.

Also, it would be ideal if some automated Dockerhub image publishing happens so users don't have to clone this repo and build image locally but instead, use something like `docker pull slackhq/nebula`.

I recommend trying this on different platforms too (macOS and Windows) just to ensure this works and doesn't break anything.

Tried with nebula lighthouse node deployed on Ubuntu 22.04.1 LTS Docker with current example configuration and Windows client (1 lighthouse and one client on local network).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+# The .dockerignore file excludes files from the container build process.
+#
+# https://docs.docker.com/engine/reference/builder/#dockerignore-file
+
+# Exclude locally vendored dependencies.
+vendor/
+
+# Exclude "build-time" ignore files.
+.dockerignore
+.gcloudignore
+
+# Exclude git history and configuration.
+.gitignore
--- a/.dockerignore
+++ b/.dockerignore
@@ -11,3 +11,7 @@ vendor/
 
 # Exclude git history and configuration.
 .gitignore
+
+# Ignore Dockerfile itself
+Dockerfile
+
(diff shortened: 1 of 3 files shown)


SergeantSerk:
I have just noticed that PR #68 also exists but not merged. While it may work, mine is more up-to-date, makes use of Makefile's `bin-docker` argument, is much smaller in terms of container footprint and doesn't need to deal with alpine.

SergeantSerk:
Also tried with nebula lighthouse node deployed on a Linode instance Debian 9.13 Docker with current example configuration and Windows client (1 lighthouse and one client on the Internet).

bitshop:
This specifically uses the linux-amd64 build so I think it needs to pull proper file from build either for running platform or all platforms (so that it can be published to a repo and work with aarm64, etc). 

Good work and idea.

johnmaguire:
We're going to try to get #1037 merged in an upcoming release. Going to close this out. Thanks for the help!

PR #742, opened 2022-09-02
Update example config with IPv6 note for allow lists

### LINKED ISSUE #469: Docs need to be expanded
Now that Nebula is at release 1.4 (with IPv6 as a network underlay, #369 ),
 it might be a good idea to document the IPv6 address syntax in the example config file (and/or README).

JohnMaguire:
Fixes #469.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -47,8 +47,9 @@ lighthouse:
   # allowed. You can provide CIDRs here with `true` to allow and `false` to
   # deny. The most specific CIDR rule applies to each remote. If all rules are
   # "allow", the default will be "deny", and vice-versa. If both "allow" and
-  # "deny" rules are present, then you MUST set a rule for "0.0.0.0/0" as the
-  # default.
+  # "deny" IPv4 rules are present, then you MUST set a rule for "0.0.0.0/0" as
+  # the default. Similarly if both "allow" and "deny" IPv6 rules are present,
+  # then you MUST set a rule for "::/0" as the default.
   #remote_allow_list:
     # Example to block IPs from this subnet from being used for remote IPs.
     #"172.16.0.0/12": false


examples/config.yml
```diff
@@ -47,8 +47,9 @@ lighthouse:
...
   # deny. The most specific CIDR rule applies to each remote. If all rules are
   # "allow", the default will be "deny", and vice-versa. If both "allow" and
-  # "deny" rules are present, then you MUST set a rule for "0.0.0.0/0" as the
-  # default.
+  # "deny" IPv4 rules are present, then you MUST set a rule for "0.0.0.0/0" as
+  # the default. Similarly if both "allow" and "deny" IPv6 rules are present,
+  # then you MUST set a rule for "::/0" as the default.
   #remote_allow_list:
     # Example to block IPs from this subnet from being used for remote IPs.
     #"172.16.0.0/12": false
```
IanVS, line 55: Maybe worth adding an IPv6 example here, for those of us who skim comments and just look at examples?

JohnMaguire: I considered it but wasn't sure what a useful overlapping IPv6 rule would look like exactly, and wondered if it'd be repetitive of the second IPv4 block.

If the goal is simply to show that IPv6 restrictions are possible, maybe we can replace this first example to use an IPv6 address... although I expect most use cases to be IPv4-based.

If someone wants to provide a useful looking example prior to approval I can add it in, or I'm happy to have this merge as-is.

jasikpark:
context: 

https://github.com/slackhq/nebula/blob/be360e694f7180633b4485c88e7a9eb8bd7d0aa9/allow_list.go#L151-L176

nbrownus (approved):

PR #746, opened 2022-09-14
Punchy only punch on established addrs

brad-defined:
- Punchy writes 1-byte packets to all RemoteList addresses known in its host map. For NAT / Firewall maintenance, it should only send packets to current tunnel addresses.
- Updated HostMap Punchy to reconfigure itself on config reload. Code updated to always create a Punchy goroutine, but only send packets if GetPunch() returns true.
- Added a `punchy.frequency` configuration option that can modify how frequently punchy sends out its 1-byte packets.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -123,6 +123,9 @@ punchy:
   # Continues to punch inbound/outbound at a regular interval to avoid expiration of firewall nat mappings
   punch: true
 
+  # frequency controls the regular interval of NAT punches. Defaults to 10 seconds.
+  #frequency: 10s
+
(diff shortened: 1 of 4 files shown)


brad-defined:
@nbrownus was concerned that punchy's existing behavior of sending 1-byte packets to all known RemoteList IP's aids in optimal path updating.

I updated the PR to make punchy behavior in this regard configurable.
- If punchy.target_all_remotes is true (default), Nebula's existing behavior of sending 1 byte packets to every known remote address in its hostmap is preserved.
- If punchy.target_all_remotes is false, Nebula will only send 1-byte punch packets to tunnel endpoints.

PR #747, opened 2022-09-15
Make hostinfo remote atomic, for consistent data access synchronization.

brad-defined:
When working on [746](https://github.com/slackhq/nebula/pull/746) I noticed that there was inconsistent lock access to HostInfo.remote pointer.

Inspired by [728](https://github.com/slackhq/nebula/pull/728) this PR makes HostInfo.remote an atomicPointer.


brad-defined:
oh, right. Gotta upgrade to go 1.19 first. :/

PR #750, opened 2022-09-19
Add template systemd service files

wpyoga:
These allow hosts to easily join multiple different Nebula networks.

Also, I have removed the `SyslogIdentifier=nebula` option, it's unnecessary because the logs already get tagged `nebula` by default.

### DIFF
--- a/dist/arch/nebula.service
+++ b/dist/arch/nebula.service
@@ -4,8 +4,9 @@ Wants=basic.target network-online.target
 After=basic.target network.target network-online.target
 
 [Service]
+WorkingDirectory=/etc/nebula
 ExecReload=/bin/kill -HUP $MAINPID
-ExecStart=/usr/bin/nebula -config /etc/nebula/config.yml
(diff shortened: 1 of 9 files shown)


wpyoga:
I have added a commit (to the same files) that makes `nebula` run inside the respective config directories. This enables this kind of configuration:

```yaml
pki:
  ca: ca.crt
  cert: host.crt
  key: host.key
```

Which makes it easier to write the config files. No need to write something like `key: /etc/nebula/MyNetwork2/host.key`.

IngwiePhoenix:
Would love if someone looked into this, this'd be very handy!

nbrownus:
Thanks for submitting this PR and sorry it took so long to respond.

I do agree that this is a good way to run multiple instances of nebula under systemd but I think it's enough of its own special case to warrant folks to understand the complexities of parameterized systemd unit files and this repo isn't the appropriate place to explore that topic.

Since all of the packaging distros have taken ownership of their nebula systemd unit files I am closing this PR.

PR #753, opened 2022-09-26
Use config.yml consistently (not config.yaml)

shish:
The inconsistency between the actual filename and the instructions tripped me up - it seems that _nearly_ everywhere uses `yml` rather than `yaml`, let's make it consistent :)


shish:
Blergh, I'll probably need to check with my company lawyers before signing a CLA D:

Will send a feature request and see if anybody else wants to fix it...

PR #759, opened 2022-09-29
add calculated_remotes

wadey:
This setting allows us to "guess" what the remote might be for a host while we wait for the lighthouse response. For networks that are designed with this in mind, it can help speed up handshake performance, as well as improve resiliency in the case that all lighthouses are down.

Example:

    lighthouse:
      # ...

      calculated_remotes:
        # For any Nebula IPs in 10.0.10.0/24, this will apply the mask and add
        # the calculated IP as an initial remote (while we wait for the response
        # from the lighthouse). Both CIDRs must have the same mask size.
        # For example, Nebula IP 10.0.10.123 will have a calculated remote of
        # 192.168.1.123

        10.0.10.0/24:
          - mask: 192.168.1.0/24
            port: 4242


udp/udp_all.go
```diff
@@ -64,6 +64,22 @@ func (ua *Addr) Copy() *Addr {
 	return &nu
 }
 
+type AddrSlice []*Addr
+
+func (a AddrSlice) Equal(b AddrSlice) bool {
```
adamrothman, line 69: This could ostensibly be implemented using [`slices.EqualFunc`](https://pkg.go.dev/golang.org/x/exp/slices#EqualFunc).

calculated_remote.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil, fmt.Errorf("calculated_remotes entry: %w", err)
+		}
+		l = append(l, c)
+	}
+
+	return l, nil
+}
+
+func newCalculatedRemotesEntry(raw interface{}) (*calculatedRemote, error) {
+	rawMap, ok := raw.(map[interface{}]interface{})
```
adamrothman, line 107: Looking at your example config, aren't these keys always strings (i.e. `map[string]any`)?

handshake_manager.go
```diff
@@ -153,8 +145,21 @@ func (c *HandshakeManager) handleOutbound(vpnIp iputil.VpnIp, f udp.EncWriter, l
...
+		// If we didn't return here a lighthouse could cause us to aggressively send handshakes
+		return
+	}
+
+	hostinfo.HandshakeLastRemotes = remotes
+
 	//TODO: this will generate a load of queries for hosts with only 1 ip (i'm not using a lighthouse, static mapped)
-	if hostinfo.remotes.Len(c.pendingHostMap.preferredRanges) <= 1 {
+	// So we only do it after attempting 5 handshakes already
+	if len(remotes) <= 1 && hostinfo.HandshakeCounter == 5 {
```
adamrothman, line 162: Is the idea that we only attempt this on (exactly) the 6th attempt? Or should it be `>= 5`?

adamrothman (dismissed):
A few questions but nothing blocking.

calculated_remote.go
wadey: This is specific to loading from the YAML config, which always loads them as `map[any]any` due to how it works. We could make this clearer though.

adamrothman (approved):

nbrownus (dismissed):

examples/config.yml
```diff
@@ -91,6 +91,19 @@ lighthouse:
     #- "1.1.1.1:4242"
     #- "1.2.3.4:0" # port will be replaced with the real listening port
 
+  # EXPERIMENTAL: This option may change or disappear in the future.
```
jasikpark, line 94: Can we document this with an "experimental" warning under https://nebula.defined.net/docs/config/?

nbrownus (approved):

PR #763, opened 2022-10-07
add ssh command to print device info

kazzmir:
This PR adds a new ssh command `device-info` that responds with the name of the tun device, and the cidr of the tun. Example output:
```
name=tun0 cidr=192.168.200.13/24
```
I need the tun device name because I would like to monitor the network device using other programs, and since the tun device name is dynamic, querying nebula seems like the best way to get it. I actually don't need the cidr, but since its available on the Device interface it seems easy enough to print as well.

If this seems reasonable I can add a json flag like the other print-xxx commands. Also the name of the ssh command is negotiable.


nbrownus:
Thanks for the PR! The name of the command seems fine and I believe the a json flag would be helpful if you don't mind adding it.

kazzmir:
:+1: ok will do soon

kazzmir:
@nbrownus updated now.

```
$ ssh user@localhost -p 4000 device-info
{"name":"tun0","cidr":"10.0.100.8/24"}
$ ssh user@localhost -p 4000 device-info -json
{"name":"tun0","cidr":"10.0.100.8/24"}
$ ssh user@localhost -p 4000 device-info -json -pretty
{
    "name": "tun0",
    "cidr": "10.0.100.8/24"
}
```

nbrownus:
@kazzmir looks good, mind signing the latest CLA?

kazzmir:
Sure, where do I go to sign it? Its suprisingly hard to find. Maybe the main readme should have a link to the cla.

nbrownus:
That is a wonderful question. A fresh commit might trigger the bot to prompt for it, I am searching for another solution in the mean time.

nbrownus (approved):

PR #767, opened 2022-10-16
Docs: Added instructions for storing your CA private key encrypted at rest with GPG

neonspectra:
Storing a CA private key in plaintext is a terribly risky and unsafe practice. The existing documentation makes a mild suggestion to encrypt your CA private key, but does not provide any recommendations for a streamlined way to do so.

We can use GPG to store `ca.key` as an encrypted file that stays encrypted on disk at all times and only gets decrypted in memory at the time a signing operation is performed, which is a much safer way for an end user to maintain such a sensitive file.

This document update explains the command syntax for how to encrypt your `ca.key` with an existing gpg key and how to easily interface the encrypted key with `nebula-cert` for signing operations without having to write the decrypted file to disk.


wadey:
Thanks for your PR! We merged support for native encryption of the private CA key with #386 (which is part of the v1.7 release). So please check out support for that, thanks again!

PR #769, opened 2022-10-18
Support NIST curve P256

wadey:
This change adds support for NIST curve P256. When you use `nebula-cert ca` or `nebula-cert keygen`, you can specify `-curve P256` to enable it. The curve to use is based on the curve defined in your CA certificate.

Internally, we use ECDSA P256 to sign certificates, and ECDH P256 to do Noise handshakes. P256 is not supported natively in Noise Protocol, so we define `DHP256` in the `noiseutil` package to implement support for it.

You cannot have a mixed network of Curve25519 and P256 certificates, since the Noise protocol will only attempt to parse using the Curve defined in the host's certificate.


cert/cert.go
```diff
@@ -324,22 +470,43 @@ func (nc *NebulaCertificate) CheckRootConstrains(signer *NebulaCertificate) erro
 }
 
 // VerifyPrivateKey checks that the public key in the Nebula certificate and a supplied private key match
-func (nc *NebulaCertificate) VerifyPrivateKey(key []byte) error {
+func (nc *NebulaCertificate) VerifyPrivateKey(curve Curve, key []byte) error {
```
forfuncsake, line 473: Could we use `nc.Details.Curve` instead of passing the curve in here?

wadey: Yeah I think you are correct we could do that. I think the intent was to verify the curve defined in the private key file matches the curve in the certificate, but I don't think this code is doing that (directly at least). I'll take a closer look.

wadey: I pushed a change to do the original intent, which is to verify the curves in the private key and certificate match: https://github.com/slackhq/nebula/pull/769/commits/edc283d27a54193d74168a72f054fbf5b5bf21c6

cert/cert.go
```diff
@@ -223,27 +335,61 @@ func UnmarshalEd25519PublicKey(b []byte) (ed25519.PublicKey, []byte, error) {
 }
 
 // Sign signs a nebula cert with the provided private key
-func (nc *NebulaCertificate) Sign(key ed25519.PrivateKey) error {
+func (nc *NebulaCertificate) Sign(key []byte) error {
```
nbrownus, line 338: It would be nice if we knew the curve of the private key here as well. The underlying sign is catching both cases though.

P256 used to sign a X25519 error: `input overflows the modulus`
Ed25519 used to sign a P256 error: `panic: ed25519: bad private key length: 32`

nbrownus:
I think we want to disregard any CA certificates that aren't using the same curve as the one in `pki.cert`. We may also want to add a check in handshake_ix.go for both stage 1 and 2 to determine if the curves match so that we can provide a better error message.

I rigged an e2e test that boots a curve25519 and a P256 client then tried to get one to handshake with the other. 

Error when handshaking from 25519 -> P256:
```
time="them 12:35:49.435059" level=error msg="Failed unmarshal handshake message" error="proto: illegal wireType 6" handshake="map[stage:1 style:ix_psk0]" udpAddr="10.0.0.1:4242"
```

Error when handshaking from P256->25519:
```
time="me 12:30:31.229009" level=error msg="Failed unmarshal handshake message" error="unexpected EOF" handshake="map[stage:1 style:ix_psk0]" udpAddr="10.0.0.2:4242"
```

I had to update some of the e2e code to test this, happy to add it to this branch if you would like.

wadey:
@nbrownus 

> I think we want to disregard any CA certificates that aren't using the same curve as the one in `pki.cert`.

Sounds good, no harm in adding them but also would be good to maybe log a warning that we won't be able to use them.

> We may also want to add a check in handshake_ix.go for both stage 1 and 2 to determine if the curves match so that we can provide a better error message.

This is tricky because Noise protocol doesn't send anything in the initial packet that indicates what the algorithm is. So we will usually just a random Noise packet parsing error if the curve doesn't match what we expect to be using. Not much else we can do, except add something to the log message saying "Maybe it is sending something for a different curve then I am using?"

cert/cert.go
wadey: I think I'll do a less invasive change of adding the curve to the method signature, but I would like to add a PublicKey and PrivateKey struct with the Curve next to the key to make things simpler in a future cleanup PR

wadey: added with [b4bec48](https://github.com/slackhq/nebula/pull/769/commits/b4bec4834133dfd6899596014c20bb1fc64ae81e)

nbrownus (approved):

PR #771, opened 2022-11-04
Fix a hang when shutting down Android

JohnMaguire:
When disconnecting from a network on Android sometimes the fd is not properly closed. I traced this down to a Read call in Interface.listenIn that was blocking. This change, in my testing, causes it to return immediately.

### DIFF
--- a/overlay/tun_android.go
+++ b/overlay/tun_android.go
@@ -12,6 +12,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/slackhq/nebula/cidr"
 	"github.com/slackhq/nebula/iputil"
+	"golang.org/x/sys/unix"
 )
 
 type tun struct {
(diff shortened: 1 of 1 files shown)


JohnMaguire:
Unfortunately after digging into this further it does not look like a solution. Back to the drawing board.

For a little more context I started wondering why this works. I know that the Go runtime behaves differently with blocking and non-blocking sockets. After reading the code for os.NewFile there is actually a check to see if the socket is already in non-blocking mode. According to Android SDK docs, it _should_ be, so this solution was surprising to me.

Using the following code, copied from an internal package in the Go stdlib I determined that it was indeed already set non-blocking:

```
var FcntlSyscall uintptr = syscall.SYS_FCNTL

func IsNonblock(fd int) (nonblocking bool, err error) {
       flag, _, e1 := syscall.Syscall(FcntlSyscall, uintptr(fd), uintptr(syscall.F_GETFL), 0)
       if e1 != 0 {
               return false, e1
       }
       return flag&syscall.O_NONBLOCK != 0, nil
}
```

In addition, while trying out this code the issue reared its head again. :(

PR #775, opened 2022-11-17
fix(sec): upgrade gopkg.in/yaml.v3 to 3.0.0

chncaption:
### What happened？
There are 1 security vulnerabilities found in gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
- [CVE-2022-28948](https://www.oscs1024.com/hd/CVE-2022-28948)

### What did I do？
Upgrade gopkg.in/yaml.v3 from v3.0.0-20210107192922-496545a6307b to 3.0.0 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS


wadey:
@chncaption please sign the CLA if you would like us to merge this. Otherwise, I plan to do a PR soon to update all deps and we will capture the update there.

wadey:
Also, the security issue is a false positive because even though `gopkg.in/yaml.v3` shows up in the go.mod file, it doesn't actually end up in the final binary (you can confirm with `go version -m`, only v2 ends up being used)

wadey:
Fixed with #780

PR #782, opened 2022-11-23
emit certificate.expiration_ttl_seconds metric

wadey:
This change emits a gauge metric indicating how much time until the certificate expires. The value will go negative once the certificate has expired.

I welcome feedback on the metric name, perhaps `certificate.expiration_seconds` would be better?


JohnMaguire:
This PR solves #608. There is some prior art located at #613.

nbrownus:
Yeah I think `certificate.expiration_seconds` or `certificate.ttl_seconds` is a bit cleaner than `certificate.expiration_ttl_seconds`

PR #783, opened 2022-11-26
change default CA expiration 1y -> 30y

### LINKED ISSUE #712: Quick Start neglects to mention CA validity (default : 1 year)
Would suggest the Quick Start guide at https://www.defined.net/nebula/quick-start/ makes it clear that following the instructions creates a CA that is only valid for one year by default and perhaps encourages use of the `-duration` flag.

There's no mention of this in the documentation, so anyone following the Quick Start guide will find their installation starts to fail after one year in use.

This has also been raised in a comment on https://github.com/slackhq/nebula/issues/111 more than once before;

- https://github.com/slackhq/nebula/issues/111#issuecomment-1010017853
- https://github.com/slackhq/nebula/issues/111#issuecomment-1053243080

Following the quick start guide should not, by default, produce something that will fail after a year without warning.

sneak:
This removes a major deployment footgun which needlessly cost me time and energy.

This would also fix #712 indirectly.

### DIFF
--- a/cmd/nebula-cert/ca.go
+++ b/cmd/nebula-cert/ca.go
@@ -32,7 +32,7 @@ func newCaFlags() *caFlags {
 	cf := caFlags{set: flag.NewFlagSet("ca", flag.ContinueOnError)}
 	cf.set.Usage = func() {}
 	cf.name = cf.set.String("name", "", "Required: name of the certificate authority")
-	cf.duration = cf.set.Duration("duration", time.Duration(time.Hour*8760), "Optional: amount of time the certificate should be valid for. Valid time units are seconds: \"s\", minutes: \"m\", hours: \"h\"")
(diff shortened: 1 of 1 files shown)


sneak:
I don't sign CLAs, because they're bullshit that allow for co-opting my work for profit by others.  Real open source projects don't need CLAs.  (cf Linux, gcc)

sneak:
If this were a real open source project, the MIT license would be enough.  Instead, however, Slack is requesting additional rights from me:

> Grant of Copyright License. Subject to the terms and conditions of this Agreement, You hereby grant to Slack and to recipients of software distributed by Slack a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable copyright license to reproduce, prepare derivative works of, publicly display, publicly perform, sublicense, and distribute Your Contributions and such derivative works.

I don't give anyone the right to sublicense my software.  I release all of my works under free software licenses.  That's more than enough.

PR #791, opened 2022-12-10
Add nss-lookup to the systemd wants

Fale:
Ensure DNS is running before starting nebula.
Fixes #745. Thanks a lot @JohnMaguire for the suggestion :-)
Closes #792 
Fixes #797

PR #792, opened 2022-12-12
Add nss-lookup to the systemd Wants targets

JohnMaguire:
This builds on #791 but hits two more `.service` files in the repository. Thanks @Fale!

I did notice that there are some other differences in these files:
- example and ansible scripts do not reference `network-online.target`, unlike Arch and Fedora.
- Arch and ansible do not have a `Before=sshd.target`, Fedora and example do.

### DIFF
--- a/dist/arch/nebula.service
+++ b/dist/arch/nebula.service
@@ -1,6 +1,6 @@
 [Unit]
-Description=nebula
-Wants=basic.target network-online.target
+Description=Nebula overlay networking tool
+Wants=basic.target network-online.target nss-lookup.target
 After=basic.target network.target network-online.target
 
 [Service]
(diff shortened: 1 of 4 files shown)


Fale:
On the `network-online.target`, I had the doubt that is pointless, since I would expect `nss-lokup.target` to be always after `network-online.target`. Thought, in the documentation I've not found anything that clearly states this, so I left it with the idea that worst-case is a check that always return true, since a more stricter test is ANDed to it

I don't know enough arch to say if `Before=sshd.target` is safe there or not, but I would assume it is.

JohnMaguire:
Good point on `network-online.target`. We haven't heard reports either so I'm not sweating it.

@coderobe Apologies for the random ping, but it appears you last packaged nebula for Arch. Any thoughts on the PR?

Fale:
I've bumped #791 since I've added additional fixes.

I've gone in the rabbit hole, and I can confirm that `nerwork-online.target` can happen after `nss-lookup.target`, and therefore in our case we need them both!

PR #794, opened 2022-12-13
Add note indicating modes have usage text

### LINKED ISSUE #520: add a note about sub help pages to the root '-h'
When nebula-cert is called with '-h' to display help, it does not show any information about which optional parameters all of the sub-modes (ca, sign etc) have. I discovered that I can list them by running 'nebula-cert ca -h' for example, but this was not obvious to me. It would be clearer if a note like "to see parameter for different options, please run 'nebula-cert [option] -h'" to the root help msg.

JohnMaguire:
Closes #520.


jasikpark (approved):
lgtm 👍 

would be great if `nebula-cert -h key-gen` worked as well, having `-h` flag out-of-order

( i might be wrong and it might work, didn't think it did last i checked though; golang flags is very opinionated about flag order )

jasikpark (approved):

nbrownus (approved):

Issue #745, opened 2022-09-05
🐛 BUG: Nebula crash on "Static host address could not be parsed"


Fale:
### What version of `nebula` are you using?

1.6.0

### What operating system are you using?

Linux

### Describe the Bug

I use FQDN for lighthouse addresses.
When Nebula starts, if the network is not yet fully up, the DNS resolution fails and Nebula crashes.
My expectation would be that Nebula does _not_ crash and retries over time in the same way it does when it can resolve the FQDN to an IP but Nebula is not able to establish a connection.

### Logs from affected hosts

```
-- Boot 6ba68dc9aa554b0094f3a442acb37401 --
Sep 05 09:22:42 x1.fale.io nebula[1259]: time="2022-09-05T09:22:42+02:00" level=info msg="Firewall rule added" firewallRule="map[caName: caSha: direction:outgoing endPort:0 groups:[] host:any ip: proto:0 startPort:0]"
Sep 05 09:22:42 x1.fale.io nebula[1259]: time="2022-09-05T09:22:42+02:00" level=info msg="Firewall rule added" firewallRule="map[caName: caSha: direction:incoming endPort:0 groups:[] host:any ip: proto:1 startPort:0]"
Sep 05 09:22:42 x1.fale.io nebula[1259]: time="2022-09-05T09:22:42+02:00" level=info msg="Firewall rule added" firewallRule="map[caName: caSha: direction:incoming endPort:22 groups:[fale ssh] host: ip: proto:6 startPort:22]"
Sep 05 09:22:42 x1.fale.io nebula[1259]: time="2022-09-05T09:22:42+02:00" level=info msg="Firewall started" firewallHash=e9fc6276b1f92e11afb6da6c768e57eae28a9aaab95b14daadb95cbce6f432af
Sep 05 09:22:42 x1.fale.io nebula[1259]: time="2022-09-05T09:22:42+02:00" level=info msg="Main HostMap created" network=192.168.100.32/24 preferredRanges="[]"
Sep 05 09:22:42 x1.fale.io nebula[1259]: time="2022-09-05T09:22:42+02:00" level=info msg="UDP hole punching enabled"
Sep 05 09:22:42 x1.fale.io nebula[1259]: time="2022-09-05T09:22:42+02:00" level=error msg="Static host address could not be parsed" error="lookup lh4.fale.io: Temporary failure in name resolution" vpnIp=192.168.100.4
```

### Config files from affected hosts

```
...
static_host_map:
  192.168.100.1:
[...]

Issue #273, opened 2020-08-07
Feature Request: Support to resolve lighthouse domain name(DDNS) when reconnect or connect failed


xiaolei0125:
My lighthouse used a public ip but not fixed,  ip changed every day, so I used a  domain name(support DDNS) in static_host_map;
`static_host_map:
  "192.168.10.1": ["lh1-nb.myddns.top:4242"]`

When nebula start,  can get correct lighthouse ip and work well,  but when my lighthouse public ip changed, other nebula node lost connecttion with lighthouse, since  nebula still used old lighthouse ip to connecting, so connecting failed.

Is possible to resolve lighthouse domain name again when reconnect or connect failed ?

xiaolei0125:
same feature as [#294 ](url)

aa51513:
What is the latest status of this issue?

If you are worried that the dns resolution result contains multiple IP addresses, you can directly select the first one, which is better than doing nothing, isn't it?

JohnMaguire:
Hi all - I just wanted to drop a quick comment to mention that the fix from #796 was released in v1.7.0. We found a bug in the behavior there and quickly pushed out v1.7.1 which should re-query domains in the `static_host_map`. Please give it a shot and let us know what you think!

Issue #176, opened 2020-01-27
Domain name in static_host_map


enykeev:
Using domain name instead of ip in static_host_map may result in nebula ending up in a state when the process is still running, but no attempts to reconnect to lighthouse will ever occur.

The situation happens during initial boot when interface is already up and `network-online.target` seem to be fulfilled, but `dhcpcd` is yet to set a nameserver. If nebula is to start in that time frame (which in my case it does 9 out of 10 times), it won't be able to parse lighthouse IP, report lighthouse unreachable due to missing static_host_map entry (see #41) and continue to run while making no attempts to establish connection.

In my opinion, if we consider #41 to be a configuration issue, the error should be fatal. If on the other hand we consider it an operational issue, we should resolve the host at connect time rather than during config parsing.

bartmichu:
I have the same problem on my Ubuntu laptop. I need to restart nebula service to get it working every time after system boots up.

bartmichu:
Nebula v1.2.0 is still affected.

windwalker78:
I confirm this is still not working 100% on 1.2.0. Seen on Ubuntu and Debian 9.

brianluong4765:
Confirmed on CentoOS 7

keitme:
This is with the current 1.3.0 release.

I have reproduced this issue on several Windows revisions (10, server 2016, server 2019). On Windows, as a workaround, I have manually set the service start up type for Nebula Network Service to Automatic (Delayed). With this set the connection establishes properly when using the FQDN of the lighthouse server rather than its IP address, but the connection may not be available until several minutes after the machine is otherwise available.

I have reproduced this issue on both the latest Mac OS X revisions as well, but I haven't developed a workaround as of yet so I am specifying the IP address of the lighthouse. I have not tested on Mac OS 11 as of yet.

rawdigits:
The only useful solution here would probably involve regularly re-querying names. If we go down this path, we should also consider re-querying even after we get a successful answer, as this would allow us to migrate to a new IP if the underlying DNS entry changes.

SgtZapper:
I think I might have encountered the same problem after a power outage had my single lighthouse unreachable for a few hours and when it came back on a new public IP none of the nodes that where connected would never reconnect without restarting nebula on them.

It would be good to try to resolve the dns entry for a host in the static hosts every now and then when a static host is unreachable.

JohnMaguire:
[...]

PR #796, opened 2022-12-14
Dns static lookerupper

brad-defined:
Fix issue #745 and #176

On startup, Nebula adds all the static_host_map entries to its internal RemoteList object for Nebula IP-to-internet-IP translation. If the static_host_map includes a hostname, Nebula attempts a DNS lookup inline for that hostname. That DNS lookup function will return at most 1 address for the entry. If the DNS lookup fails, Nebula exits. The DNS query is never re-executed for the lifetime of the Nebula process.

This PR introduces the following changes:
- For static_host_map entries with hostnames, a background goroutine runs in a loop, looking up the DNS entries. If the results are different, it kicks off a RemoteList rebuild, allowing Nebula to consume the newly returned addresses.
- For static_host_map entries without a hostname (meaning, the list only includes IP addresses), no goroutine is created.
- Since the DNS lookup is not in the startup path, a DNS lookup error results in a log message, but Nebula continues to run. If the lookup that failed was for the lighthouse, Nebula will not be able to connect to the lighthouse and participate in the Nebula network until the DNS loop runs again and successfully resolves the address. (I made this default to 5 minutes, but maybe we want the loop to run more frequently to cover this case, or make it retry faster if it hit a DNS lookup failure.)
- hidden config options are included which allow for the configuration of DNS lookup timeouts, the loop cadence, and which networks are supported
- the DNS function used will return all returned DNS entries for the given network, rather than being limited to 1 IP.
- the network can exclude ipv6 addresses (which are problematic for lighthouses in particular, as a host connecting to a lighthouse over an IPv6 address prevents the lighthouse from learning the host's IPv4 NAT address)


JohnMaguire:
This fixes #273. I believe #745 and #176 are solved more simply through #791.

brad-defined:
rebased

brianbraunstein:
huge thank you for this one. would be great if it was documented.

PR #804, opened 2023-01-13
Generic timerwheel

nbrownus:
Implements a generic `TimerWheel` and removes the need for the `SystemTimerWheel`, enabling timer wheels with other storage types. Also cleans up some comments and exports necessary functions.


timeout.go
```diff
@@ -26,31 +22,33 @@ type TimerWheel struct {
...
+	Head *TimeoutItem[T]
+	Tail *TimeoutItem[T]
 }
 
 // TimeoutItem Represents an item within a tick
-type TimeoutItem struct {
-	Packet firewall.Packet
-	Next   *TimeoutItem
+type TimeoutItem[T any] struct {
+	Item T
```
brad-defined, line 43: Did you consider making `Item` type `any`?
```
type TimeoutList struct {
        Head *TimeoutItem
        Tail *TimeoutItem
}

type TimeoutItem struct {
        Item any
        Next *timeoutItem
}
```

timeout.go
```diff
@@ -61,25 +59,25 @@ func NewTimerWheel(min, max time.Duration) *TimerWheel {
...
 	return &tw
 }
 
-// Add will add a firewall.Packet to the wheel in it's proper timeout
-func (tw *TimerWheel) Add(v firewall.Packet, timeout time.Duration) *TimeoutItem {
+// Add will add an item to the wheel in its proper timeout
+func (tw *TimerWheel[T]) Add(v T, timeout time.Duration) *TimeoutItem[T] {
 	// Check and see if we should progress the tick
-	tw.advance(time.Now())
+	tw.Advance(time.Now())
```
brad-defined, line 80: Lost to the sands of time, timeout_system does not advance during Add. It has the line of code to advance commented out. Git has no history why. (I can speculate that Add is called in the hot path, and wheel work is desired to be avoided in the hot path.)
It is a difference between the timeout.go timer wheel and timeout_system timer wheel which might be a wrinkle in using one for both - unless the advance during Add becomes some kind of config parameter.

timeout_system.go
```diff
@@ -1,199 +0,0 @@
...
-
-	return &tw
-}
-
-func (tw *SystemTimerWheel) Add(v iputil.VpnIp, timeout time.Duration) *SystemTimeoutItem {
-	tw.lock.Lock()
-	defer tw.lock.Unlock()
-
-	// Check and see if we should progress the tick
-	//tw.advance(time.Now())
```
brad-defined, line 85: here's where the new timer wheel differs.

timeout.go
nbrownus: Not threading `T` through to it requires a type assertion at `TimerWheel.Purge` and incurs a small performance penalty.

```
name                 old time/op  new time/op  delta
TimerWheel_Purge-10  2.49ns ± 0%  2.63ns ± 0%  +5.69%  (p=0.000 n=19+17)
```

There's also a bit of awkwardness since `TimerWheel.Add` currently returns the added `TimeoutItem`. `TimeoutItem.Item` could be changed to something that can't be casted to `T`. Nothing uses the returned `TimeoutItem` today so we could remove that.

timeout.go
nbrownus: Updated to allow the `HandshakeManager` and `ConnectionManager` to continue with the previous behavior of `Advance`ing on their own

timeout.go
```diff
@@ -61,26 +65,31 @@ func NewTimerWheel(min, max time.Duration) *TimerWheel {
 	// timeout
 	wLen := int((max / min) + 2)
 
-	tw := TimerWheel{
+	tw := TimerWheel[T]{
```
brad-defined, line 68: nit: I didn't like the fact that `lastTick` isn't initialized until Advance() is called. The net effect is that delays of events placed in the timer wheel don't take effect until after the first Advance() is called. So, if I place a 10s event in the wheel, and then call Advance after 5s, the event doesn't Purge until 15s after it was initially placed in the wheel.

brad-defined (approved):

PR #805, opened 2023-01-13
WIP: Track connections by index id instead of vpn ip

nbrownus:
Changes the connection manager to track tunnels based on their local index instead of the vpn ip.

Both `HostMap.Hosts` and `HostMap.Indexes` pointers are still removed, matching the previous behavior. This will enable a future PR to keep the `HostMap.Hosts` pointer in cases where its `hostinfo` does not point to the same local index as `HostMap.Indexes` which means we can do more graceful handshake race resolution.

Depends on #804 which after it merges this branch can be rebased and pointed to `master`

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/slackhq/nebula/header"
-	"github.com/slackhq/nebula/iputil"
 )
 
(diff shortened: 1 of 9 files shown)


jasikpark (commented):
looks reasonable to me, but i don't understand this code to the point of knowing possible erorrs other than simple syntax / logic errors

PR #807, opened 2023-01-18
Track connections by local index id instead of vpn ip

nbrownus:
Changes the connection manager to track tunnels based on their local index instead of the vpn ip.

Both `HostMap.Hosts` and `HostMap.Indexes` pointers are still removed, matching the previous behavior. This will enable a future PR to keep the `HostMap.Hosts` pointer in cases where its hostinfo does not point to the same local index as `HostMap.Indexes` which means we can do more graceful handshake race resolution.

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -274,13 +274,13 @@ func (n *connectionManager) handleInvalidCertificate(now time.Time, hostinfo *Ho
 		return false
 	}
 
-	valid, _ := remoteCert.Verify(now, n.intf.caPool)
+	valid, err := remoteCert.Verify(now, n.intf.caPool)
 	if valid {
 		return false
 	}
 
 	fingerprint, _ := remoteCert.Sha256Sum()
-	hostinfo.logger(n.l).
+	hostinfo.logger(n.l).WithError(err).
 		WithField("fingerprint", fingerprint).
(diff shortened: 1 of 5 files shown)


connection_manager.go
```diff
@@ -280,22 +274,21 @@ func (n *connectionManager) handleInvalidCertificate(now time.Time, vpnIp iputil
...
-	valid, err := remoteCert.Verify(now, n.intf.caPool)
+	valid, _ := remoteCert.Verify(now, n.intf.caPool)
 	if valid {
 		return false
 	}
 
 	fingerprint, _ := remoteCert.Sha256Sum()
-	n.l.WithField("vpnIp", vpnIp).WithError(err).
-		WithField("certName", remoteCert.Details.Name).
+	hostinfo.logger(n.l).
```
wadey, line 283: This no longer logs the reason the cert was rejected, is that intentional?

wadey (commented):
One question about the logging changes

connection_manager.go
nbrownus: Nice catch, was not intentional

wadey (approved):
LGTM, I was concerned originally we could have a hidden leak in here if something didn't get cleaned up properly, but I have no reason to think that's actually going to happen.

PR #812, opened 2023-01-23
Build for openbsd

nbrownus:
Adds a build for `openbsd` and restructures a few things to make supporting bsd family of OS easier.

We should think about this change in relation to #808 

`openbsd` was requested in #806

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ name: Create release and upload binaries
 
 jobs:
   build-linux:
-    name: Build Linux All
+    name: Build Linux/BSD All
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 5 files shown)


JohnMaguire:
Rebased on master in order to pull in changes from #808 

Would be nice to get this merged. We have a user on Nebula OSS Slack who'd like to try running on OpenBSD.

JohnMaguire:
@wadey did a little bit of work towards this goal in https://github.com/wadey/nebula/tree/smoke-openbsd

To run it with Virtualbox and Vagrant, use `make smoke-vagrant/openbsd-amd64`

It's currently failing with:

```
time="2023-01-23T21:48:50Z" level=fatal msg="failed to run 'route add': exit status 1"
```

@wadey thinks the following may solve the issue: `sudo /sbin/route -n add -inet 192.168.100.3/24 192.168.100.3`

```
[vagrant@openbsd7 ~]$ route get 192.168.100.1
   route to: 192.168.100.1
destination: 192.168.100.0
       mask: 255.255.255.0
    gateway: 192.168.100.3
  interface: tun0
 if address: 192.168.100.3
   priority: 8 (static)
      flags: <UP,GATEWAY,DONE,STATIC>
     use       mtu    expire
       1         0         0
```

6167656e74323431:
I think the command you probably want here is `/sbin/route -n add -net 192.168.100.3/24 -link -iface tun0`.

wadey:
We should add `release-openbsd`  to the Makefile and add it to the release workflow as well here:

- https://github.com/slackhq/nebula/blob/e5af94e27a6907c5adbe5eea2649e480bce1af89/.github/workflows/release.yml#L22

wadey (approved):

PR #820, opened 2023-03-02
Remove handshake race avoidance

nbrownus:
This change removes handshake race avoidance and allows for multiple tunnels between 2 hosts to exist.

The main drivers for this are
- Enables future re-handshaking PR. We can start a handshake even if there is an existing tunnel to the remote. If the re-handshake succeeds then both tunnels remain active until `ConnectionManager` determines one is no longer in use.
- Allows for rapid recovery in an unclean shutdown, a previous race loser will now get the same benefits as a previous race winner. 
- Opens the possibility for rapid recovery when roaming by detecting the network change and re-handshaking. This is `rebind` on mobile.

`ConnectionManager` will try to guide host pairs that have more than 1 tunnel back to a single tunnel.

The logic is basically:
1. The hostinfo is receiving traffic
2. The hostinfo is not primary
3. The vpn of the hostinfo is greater than our own

Responders will begin tracking tunnels at the completion of a stage 1 packet. This is necessary to get `ConnectionManager` to try and coerce the state back to a single tunnel if multiple exists. A side effect will be that unused tunnels are torn down, which will help cull dead tunnel on lighthouses.

`HandshakeManager` now has another map which tracks what should be in its wheel. This may have been the cause of the TODO note in that file about fast firing handshakes. Rapid fire handshakes would occur in the unclean shutdown e2e tests since it would add the same vpn ip to the pending hostmap before the timer wheel fired.

Relays should behave similar to today. If a `CreateRelayRequest` is received, the hostinfo that it was received on is immediately made the primary hostinfo. This side steps the `ConnectionManager` logic because we need to establish the relay state on the correct hostinfo and we just received the packet, so we know its good a tunnel.

A new flag, `-by-index`, is added to both `list-hostmap` and `list-pending-hostmap` ssh commands. Returns hostinfos that reside in the index map since there may be multiple for a given vpn ip and the primary hosts map would not show them.

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -181,6 +181,14 @@ func (n *connectionManager) HandleMonitorTick(now time.Time, p, nb, out []byte) 
 			continue
 		}
 
+		// Does the vpnIp point to this hostinfo or is it ancillary? If we have ancillary hostinfos then we need to
+		// decide if this should be the main hostinfo if we are seeing traffic on it
+		primary, _ := n.hostMap.QueryVpnIp(hostinfo.vpnIp)
+		mainHostInfo := true
+		if primary != nil && primary != hostinfo {
+			mainHostInfo = false
+		}
+
 		// If we saw an incoming packets from this ip and peer's certificate is not
 		// expired, just ignore.
 		if traf {
@@ -191,6 +199,20 @@ func (n *connectionManager) HandleMonitorTick(now time.Time, p, nb, out []byte) 
 			}
 			n.ClearLocalIndex(localIndex)
 			n.ClearPendingDeletion(localIndex)
+
+			if !mainHostInfo {
+				if hostinfo.vpnIp > n.intf.myVpnIp {
+					// We are receiving traffic on the non primary hostinfo and we really just want 1 tunnel. Make
+					// This the primary and prime the old primary hostinfo for testing
+					n.hostMap.MakePrimary(hostinfo)
+					n.Out(primary.localIndexId)
+				} else {
+					// This hostinfo is still being used despite not being the primary hostinfo for this vpn ip
+					// Keep tracking so that we can tear it down when it goes away
+					n.Out(hostinfo.localIndexId)
+				}
+			}
+
 			continue
 		}
 
@@ -198,7 +220,7 @@ func (n *connectionManager) HandleMonitorTick(now time.Time, p, nb, out []byte) 
 			WithField("tunnelCheck", m{"state": "testing", "method": "active"}).
 			Debug("Tunnel status")
 
-		if hostinfo != nil && hostinfo.ConnectionState != nil {
+		if hostinfo != nil && hostinfo.ConnectionState != nil && mainHostInfo {
 			// Send a test packet to trigger an authenticated tunnel test, this should suss out any lingering tunnel issues
 			n.intf.sendMessageToVpnIp(header.Test, header.TestRequest, hostinfo, p, nb, out)
 
--- a/hostmap.go
+++ b/hostmap.go
@@ -180,6 +180,9 @@ type HostInfo struct {
 
 	lastRoam       time.Time
 	lastRoamRemote *udp.Addr
+
+	// Used to track other hostinfos for this vpn ip since only 1 can be primary (used for xmit)
+	next, prev *HostInfo
 }
 
 type ViaSender struct {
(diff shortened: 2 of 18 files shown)


hostmap.go
```diff
@@ -429,29 +432,83 @@ func (hm *HostMap) DeleteRelayIdx(localIdx uint32) {
...
 }
 
+func (hm *HostMap) MakePrimary(hostinfo *HostInfo) {
+	hm.Lock()
+	defer hm.Unlock()
+	hm.unlockedMakePrimary(hostinfo)
+}
+
+func (hm *HostMap) unlockedMakePrimary(hostinfo *HostInfo) {
+	//TODO: should we try and coalesce the relay state from oldHostinfo here?
```
brad-defined, line 442: I think the relays should be tied to a single HostInfo object, and not migrated from one HostInfo instance to another.

Relay packets are authenticated with the keys stored in the HostInfo ConnectionState, and each HostInfo for the same vpn IP will have different encryption/decryption keys. Knowing which keys to use, and when it's OK to switch from one HostInfo's keys to another would be tricky:tm:.

Relay state is also shared between HostInfo objects on different peers. Migrating RelayState from one HostInfo object instance to another HostInfo instance on one peer implies that the HostInfo on the remote peer also has the corresponding RelayState. However, it's possible that a handshake that resulted in 2 HostInfo objects on _this_ peer is due to the _remote_ peer restarting its Nebula process, losing its initial HostInfo state. In that case, _this_ peer would migrate the Relay state that doesn't actually exist on the _remote_ peer.

I believe the best way to detect this kind of divergence is by way of the ConnectionState keys. As long as the remote peer still has the ConnectionState keys, _this_ peer can reliably conclude that the _remote_ peer also still has associated RelayState info, and vice versa.

hostmap.go
```diff
@@ -520,15 +577,22 @@ func (hm *HostMap) queryVpnIp(vpnIp iputil.VpnIp, promoteIfce *Interface) (*Host
...
 		remoteCert := hostinfo.ConnectionState.peerCert
 		dnsR.Add(remoteCert.Details.Name+".", remoteCert.Details.Ips[0].IP.String())
 	}
 
+	existing := hm.Hosts[hostinfo.vpnIp]
 	hm.Hosts[hostinfo.vpnIp] = hostinfo
+
+	if existing != nil {
+		hostinfo.next = existing
+		existing.prev = hostinfo
```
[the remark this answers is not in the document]
nbrownus: The expectation here is that the caller has the hostmap lock and only `next` and `prev` are accessed when doing hostmap operations. We could lock the hostinfos but it would be duplicative of the hostmap lock.

brad-defined (approved):

hostmap.go
```diff
@@ -520,15 +576,22 @@ func (hm *HostMap) queryVpnIp(vpnIp iputil.VpnIp, promoteIfce *Interface) (*Host
...
+func (hm *HostMap) unlockedAddHostInfo(hostinfo *HostInfo, f *Interface) {
 	if f.serveDns {
 		remoteCert := hostinfo.ConnectionState.peerCert
 		dnsR.Add(remoteCert.Details.Name+".", remoteCert.Details.Ips[0].IP.String())
 	}
 
+	existing := hm.Hosts[hostinfo.vpnIp]
 	hm.Hosts[hostinfo.vpnIp] = hostinfo
+
+	if existing != nil {
```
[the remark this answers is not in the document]
nbrownus: Added a limit of 5

hostmap.go
```diff
@@ -429,29 +433,81 @@ func (hm *HostMap) DeleteRelayIdx(localIdx uint32) {
...
+	if oldHostinfo == nil {
+		return
+	}
+
+	hostinfo.next = oldHostinfo
+	oldHostinfo.prev = hostinfo
+	hostinfo.prev = nil
+}
+
 func (hm *HostMap) unlockedDeleteHostInfo(hostinfo *HostInfo) {
```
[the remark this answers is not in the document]
nbrownus: Will only free lighthouse state on the final hostinfo being closed in outside.go now

hostmap.go
nbrownus: Will be handled in a very near future PR. The current failure scenarios match the existing failure scenarios in master to the best of my knowledge/testing.

connection_manager.go
```diff
@@ -181,6 +181,14 @@ func (n *connectionManager) HandleMonitorTick(now time.Time, p, nb, out []byte)
 			continue
 		}
 
+		// Does the vpnIp point to this hostinfo or is it ancillary? If we have ancillary hostinfos then we need to
+		// decide if this should be the main hostinfo if we are seeing traffic on it
+		primary, _ := n.hostMap.QueryVpnIp(hostinfo.vpnIp)
+		mainHostInfo := true
+		if primary != nil && primary != hostinfo {
+			mainHostInfo = false
+		}
```
wadey, line 190: Is it okay that we don't have a lock on hostMap here? (something else could add an entry to the hostMap for this vpnIP after we do the query)

hostmap.go
```diff
@@ -395,9 +403,15 @@ func (hm *HostMap) DeleteReverseIndex(index uint32) {
...
-func (hm *HostMap) DeleteHostInfo(hostinfo *HostInfo) {
+// DeleteHostInfo will fully unlink the hostinfo and return true if it was the final hostinfo for this vpn ip
+func (hm *HostMap) DeleteHostInfo(hostinfo *HostInfo) bool {
 	// Delete the host itself, ensuring it's not modified anymore
 	hm.Lock()
+	final := true
+	if hostinfo.next != nil || hostinfo.prev != nil {
+		// If we have a previous or next hostinfo then we are not the last one for this vpn ip
+		final = false
+	}
```
wadey, line 414: Just a nit, but maybe a little clearer?

```suggestion
        // If we have a previous or next hostinfo then we are not the last one for this vpn ip
	final := (hostinfo.next == nil && hostinfo.prev == nil)
```

hostmap.go
```diff
@@ -537,6 +612,16 @@ func (hm *HostMap) addHostInfo(hostinfo *HostInfo, f *Interface) {
...
 			Debug("Hostmap vpnIp added")
 	}
+
+	i := 1
+	check := hostinfo
+	for check != nil {
+		if i > MaxHostInfosPerVpnIp {
+			hm.unlockedDeleteHostInfo(check)
+		}
+		check = check.next
```
wadey, line 622: I think unlockedDeleteHostInfo will set check.next to nil, so do we need to re-order to grab check.next just in case its non-nil before unlockedDeleteHostInfo erases it?

In theory it can't be non-nil since the check would have run last time and prevented there being more entries I think.

connection_manager.go
nbrownus: We have the lock in `QueryVpnIp` so we will be able to determine if this hostinfo, that is seeing traffic, was primary or not at that instant.

The only things that I think can race here are a `closeTunnel and a new handshake completing. In both cases there they will cause a primary swap, so maybe there will be some jitter in the primary for a few connection manager ticks?

In short I think it should be fine :tm:

PR #825, opened 2023-03-15
Allow `listen.host` to contain names

### LINKED ISSUE #817: Feature: `listen.host` works with dns names
`listen.host` should take dns names and resolve them at start time for the bind address. This would support systems like fly.io that need to bind against a specific address for return traffic to work correctly.

nbrownus:
On some networks (like fly.io, DO) it is desirable to listen on specific address and the easiest way to discover that address is by using a known name.

Closes #817

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -107,7 +107,7 @@ lighthouse:
 # Port Nebula will be listening on. The default here is 4242. For a lighthouse node, the port should be defined,
 # however using port 0 will dynamically assign a port and is recommended for roaming nodes.
 listen:
-  # To listen on both any ipv4 and ipv6 use "[::]"
+  # To listen on both any ipv4 and ipv6 use "::"
   host: 0.0.0.0
   port: 4242
   # Sets the max number of packets to pull from the kernel for each syscall (under systems that support recvmmsg)
--- a/main.go
+++ b/main.go
@@ -151,8 +151,21 @@ func Main(c *config.C, configTest bool, buildVersion string, logger *logrus.Logg
 	port := c.GetInt("listen.port", 0)
 
 	if !configTest {
+		rawListenHost := c.GetString("listen.host", "0.0.0.0")
+		var listenHost *net.IPAddr
+		if rawListenHost == "[::]" {
+			// Old guidance was to provide the literal `[::]` in `listen.host` but that won't resolve.
+			listenHost = &net.IPAddr{IP: net.IPv6zero}
+
+		} else {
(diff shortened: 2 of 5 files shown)


examples/config.yml
```diff
@@ -107,7 +107,7 @@ lighthouse:
 # Port Nebula will be listening on. The default here is 4242. For a lighthouse node, the port should be defined,
 # however using port 0 will dynamically assign a port and is recommended for roaming nodes.
 listen:
-  # To listen on both any ipv4 and ipv6 use "[::]"
```
jasikpark, line 110: ohh I guess this is the zero syntax w/o a port number, but the `[` and `]` are optional?

main.go
```diff
@@ -151,8 +151,21 @@ func Main(c *config.C, configTest bool, buildVersion string, logger *logrus.Logg
...
 
 	if !configTest {
+		rawListenHost := c.GetString("listen.host", "0.0.0.0")
+		var listenHost *net.IPAddr
+		if rawListenHost == "[::]" {
+			// Old guidance was to provide the literal `[::]` in `listen.host` but that won't resolve.
+			listenHost = &net.IPAddr{IP: net.IPv6zero}
+
+		} else {
+			listenHost, err = net.ResolveIPAddr("ip", rawListenHost)
```
jasikpark, line 161: https://pkg.go.dev/net#ResolveIPAddr

> The address parameter can use a host name, but this is not recommended, because it will return at most one of the host name's IP addresses.

I guess that doesn't matter?

udp/udp_generic.go
```diff
@@ -23,9 +23,9 @@ type Conn struct {
 	l *logrus.Logger
 }
 
-func NewListener(l *logrus.Logger, ip string, port int, multi bool, batch int) (*Conn, error) {
+func NewListener(l *logrus.Logger, ip net.IP, port int, multi bool, batch int) (*Conn, error) {
 	lc := NewListenConfig(multi)
-	pc, err := lc.ListenPacket(context.TODO(), "udp", fmt.Sprintf("%s:%d", ip, port))
+	pc, err := lc.ListenPacket(context.TODO(), "udp", net.JoinHostPort(ip.String(), fmt.Sprintf("%v", port)))
```
jasikpark, line 28: commenting these function docs for my own understanding, sorry if this is too much:

https://pkg.go.dev/net#JoinHostPort

jasikpark (commented):
nice

wadey (approved):

jasikpark:
any docs on this feature to add to https://nebula.defined.net?

PR #827, opened 2023-03-16
Fix relay

brad-defined:
Relay creation may race, meaning a host that has sent a CreateRelayRequest may receive a CreateRelayRequest message for the exact same tunnel, before receiving a CreateRelayResponse. When this happens, a relay's locally stored remote index will be 0. When a corresponding CreateRelayRequest arrives, it'll have a non-zero value for that remote index. The current code on Master detects this mismatch as a broken relay, and tears down the relay state. Destroying this relay state breaks all possible connectivity for the tunnel.

To work in all cases, once a relay's index is created and advertised (in a CreateRelayRequest), it should never be torn down until the associated HostInfo is destroyed. This PR updates the handling of CreateRelayRequests to avoid tearing down relay state in these detected mismatch cases.

I also added some synchronization around the Relay struct, using an atomic pointer.

I edited an existing test that could reproduce errors on Master, but which runs correctly with these updates. (Nate made the test.)

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -50,13 +50,6 @@ type Relay struct {
 	PeerIp      iputil.VpnIp
 }
 
-// For synchronization, treat the Relay struct as immutable. To edit the Relay
-// struct, make a copy of an existing value, edit the fileds in the copy, and
-// then store a pointer to the copy in RelayWrapper.
-type RelayWrapper struct {
-	Relay atomic.Pointer[Relay]
-}
-
 type HostMap struct {
 	sync.RWMutex    //Because we concurrently read and write to our maps
 	name            string
@@ -70,12 +63,15 @@ type HostMap struct {
 	l               *logrus.Logger
 }
 
+// For synchronization, treat the pointed-to Relay struct as immutable. To edit the Relay
+// struct, make a copy of an existing value, edit the fileds in the copy, and
+// then store a pointer to the new copy in both realyForBy* maps.
 type RelayState struct {
 	sync.RWMutex
 
-	relays        map[iputil.VpnIp]struct{}      // Set of VpnIp's of Hosts to use as relays to access this peer
-	relayForByIp  map[iputil.VpnIp]*RelayWrapper // Maps VpnIps of peers for which this HostInfo is a relay to some Relay info
-	relayForByIdx map[uint32]*RelayWrapper       // Maps a local index to some Relay info
+	relays        map[iputil.VpnIp]struct{} // Set of VpnIp's of Hosts to use as relays to access this peer
+	relayForByIp  map[iputil.VpnIp]*Relay   // Maps VpnIps of peers for which this HostInfo is a relay to some Relay info
+	relayForByIdx map[uint32]*Relay         // Maps a local index to some Relay info
 }
 
 func (rs *RelayState) DeleteRelay(ip iputil.VpnIp) {
@@ -88,11 +84,7 @@ func (rs *RelayState) GetRelayForByIp(ip iputil.VpnIp) (*Relay, bool) {
 	rs.RLock()
 	defer rs.RUnlock()
 	r, ok := rs.relayForByIp[ip]
-	var res *Relay
-	if ok {
-		res = r.Relay.Load()
-	}
-	return res, ok
+	return r, ok
 }
 
 func (rs *RelayState) InsertRelayTo(ip iputil.VpnIp) {
@@ -134,11 +126,10 @@ func (rs *RelayState) CopyRelayForIdxs() []uint32 {
 func (rs *RelayState) RemoveRelay(localIdx uint32) (iputil.VpnIp, bool) {
 	rs.Lock()
 	defer rs.Unlock()
-	relay, ok := rs.relayForByIdx[localIdx]
+	r, ok := rs.relayForByIdx[localIdx]
 	if !ok {
 		return iputil.VpnIp(0), false
 	}
-	r := relay.Relay.Load()
 	delete(rs.relayForByIdx, localIdx)
 	delete(rs.relayForByIp, r.PeerIp)
 	return r.PeerIp, true
@@ -151,10 +142,11 @@ func (rs *RelayState) CompleteRelayByIP(vpnIp iputil.VpnIp, remoteIdx uint32) bo
 	if !ok {
 		return false
 	}
-	newRelay := *r.Relay.Load()
+	newRelay := *r
 	newRelay.State = Established
 	newRelay.RemoteIndex = remoteIdx
-	r.Relay.Store(&newRelay)
+	rs.relayForByIdx[r.LocalIndex] = &newRelay
+	rs.relayForByIp[r.PeerIp] = &newRelay
 	return true
 }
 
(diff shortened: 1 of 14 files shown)


brad-defined:
Pushed some updates:

(1) Sometimes a HostInfo that is not the main HostInfo sees traffic, which prevents it from being deleted by ConnectionManager. Since no more outgoing traffic is sent over this non-main HostInfo, it never cycles back through ConnectionManager. PR updates ConnectionManager to place a HostInfo back onto the watch list if traffic is seen as received during the deletion phase.
(2) When tearing down all tunnels, we must iterate all Indexes, not all Hosts (which will not return non-main HostInfos).
(3) Included a test that @nbrownus wrote that sussed out some races through relays
(4) Sped up some deletion timers during tests to 4 seconds, in order to make tests that wait for hostinfos to be removed to run faster.
(5) Included some locks around hostmap access during tests, which occasionally raced with Nebula code accessing the same main host maps.
(6) Modified HostInfo teardown code to take multiple host infos into account. Only removes Relay state if the HostInfo being deleted is the final HostInfo for that vpnIp in the hostmap.
(7) Updated tests to not return a copy of RelayState, as this could copy a locked mutex and deadlock the tests.

hostmap.go
```diff
@@ -50,6 +50,13 @@ type Relay struct {
 	PeerIp      iputil.VpnIp
 }
 
+// For synchronization, treat the Relay struct as immutable. To edit the Relay
+// struct, make a copy of an existing value, edit the fileds in the copy, and
+// then store a pointer to the copy in RelayWrapper.
+type RelayWrapper struct {
```
nbrownus, line 56: It looks like every time we try and get this from a map in `RelayState` we also have an `ok` check before calling `Load()`.

```
type RelayState struct {
	sync.RWMutex

	relays        map[iputil.VpnIp]struct{}
	relayForByIp  map[iputil.VpnIp]atomic.Pointer[Relay]
	relayForByIdx map[uint32]atomic.Pointer[Relay]
}

func (rs *RelayState) RemoveRelay(localIdx uint32) (iputil.VpnIp, bool) {
	rs.Lock()
	defer rs.Unlock()
	relay, ok := rs.relayForByIdx[localIdx]
	if !ok {
		return iputil.VpnIp(0), false
	}
	r := relay.Load()
	delete(rs.relayForByIdx, localIdx)
	delete(rs.relayForByIp, r.PeerIp)
	return r.PeerIp, true
}
```

Should behave the same without the additional indirection unless I have missed something.

outside.go
```diff
@@ -89,12 +89,8 @@ func (f *Interface) readOutsidePackets(addr *udp.Addr, via interface{}, out []by
...
 			if !ok {
 				// The only way this happens is if hostmap has an index to the correct HostInfo, but the HostInfo is missing
-				// its internal mapping. This shouldn't happen!
-				hostinfo.logger(f.l).WithField("hostinfo", hostinfo.vpnIp).WithField("remoteIndex", h.RemoteIndex).Errorf("HostInfo missing remote index")
-				// Delete my local index from the hostmap
-				f.hostMap.DeleteRelayIdx(h.RemoteIndex)
-				// When the peer doesn't receive any return traffic, its connection_manager will eventually clean up
-				// the broken relay when it cleans up the associated HostInfo object.
+				// its internal mapping. This should never happen.
+				hostinfo.logger(f.l).WithFields(logrus.Fields{"hostinfo": hostinfo.vpnIp, "remoteIndex": h.RemoteIndex}).Error("HostInfo missing remote relay index")
```
nbrownus, line 93: `hostInfo` -> `vpnIp` to match logging convention

hostmap.go
brad-defined: My motivation there was to be able to edit the Relay in-place without having to mess around with the maps. (This happens during the CompleteRelayByIp() and CompleteRelayByIdx() calls in this PR.) Since those methods have been moved to be RelayState methods in this PR, they use locks held on the RelayState struct. So, it's probably simpler/easier to maintain both maps with the lock held.

Using the atomics would require both maps use the _same_ atomic, so it'd have to be a `map[thing]*atomic.Pointer[Relay]`, but all that really gains us at this point is the ability to use a RLock in the CompleteRelayByIp|Idx() calls...which I apparently didn't do anyway, and maybe doesn't gain us a whole lot.

outside.go
brad-defined: done!

outside.go
```diff
@@ -89,12 +89,8 @@ func (f *Interface) readOutsidePackets(addr *udp.Addr, via interface{}, out []by
...
 			if !ok {
 				// The only way this happens is if hostmap has an index to the correct HostInfo, but the HostInfo is missing
-				// its internal mapping. This shouldn't happen!
-				hostinfo.logger(f.l).WithField("hostinfo", hostinfo.vpnIp).WithField("remoteIndex", h.RemoteIndex).Errorf("HostInfo missing remote index")
-				// Delete my local index from the hostmap
-				f.hostMap.DeleteRelayIdx(h.RemoteIndex)
-				// When the peer doesn't receive any return traffic, its connection_manager will eventually clean up
-				// the broken relay when it cleans up the associated HostInfo object.
+				// its internal mapping. This should never happen.
+				hostinfo.logger(f.l).WithFields(logrus.Fields{"hostInfo": hostinfo.vpnIp, "remoteIndex": h.RemoteIndex}).Error("HostInfo missing remote relay index")
```
nbrownus, line 93: Handful of `hostInfo` -> `vpnIp` log field changes here to match convention.

relay_manager.go
```diff
@@ -155,44 +152,63 @@ func (rm *relayManager) handleCreateRelayResponse(h *HostInfo, f *Interface, m *
...
-			WithError(err).Error("relayManager Failed to marhsal Control CreateRelayResponse message to create relay")
+			WithError(err).Error("relayManager Failed to marshal Control CreateRelayResponse message to create relay")
 	} else {
 		f.SendMessageToVpnIp(header.Control, 0, peerHostInfo.vpnIp, msg, make([]byte, 12), make([]byte, mtu))
+		rm.l.WithFields(logrus.Fields{
+			"relayFrom":    iputil.VpnIp(resp.RelayFromIp),
+			"relayTarget":  iputil.VpnIp(resp.RelayToIp),
+			"initiatorIdx": resp.InitiatorRelayIndex,
+			"responderIdx": resp.ResponderRelayIndex,
+			"hostInfo":     peerHostInfo.vpnIp}).
```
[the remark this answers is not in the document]
brad-defined: oh, I thought you wanted me to change the capitalization of it.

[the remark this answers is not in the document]
brad-defined: fixed again

relay_manager.go
```diff
@@ -88,17 +88,14 @@ func AddRelay(l *logrus.Logger, relayHostInfo *HostInfo, hm *HostMap, vpnIp iput
 
 // EstablishRelay updates a Requested Relay to become an Established Relay, which can pass traffic.
 func (rm *relayManager) EstablishRelay(relayHostInfo *HostInfo, m *NebulaControl) (*Relay, error) {
-	relay, ok := relayHostInfo.relayState.QueryRelayForByIdx(m.InitiatorRelayIndex)
+	relay, ok := relayHostInfo.relayState.CompleteRelayByIdx(m.InitiatorRelayIndex, m.ResponderRelayIndex)
 	if !ok {
 		rm.l.WithFields(logrus.Fields{"relayHostInfo": relayHostInfo.vpnIp,
```
nbrownus, line 93: I think `relayHostInfo` -> `relayVpnIp`.

Probably would be good to do a separate pass to make sure all logs are using the same field names.

nbrownus (approved):

PR #828, opened 2023-03-17
Remove dead code and re-order transit from pending to main hostmap on stage 2

nbrownus:
@brad-defined and I found multiple hostinfos for a given vpn ip were unlinked inappropriately by the handshake manager.

As part of the spelunking to discover what was unlinking the hostinfos I tripped over some dead code.

Longer term I think we should pull pending hostmap entirely into handshake manager since their needs are quite different anymore.

### DIFF
--- a/handshake_manager.go
+++ b/handshake_manager.go
@@ -384,8 +384,9 @@ func (c *HandshakeManager) Complete(hostinfo *HostInfo, f *Interface) *HostInfo 
 	}
 
 	existingHostInfo := c.mainHostMap.Hosts[hostinfo.vpnIp]
-	c.mainHostMap.unlockedAddHostInfo(hostinfo, f)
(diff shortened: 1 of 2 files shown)


handshake_manager.go
```diff
@@ -384,8 +384,8 @@ func (c *HandshakeManager) Complete(hostinfo *HostInfo, f *Interface) *HostInfo
 	}
 
 	existingHostInfo := c.mainHostMap.Hosts[hostinfo.vpnIp]
-	c.mainHostMap.unlockedAddHostInfo(hostinfo, f)
 	c.pendingHostMap.unlockedDeleteHostInfo(hostinfo)
```
brad-defined, line 387: This landmine deserves a comment!

brad-defined (approved):

brad-defined (approved):

PR #830, opened 2023-03-25
Listen ports for nodes behind NAT

tahyonline:
If nodes on the same local network behind NAT are set to the same listen port, the router might not be able to correctly route the response packets from the lighthouse and other nodes outside the local network back to the correct node.

As a result, nodes on the local network will suddenly stop being able to access the Nebula network, if more than one node generates traffic at the same time. The issue is quite "mysterious", as each node works individually and there is no error indication other than timeouts.

Recommend adding a note to the sample configuration to help prevent others spending hours on figuring out why a configuration copied from a working node would suddenly not work on another.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -106,6 +106,8 @@ lighthouse:
 
 # Port Nebula will be listening on. The default here is 4242. For a lighthouse node, the port should be defined,
 # however using port 0 will dynamically assign a port and is recommended for roaming nodes.
+# If you run several nodes on the same local network behind a NAT router, e.g., residential DSL or mobile broadband,
+# you must set different ports for each node or use port 0 for dynamic assignment.
 listen:
   # To listen on both any ipv4 and ipv6 use "[::]"
   host: 0.0.0.0


tahyonline:
Internal router behind a carrier provided router behind CG-NAT... which should be (or become) pretty much the norm for all except corporate networks, where edge routers can still get real IP addresses.

In my case, the internal router will route to Starlink or mobile broadband as backup if Starlink is down, but the primary link was up.

brad-defined:
@tahyonline thanks! what brand and model of internal router are you using? And, what NAT configuration settings are you using on that router?

tahyonline:
@brad-defined
It's a Pepwave MAX On-The-Go MAX-OTG-U4, doing NAT and DHCP.

There are three internal VLANs: main, guest and IoT. All Nebula nodes are on the main net. To my knowledge, none of the devices on the IoT network use the same UDP port as Nebula and the port is not allowed anyway on that VLAN. No stations on the guest network.

There are the two WAN connections, both DHCP and NAT enabled, and there are no other NAT-related settings.

Were there any other settings that might be pertinent?

johnmaguire:
Thanks for the PR, we don't believe this is a common issue with NATs in general. Thanks for mentioning the issue in case it comes up again.

PR #835, opened 2023-03-30
Use connection manager to drive NAT maintenance

nbrownus:
Similar to #746

This refactors `connectionManager` to drive hole punch maintenance instead of `HostMap`.

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -2,7 +2,6 @@ package nebula
 
 import (
 	"context"
-	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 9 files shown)


connection_manager.go
```diff
@@ -322,8 +246,24 @@ func (n *connectionManager) handleInvalidCertificate(now time.Time, hostinfo *Ho
...
-
-	n.ClearLocalIndex(hostinfo.localIndexId)
-	n.ClearPendingDeletion(hostinfo.localIndexId)
+	delete(n.pendingDeletion, hostinfo.localIndexId)
 	return true
 }
+
+func (n *connectionManager) sendPunch(hostinfo *HostInfo) {
+	if hostinfo.remote == nil || !n.punchy.GetPunch() {
+		// Don't punch to relayed hosts
```
brad-defined, line 255: should GetTargetEverything spam the network with 1-byte packets for every advertised remote address? If so, then it should probably spam all those addrs even for hosts currently going through a relay.

Especially if the purpose of it is to prime the network (host firewalls, etc.) for a tunnel migration.

brad-defined (dismissed):

PR #838, opened 2023-03-31
Rehandshaking

nbrownus:
Nebula can soft-reload its host certificate on a HUP signal. When it reloads a certificate, any new peer tunnels created will use the updated certificate. However, any previously established tunnels will not be updated with this new certificate information.

With this change, Nebula will detect when an older certificate is in use on an existing tunnel, and rebuild the tunnel with the newly updated host certificate. Once hosts have established a new tunnel, packet traversal is migrated to use the new tunnel, and the old tunnel is torn down when its lack of use is detected.

This PR also includes migrating relay information from the previously established tunnel to the new tunnel. A relay established over an older tunnel will be migrated over to the new tunnel, ensuring that the old tunnel can be torn down from lack of use.

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -130,8 +130,15 @@ func (n *connectionManager) getAndResetTrafficCheck(localIndex uint32) (bool, bo
 }
 
 func (n *connectionManager) AddTrafficWatch(localIndex uint32) {
-	n.Out(localIndex)
+	// Use a write lock directly because it should be incredibly rare that we are ever already tracking this index
+	n.outLock.Lock()
+	if _, ok := n.out[localIndex]; ok {
+		n.inLock.Unlock()
+		return
+	}
+	n.out[localIndex] = struct{}{}
 	n.trafficTimer.Add(localIndex, n.checkInterval)
+	n.outLock.Unlock()
 }
 
 func (n *connectionManager) Start(ctx context.Context) {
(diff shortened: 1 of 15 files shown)


jasikpark:
how is age of the cert determined?

connection_manager.go
```diff
@@ -99,8 +130,15 @@ func (n *connectionManager) getAndResetTrafficCheck(localIndex uint32) (bool, bo
 }
 
 func (n *connectionManager) AddTrafficWatch(localIndex uint32) {
-	n.Out(localIndex)
+	// Use a write lock directly because it should be incredibly rare that we are ever already tracking this index
+	n.outLock.Lock()
+	if _, ok := n.out[localIndex]; ok {
+		n.inLock.Unlock()
```
wadey, line 136:
```suggestion
		n.outLock.Unlock()
```

relay_manager.go
```diff
@@ -278,7 +280,7 @@ func (rm *relayManager) handleCreateRelayRequest(h *HostInfo, f *Interface, m *N
 				logMsg.
 					WithError(err).Error("relayManager Failed to marshal Control message to create relay")
 			} else {
-				f.SendMessageToVpnIp(header.Control, 0, target, msg, make([]byte, 12), make([]byte, mtu))
+				f.SendMessageToHostInfo(header.Control, 0, peer, msg, make([]byte, 12), make([]byte, mtu))
```
wadey, line 283: Was this a bug that it was `target` before?

wadey (approved):
Overall LGTM and working live

connection_manager.go
```diff
@@ -136,18 +174,130 @@ func (n *connectionManager) Run(ctx context.Context) {
 }
 
 func (n *connectionManager) doTrafficCheck(localIndex uint32, p, nb, out []byte, now time.Time) {
-	hostinfo, err := n.hostMap.QueryIndex(localIndex)
-	if err != nil {
+	decision, hostinfo, primary := n.makeTrafficDecision(localIndex, p, nb, out, now)
+
+	switch decision {
+	case deleteTunnel:
+		n.hostMap.DeleteHostInfo(hostinfo)
```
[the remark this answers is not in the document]
nbrownus: I wouldn't want to use that one specifically but I could see a good argument for dropping it from the lighthouse cache. The only time we hit this case today is when we aren't receiving traffic even after sending a test packet and waiting some time.

I can spin another PR pretty quick to take care of that after merging this.

PR #839, opened 2023-04-03
Add ability to respect the system route table for unsafe route on linux

### LINKED ISSUE #274: Feature Request: Add custom routes outside the nebula's unsafe_routes (for K8S Flannel)
I run the docker container on multiple nebula nodes, Containers on each node have different networks, such as 172.16.12.0/24, 172.16.21.0/24;

I try to add custom routes in Linux route tables like this:
route add -net 172.16.12.0 netmask 255.255.255.0 gw 192.168.10.12
route add -net 172.16.21.0 netmask 255.255.255.0 gw 192.168.10.21

But it doesn't work, I can't ping container's ip on other Nebula nodes;
I have to specify the routes for each container network in nebula unsafe_routes, like this;
 unsafe_routes:
    - route: 172.16.12.0/24
      via: 192.168.10.12
    - route: 172.16.21.0/24
      via: 192.168.10.21
	  
Use unsafe_routes, can ping container's ip on other Nebula nodes.

When we use K8S Flannel to manage container networks, the network is dynamic allocation by K8S Flannel, not fixed.
Flannel can manage the routing of container network by itself, and add the corresponding route to the host.

So the problem is that Flannel can not work with Nebula, and other app like Flannel which need route also can not work with Nebula；

It's seem Nebula will queryUnsafeRoute to determine which ip can routable, so if add custom routes outside the nebula's unsafe_routes, Nebula will drop it;

I know many P2P VPN tools can support to add custom routes in host route table, such as ZeroTier, Tinc VPN and WireGuard,  can Nebula also support this ?

nbrownus:
Adds a new configuration option that defaults to false. If set to true then `Tun` on linux will use a netlink socket to listen to route table changes on the system.

Example:
```
sudo ip route add 1.1.1.1/32 via 192.168.10.1 dev nebula1
```

`ping 1.1.1.1` would be tunneled through nebula to the host at 192.168.10., which is an unsafe router that is allowed to route the traffic, at which point it exits nebula destined for 1.1.1.1.

This change is similar to #831.

Closes #274

### DIFF
--- a/cidr/tree4.go
+++ b/cidr/tree4.go
@@ -60,6 +60,15 @@ func (tree *Tree4) AddCIDR(cidr *net.IPNet, val interface{}) {
 
 	// We already have this range so update the value
 	if next != nil {
+		addCIDR := cidr.String()
+		for i, v := range tree.list {
+			if addCIDR == v.CIDR.String() {
+				tree.list = append(tree.list[:i], tree.list[i+1:]...)
+				break
+			}
+		}
+
+		tree.list = append(tree.list, entry{CIDR: cidr, Value: &val})
 		node.value = val
 		return
 	}
--- a/cidr/tree4.go
+++ b/cidr/tree4.go
@@ -13,8 +13,14 @@ type Node struct {
 	value  interface{}
 }
 
+type entry struct {
+	CIDR  *net.IPNet
+	Value *interface{}
+}
+
 type Tree4 struct {
 	root *Node
+	list []entry
 }
 
 const (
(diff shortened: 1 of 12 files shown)


cidr/tree4.go
```diff
@@ -74,9 +81,10 @@ func (tree *Tree4) AddCIDR(cidr *net.IPNet, val interface{}) {
 
 	// Final node marks our cidr, set the value
 	node.value = val
+	tree.list = append(tree.list, entry{CIDR: cidr, Value: &val})
```
wadey, line 84: Do we need to update the `tree.list` entry earlier for the "We already have this range so update the value" short circuit?

nbrownus: Nice catch and yes.

PR #840, opened 2023-04-04
Use fewer test packets

nbrownus:
Old connection manager would only track connections with outbound traffic, if there was no inbound then a test packet would be sent to ensure the tunnel was still viable.

This change matches the previous behavior as well as allow for non primary lighthouse tunnels to be torn down eventually


wadey (approved):
This looks correct to me

brad-defined (approved):

PR #843, opened 2023-04-05
Pull hostmap and pending hostmap apart, remove unused functions

nbrownus:
The main `HostMap` has concerns that the pending host map does not and has led to bugs in recent changes.

This change puts `HandshakeManager` in charge of its own dedicated maps that share some similarity with the main `HostMap` but lacks all the extra cruft. This also removes some unused code from the main `HostMap` as a result of the split.

My main concern is: `HandshakeManager.AddVpnIp` will go for a write lock immediately, `Interface.getOrHandshake` won't attempt check the pending hostmap before making a call to this function. This avoids a read lock to write lock trading race that lead to the overcomplication of `Interface.getOrHandshake` long ago. The concern is when restarting nebula on a busy host with many tunnels, there will be a period of strong write lock contention until the tunnels form or the services using the overlay slow down transmitting until replies come in.

If we believe this is an issue I can go back to behavior similar to the old way that still sets us up to simplify `Interface.getOrHandshake` in the future with a read lock check and a write lock check before proceeding to add.

### DIFF
--- a/outside.go
+++ b/outside.go
@@ -64,9 +64,9 @@ func (f *Interface) readOutsidePackets(addr *udp.Addr, via *ViaSender, out []byt
 	var hostinfo *HostInfo
 	// verify if we've seen this index before, otherwise respond to the handshake initiation
 	if h.Type == header.Message && h.Subtype == header.MessageRelay {
-		hostinfo, _ = f.hostMap.QueryRelayIndex(h.RemoteIndex)
+		hostinfo = f.hostMap.QueryRelayIndex(h.RemoteIndex)
 	} else {
-		hostinfo, _ = f.hostMap.QueryIndex(h.RemoteIndex)
+		hostinfo = f.hostMap.QueryIndex(h.RemoteIndex)
 	}
 
 	var ci *ConnectionState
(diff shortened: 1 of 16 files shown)


maggie44:
I am looking to poll the hostmap every 4 seconds to be able to keep an up to date list of connected devices for use on a dashboard. Would that be blocked by these locks or create a lock itself that would prevent devices joining?

nbrownus:
That would require a read lock and could slow down a few interactions but its probably fine. `Control.ListHostmapHosts` or `Control.ListHostmapIndexes` would be your best bet at accessing that information as well.

maggie44:
I've been using `hostMap := Ctrl.ListHostmap(false)` exposed by `*nebula.Control`, but looks like it is doing the same thing (not accounting for anything in this PR mind). 

It could be 100,000s of nodes, but presumably it is just a lookup of data already gathered, so thinking it shouldn't be a noticeable delay or lock 🤞

outside.go
```diff
@@ -449,12 +449,9 @@ func (f *Interface) handleRecvError(addr *udp.Addr, h *header.H) {
 			Debug("Recv error received")
 	}
 
-	// First, clean up in the pending hostmap
-	f.handshakeManager.pendingHostMap.DeleteReverseIndex(h.RemoteIndex)
```
wadey, line 453: we lose the delete here, is that okay?

wadey (approved):
Looks good except one comment

PR #845, opened 2023-04-09
Set NXDOMAIN if there's no Answer to return

### LINKED ISSUE #841: 🐛 BUG: Lighthouse DNS should return NXDOMAIN for names it doesn't know about

dnclient 0.1.9

Linux

### Describe the Bug

`dig @100.88.0.1 gmail.com`  should result in NXDOMAIN but with DNClient 0.1.9, it returns NOERROR with an empty list. The problem is the OS stub resolver `systemd-resolved` interprets that as a valid name which happened not to have A record.

When I have multiple DNS servers for `systemd-resolved` to query, sometimes the lighthouse replies the quickest among them, but without NXDOMAIN. Then `systemd-resolved` believes `gmail.com` doesn't have any A records, at which point `getaddrinfo()` returns an error without waiting for my home router's DNS or 8.8.8.8/1.1.1.1. If the lighthouse returned NXDOMAIN, then `systemd-resovled` would have waited for my home router to resolve `gmail.com`.

I'm using Network Manager whose CLI is more user friendly than `systemd-networkd` especially for WiFi connectivity. Maybe people using `systemd-networkd` can specifically say "hey whenever you [networkd] *see* an interface named `defined1`, use `100.88.0.1` as a DNS server associated with it" but as far as I know, NetworkManager isn't capable enough to say anything to interfaces it isn't managing. So I'm stuck with having the lighthouse as yet another DNS server for the system, to which queries for `gmail.com` and so on can be sent.

### Logs from affected hosts

_No response_

_No response_

nodakai:
This change has not been tested at all, partly because I only have an environment based on DNClient rather than OSS Nebula.
(At least `make bin` ran successfully on GitHub Codespaces)

Fix #841

### DIFF
--- a/dns_server.go
+++ b/dns_server.go
@@ -96,6 +96,10 @@ func parseQuery(l *logrus.Logger, m *dns.Msg, w dns.ResponseWriter) {
 			}
 		}
 	}
+
+	if len(m.Answer) == 0 {
+		m.Rcode = dns.RcodeNameError
+	}
 }
 
 func handleDnsRequest(l *logrus.Logger, w dns.ResponseWriter, r *dns.Msg) {


nbrownus:
@nodakai mind signing the newer CLA?

nbrownus (approved):

PR #851, opened 2023-04-20
Have lighthouses ack updates to reduce test packet traffic

nbrownus:
This change has the lighthouse send an `ack` packet on the `lighthouse` message type after receiving an update. The goal is to get some return traffic on the line so the client doesn't have to always send test packets to determine if the lighthouse is still alive.


examples/config.yml
```diff
@@ -42,6 +42,13 @@ lighthouse:
   hosts:
     - "192.168.100.1"
 
+  # Toggle sending test packets to lighthouse hosts to assert the tunnels are still viable.
+  # Disabling sending test packets to lighthouses can reduce bookkeeping traffic
+  # and load on lighthouses with many tunnels.
+  # Setting to false may lead to connectivity issues if a lighthouse host is behind a NAT or firewall that
+  # tracks outbound packets from the lighthouse to the clients.
+  #test: true
```
jasikpark, line 50: is `lighthouse.test` a specific enough key? 🤷 w/o this comment i probably wouldn't know what it's testing

jasikpark (commented):
sounds good to me 😁

examples/config.yml
jasikpark: `lighthouse.send_test_packets = true`?

jasikpark (approved):
lgtm

wadey (approved):

PR #859, opened 2023-05-04
Default `pki.disconnect_invalid` to true and make it reloadable

nbrownus:
With #838 merged there is a path to refresh certificates and rekey tunnels without having to restart the nebula process. It would be a good practice to flip this to `true`.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -40,7 +40,6 @@ type InterfaceConfig struct {
 	routines                int
 	MessageMetrics          *MessageMetrics
 	version                 string
-	disconnectInvalid       bool
 	relayManager            *relayManager
 	punchy                  *Punchy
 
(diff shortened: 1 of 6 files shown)


jasikpark (commented):
would be nice! i don't understand the code, but it looks right?

interface.go
```diff
@@ -170,7 +169,6 @@ func NewInterface(ctx context.Context, c *InterfaceConfig) (*Interface, error) {
 		writers:            make([]*udp.Conn, c.routines),
 		readers:            make([]io.ReadWriteCloser, c.routines),
 		caPool:             c.caPool,
-		disconnectInvalid:  c.disconnectInvalid,
```
JohnMaguire, line 173: Do we need `ifce.disconnectInvalid.Store(c.disconnectInvalid)` below?

JohnMaguire: It looks like no, because we call `ifce.reloadDisconnectInvalid(c)` after calling `NewInterface` from main.go.

This is contrast to `ifce.certState` which is handled internally in this function.

I think this approach is a little awkward because the caller may not be aware they need to initialize this value.

johnmaguire (approved):

PR #886, opened 2023-05-31
Fix reconfig freeze attempting to send to an unbuffered, unread channel

brad-defined:
Only create stop channel when a DNS background goroutine is created. Only send the stop signal when the channel exists.
Also create the channel with a buffer to size 1 so that the stop message can be immediately sent even if the goroutine is busy doing DNS lookups.

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -265,14 +265,13 @@ func (lh *LightHouse) reload(c *config.C, initial bool) error {
 		// Clean up. Entries still in the static_host_map will be re-built.
 		// Entries no longer present must have their (possible) background DNS goroutines stopped.
 		if existingStaticList := lh.staticList.Load(); existingStaticList != nil {
-			lh.Lock()
+			lh.RLock()
 			for staticVpnIp := range *existingStaticList {
-				am := lh.unlockedGetRemoteList(staticVpnIp)
-				if am != nil {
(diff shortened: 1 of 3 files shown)


lighthouse.go
```diff
@@ -262,6 +262,19 @@ func (lh *LightHouse) reload(c *config.C, initial bool) error {
 
 	//NOTE: many things will get much simpler when we combine static_host_map and lighthouse.hosts in config
 	if initial || c.HasChanged("static_host_map") || c.HasChanged("static_map.cadence") || c.HasChanged("static_map.network") || c.HasChanged("static_map.lookup_timeout") {
+		// Clean up. Entries still in the static_host_map will be re-built.
+		// Entries no longer present must have their (possible) background DNS goroutines stopped.
+		if existingStaticList := lh.staticList.Load(); existingStaticList != nil {
+			lh.Lock()
+			for staticVpnIp := range *existingStaticList {
+				am := lh.unlockedGetRemoteList(staticVpnIp)
```
nbrownus, line 270: I'm not sure how it would come up in practice but `unlockedGetRemoteList` has a side effect of adding an entry if it doesn't exist, this is why it needs a write lock.

We could check `lh.addrMap[staticVpnIp]` directly and use a read lock instead

nbrownus (approved):

wadey (approved):

PR #903, opened 2023-06-14
Add support for naming FreeBSD tun devices

### LINKED ISSUE #309: FreeBSD requires tunX
It seems that FreeBSD requires tunX, and trying neb1 causes this error message:-

Nebula Error Message:
ERRO[0000] Failed to get a tun/tap device                error="tun.dev must match `tun[0-9]+`"

Having spent some hours configuring up Nebula within OPNsense, alas I'm defeated at the last step as the OPNsense webUI prevents tunX from appearing.
I've tried renaming an existing interface, ifconfig neb2 name tun2, but that doesn't solve issue.

Is there a work-around?
Can the restriction be removed, because then there could be a OPNsense Plugin!!
Thanks

JohnMaguire:
Fixes #309.

If no name is specified, the system chooses one. If a name _is_ specified, and the tun exists already, we'll try to use it. If it doesn't exist, we'll rename the one the system gives us.

This could've been done with ifconfig, but I figured this was a good start towards syscalls for FreeBSD.

### DIFF
--- a/overlay/tun_freebsd.go
+++ b/overlay/tun_freebsd.go
@@ -67,9 +67,7 @@ func (t *tun) Close() error {
 		ifreq := ifreqDestroy{Name: t.deviceBytes()}
 
 		// Destroy the interface
-		fmt.Println("Destroying interface...")
 		err = ioctl(uintptr(s), syscall.SIOCIFDESTROY, uintptr(unsafe.Pointer(&ifreq)))
-		fmt.Println("destroyed!")
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


overlay/tun_notwin.go
```diff
@@ -0,0 +1,14 @@
+//go:build !windows
+// +build !windows
```
wadey, line 2: Is there any harm in just putting this function in `overlay/tun.go`? I dont think we need a notwin file for it, it will still compile on windows just not be used.

wadey: oh haha

https://github.com/slackhq/nebula/pull/903/commits/5159263448ee51bf26ceb05f4b737586fab833d5

overlay/tun_freebsd.go
```diff
@@ -95,15 +93,18 @@ func newTun(l *logrus.Logger, deviceName string, cidr *net.IPNet, defaultMTU int
 		return nil, err
 	}
 
-	fileFd := file.Fd()
-	err = syscall.SetNonblock(int(fileFd), true)
```
JohnMaguire: This call is necessary due to a side-effect of `file.Fd()`. By default, the FIle returned by OpenFile is in non-blocking mode. If we leave it in blocking mode, Nebula will hang when we try to destroy the interface in the Close method (part of cleanly shutting down Nebula.)

Based on [discussion with Go maintainers](https://go-review.googlesource.com/c/go/+/449796), it seems that `file.Fd()` should only be called if you're done using the File object. In this case we are not, so it's better to use SyscallConn.

We do also call `syscall.SetNonblock` in `tun_darwin.go`. I suspect this is because the fd returned by `unix.Socket()` is not non-blocking by default, and we want to place it in non-blocking mode before calling NewFile:

> NewFile returns a new File with the given file descriptor and name. The returned value will be nil if fd is not a valid file descriptor. On Unix systems, if the file descriptor is in non-blocking mode, NewFile will attempt to return a pollable File (one for which the SetDeadline methods work).

The only surprising bit to me is that in `tun_linux.go` we call `file.Fd()` in both newTun and newTunFromFd. This means that the fd is now in blocking mode, and we continue to use the File object in Nebula... this seems wrong but I haven't heard anyone report hangs on Linux. It's possible that the `(t *tun) Write(b []byte) (int, error)` function could be re-written to use SyscallConn, negating the need to call `file.Fd` in newTun, but I'm not sure what implications this would have.

nbrownus (approved):
Looks fine to me, we should rebind iOS and get a test build out after this merges.

overlay/tun_freebsd.go
noseshimself: Forgive my naive question, but why don't you ask the file descriptor?

res = fcntl (FileFd, F_GETFL)

should return all you need to know; all you have to check is O_NONBLOCK.

PR #913, opened 2023-07-06
Add http pprof profiling endpoint 

timvaillancourt:
This PR adds the ability to gather [`pprof` profiles](https://pkg.go.dev/net/http/pprof) from Nebula in order to understand/improve performance and understand the high CPU usage we see in Slack's production Vitess platform

This change requires making the http server opened by `stats.go` to become more generic than "just stats", so some logic has been shuffled into `http.go` and a new, generic config variable `http.listen` has been added to define this port

For backwards compatibility, `http.listen` will fallback to using the value for `stats.listen` if `http.listen` is undef. This fallback should be removed in future releases


wadey:
Thanks! We will consider this, but I'm not sure if you are aware there is already a method to get pprof traces using the SSH debug interface.

- https://nebula.defined.net/docs/config/sshd/

Once you connect you can run `help` to see the methods, but the ones you might care about are `start-cpu-profile`, `stop-cpu-profile`.

Defined here in the code: https://github.com/slackhq/nebula/blob/v1.7.2/ssh.go#L228

PR #929, opened 2023-07-24
Attempt to notify systemd of service readiness on linux

### LINKED ISSUE #921: Sample systemd unit "Before" insufficient

1.7.2

Linux

### Describe the Bug

In the example systemd service unit, `Before` is used to delay starting of sshd until nebula is started. However, this is insufficient because the implied unit type is `simple`. This does not wait until the tun interface is actually started and has an IP address, so dependent services like ssh that are configured to bind to the nebula tun device's local IP address can still fail to bind.

Even adding a `Requires=sys-devices-virtual-net-nebula1.device` (or similar) dependency to `sshd.service` is insufficient as this only waits for the device to exist -- there is a window after the device is created and before it is assigned an IP address where binding can still fail.

I believe a proper fix would be to support systemd's `Type=forking` mode, where nebula would create and configure the tun device before forking into the background.  This would fix the problem by making systemd wait to start the `Before` units until the tun device is actually up and configured.

(Note that `sshd.service` typically has `Restart=on-failure` so this problem would correct itself for sshd specifically, but other services without such a directive will simply enter the failed state and the situation would require manual intervention to correct.)

### Logs from affected hosts

_No response_

_No response_

nbrownus:
This will send `READY=1` to the unix domain socket described by the `NOTIFY_SOCKET` environment variable, if it was provided, after we have called `Control.Activate()`.

Should resolve timing issues with services that depend on nebulas udp listener(s) and tun device being up.

Closes #921

### DIFF
--- a/cmd/nebula/main.go
+++ b/cmd/nebula/main.go
@@ -65,6 +65,7 @@ func main() {
 
 	if !*configTest {
 		ctrl.Start()
+		notifyReady(l)
 		ctrl.ShutdownBlock()
 	}
 
--- a/cmd/nebula/notify_linux.go
+++ b/cmd/nebula/notify_linux.go
@@ -22,19 +22,19 @@ func notifyReady(l *logrus.Logger) {
 
 	conn, err := net.DialTimeout("unixgram", sockName, time.Second)
 	if err != nil {
-		l.WithError(err).Debugln("failed to connect to systemd notification socket")
+		l.WithError(err).Error("failed to connect to systemd notification socket")
 		return
 	}
 	defer conn.Close()
 
 	err = conn.SetWriteDeadline(time.Now().Add(time.Second))
 	if err != nil {
-		l.WithError(err).Debugln("failed to set the write deadline for the systemd notification socket")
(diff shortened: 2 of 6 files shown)


JohnMaguire (dismissed):
This looks correct to me.

wadey (approved):

krabiswabbie:
Hello,

I wanted to report an ongoing issue with the example systemd service description provided in `examples/service_scripts`. Despite the feature being marked as resolved, I've found that Nebula fails to notify `systemd` of its successful startup when the `systemd` service is configured with `Type=notify`. As a result, the Nebula service continuously restarts due to a timeout.

This behavior is observed in the default installation of `v1.7.2` on both `amd64` and `arm64` architectures. I'm aware that switching the `systemd` service type to `simple` is a workaround, but thought it would be beneficial for the team to know that the issue persists, especially considering it might affect new users trying to set up Nebula using the example provided.

Thank you for your dedication to this project.

JohnMaguire:
Hi @krabiswabbie - this PR has not made it into a release yet. You can try running one of the nightly builds now if you like: https://github.com/NebulaOSS/nebula-nightly/releases

PR #935, opened 2023-07-25
Send the lh update worker into its own routine instead of taking over the reload routine

nbrownus:
@JohnMaguire came across this issue today. 

If the `lighthouse.interval` config is changed and nebula reloads it then the reload routine is taken over. This will cause nebula to be unable to shutdown cleanly and block future reloads from occurring.

```
goroutine 11 [select]:
runtime.gopark(0xc0001f56b0?, 0x2?, 0x0?, 0x0?, 0xc0001f567c?)
        runtime/proc.go:381 +0xd6 fp=0xc0001f54f8 sp=0xc0001f54d8 pc=0x43b9b6
runtime.selectgo(0xc0001f56b0, 0xc0001f5678, 0xc000191200?, 0x0, 0x0?, 0x1)
        runtime/select.go:327 +0x7be fp=0xc0001f5638 sp=0xc0001f54f8 pc=0x44bc5e
github.com/slackhq/nebula.(*LightHouse).LhUpdateWorker(0xc000184a50, {0xd3e608?, 0xc000036fa0}, {0xd3e8e0?, 0xc000191200})
        github.com/slackhq/nebula@v1.7.2/lighthouse.go:774 +0x20c fp=0xc0001f56e8 sp=0xc0001f5638 pc=0xa70dac
github.com/slackhq/nebula.(*LightHouse).reload(0xc000184a50, 0x1171f00?, 0x0)
        github.com/slackhq/nebula@v1.7.2/lighthouse.go:220 +0x775 fp=0xc0001f5b18 sp=0xc0001f56e8 pc=0xa6c315
github.com/slackhq/nebula.NewLightHouseFromConfig.func1(0xc000034cc0?)
        github.com/slackhq/nebula@v1.7.2/lighthouse.go:134 +0x32 fp=0xc0001f5ba0 sp=0xc0001f5b18 pc=0xa6b9d2
github.com/slackhq/nebula/config.(*C).ReloadConfig(0xc000034cc0)
        github.com/slackhq/nebula@v1.7.2/config/config.go:158 +0x3a3 fp=0xc0001f5e18 sp=0xc0001f5ba0 pc=0x896383
```

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -64,11 +64,10 @@ type LightHouse struct {
 	staticList  atomic.Pointer[map[iputil.VpnIp]struct{}]
 	lighthouses atomic.Pointer[map[iputil.VpnIp]struct{}]
 
-	interval        atomic.Int64
-	updateCancel    context.CancelFunc
-	updateParentCtx context.Context
-	updateUdp       EncWriter
-	nebulaPort      uint32 // 32 bits because protobuf does not have a uint16
+	interval     atomic.Int64
+	updateCancel context.CancelFunc
+	ifce         EncWriter
+	nebulaPort   uint32 // 32 bits because protobuf does not have a uint16
 
 	advertiseAddrs atomic.Pointer[[]netIpAndPort]
 
@@ -217,7 +216,7 @@ func (lh *LightHouse) reload(c *config.C, initial bool) error {
 				lh.updateCancel()
 			}
 
-			lh.LhUpdateWorker(lh.updateParentCtx, lh.updateUdp)
+			lh.StartUpdateWorker()
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


lighthouse.go
```diff
@@ -764,7 +762,7 @@ func (lh *LightHouse) LhUpdateWorker(ctx context.Context, f EncWriter) {
 	}
 
 	clockSource := time.NewTicker(time.Second * time.Duration(interval))
-	updateCtx, cancel := context.WithCancel(ctx)
+	updateCtx, cancel := context.WithCancel(lh.ctx)
 	lh.updateCancel = cancel
```
brad-defined, line 766: This lh variable update must be synchronized. (I think that's why the test failed - the lh.updateCancel variable was read during the test's subsequent update in the main test goroutine before this line wrote the new variable in the spawned goroutine.)

I guess this could be solved by making this portion of the function synchronous, and spawning the goroutine within this function a few lines down for the actual worker loop.

brad-defined (approved):

lighthouse.go
```diff
@@ -754,33 +753,34 @@ func NewUDPAddrFromLH6(ipp *Ip6AndPort) *udp.Addr {
...
+func (lh *LightHouse) StartUpdateWorker() {
 	interval := lh.GetUpdateInterval()
 	if lh.amLighthouse || interval == 0 {
 		return
 	}
 
 	clockSource := time.NewTicker(time.Second * time.Duration(interval))
-	updateCtx, cancel := context.WithCancel(ctx)
+	updateCtx, cancel := context.WithCancel(lh.ctx)
+	// We seem to have a race right here, the first caller never gets canceled because the updateCancel is set by the 2nd?
```
wadey, line 764: Do we still have the race here after this PR? Do we need this comment still?

nbrownus: Good catch, that was a note to myself the other night. Removed

wadey (approved):

PR #940, opened 2023-07-31
Limit how often a busy tunnel can requery the lighthouse

nbrownus:
Active tunnels cause nebula to re-query the lighthouse every so often, roughly every 5k packets, to make sure we are using the best remote ip address possible for the remote host. On busy tunnels this limit can be hit very quickly and cause unnecessary load on the lighthouses.

This change adds a way to limit the minimum amount of time before the next query is allowed. `timers.requery_wait_seconds` is set to 60 seconds by default and relies on the original counter behavior prior to evaluation. This means an infrequently active tunnel could miss a re-query if the stars align just right.

This also moves the try promote and re-query every counter to config options for a matter of completeness. Initial performance tests look equivalent but I am not opposed to only adding the 1 new config option and leaving the 2 old items as constants.

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -418,7 +418,7 @@ func (f *Interface) reloadMisc(c *config.C) {
 	if c.HasChanged("timers.requery_wait_duration") {
 		n := c.GetDuration("timers.requery_wait_duration", defaultReQueryWait)
 		f.reQueryWait.Store(int64(n))
-		f.l.Info("timers.requery_wait_duration")
+		f.l.Info("timers.requery_wait_duration has changed")
 	}
 }
 
--- a/interface.go
+++ b/interface.go
@@ -412,7 +412,7 @@ func (f *Interface) reloadMisc(c *config.C) {
 	if c.HasChanged("counters.requery_every_packets") {
 		n := c.GetUint32("counters.requery_every_packets", defaultReQueryEvery)
 		f.reQueryEvery.Store(n)
-		f.l.Info("counters.requery_every_packets")
+		f.l.Info("counters.requery_every_packets has changed")
 	}
 
 	if c.HasChanged("timers.requery_wait_duration") {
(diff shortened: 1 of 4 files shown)


interface.go
```diff
@@ -389,6 +402,26 @@ func (f *Interface) reloadSendRecvError(c *config.C) {
...
+	if c.HasChanged("counters.requery_every_packets") {
+		n := c.GetUint32("counters.requery_every_packets", defaultReQueryEvery)
+		f.reQueryEvery.Store(n)
+		f.l.Info("counters.requery_every_packets")
+	}
+
+	if c.HasChanged("timers.requery_wait_duration") {
+		n := c.GetDuration("timers.requery_wait_duration", defaultReQueryWait)
+		f.reQueryWait.Store(int64(n))
+		f.l.Info("timers.requery_wait_duration")
```
wadey, line 421:
```suggestion
		f.l.Info("timers.requery_wait_duration has changed")
```

interface.go
```diff
@@ -389,6 +402,26 @@ func (f *Interface) reloadSendRecvError(c *config.C) {
...
+	if c.HasChanged("counters.try_promote") {
+		n := c.GetUint32("counters.try_promote", defaultPromoteEvery)
+		f.tryPromoteEvery.Store(n)
+		f.l.Info("counters.try_promote has changed")
+	}
+
+	if c.HasChanged("counters.requery_every_packets") {
+		n := c.GetUint32("counters.requery_every_packets", defaultReQueryEvery)
+		f.reQueryEvery.Store(n)
+		f.l.Info("counters.requery_every_packets")
```
wadey, line 415:
```suggestion
		f.l.Info("counters.requery_every_packets has changed")
```

wadey (commented):
Added a few comments, can you also document this change in the example config?

wadey (approved):

PR #952, opened 2023-08-08
Combine ca, cert, and key handling

nbrownus:
The goal here is to gather all the pki config into a single struct that can be passed around and consumed safely. A future PR will remove some references to `Interface` trading it for its own local pointer to `PKI`.


pki.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	c.RegisterReloadCallback(func(c *config.C) {
+		cErr := pki.reload(c, false)
+		if cErr != nil {
+			cErr.Log(l)
+		}
+	})
+
+	l.Error("SO NO ERROR THEN")
```
wadey, line 46: Maybe dont need this log line or the one on line 35

pki.go
```diff
@@ -0,0 +1,250 @@
...
+
+func (p *PKI) GetCertState() *CertState {
+	return p.cs.Load()
+}
+
+func (p *PKI) GetCAPool() *cert.NebulaCAPool {
+	return p.caPool.Load()
+}
+
+func (p *PKI) reload(c *config.C, initial bool) *util.ContextualError {
```
wadey, line 58: Usually its a bad pattern to return anything but the `error` interface or weird things can happen with non-nil types and stuff if something tries to wrap this as an `error` later.

https://go.dev/doc/faq#nil_error

> It's a good idea for functions that return errors always to use the error type in their signature (as we did above) rather than a concrete type such as *MyError, to help guarantee the error is created correctly. As an example, [os.Open](https://go.dev/pkg/os/#Open) returns an error even though, if not nil, it's always of concrete type [*os.PathError](https://go.dev/pkg/os/#PathError).

It's probably okay since these methods are private and internal, but it can cause really hard to figure out bugs

pki.go
```diff
@@ -0,0 +1,250 @@
...
+			)
+		}
+	}
+
+	p.cs.Store(cs)
+	if initial {
+		p.l.WithField("cert", cs.Certificate).Debug("Client nebula certificate")
+	} else {
+		p.l.WithField("cert", cs.Certificate).Info("Client cert refreshed from disk")
+	}
```
wadey, line 103: If we are going to log this when refreshed, I wonder if we should just log it on initial too (otherwise, you are missing the initial certificate in the logs, only refreshed ones).

pki.go
```diff
@@ -0,0 +1,250 @@
...
+	// Marshal the certificate to ensure it is valid
+	rawCertificate, err := certificate.Marshal()
+	if err != nil {
+		return nil, fmt.Errorf("invalid nebula certificate on interface: %s", err)
+	}
+
+	publicKey := certificate.Details.PublicKey
+	cs := &CertState{
+		RawCertificate: rawCertificate,
+		Certificate:    certificate, // PublicKey has been set to nil above
```
wadey, line 128:
```suggestion
		Certificate:    certificate,
```

I think this comment is a lie now? It actually gets set to nil below on line 133.

pki.go
nbrownus: Totally agree, just kind of lost on the approach we should take going forward.

We are moving towards this pattern where a struct can initialize itself from config and can handle its own reload using mostly the same code path, I think its a good direction to move in. But it creates some awkwardness with error handling and we've solved these situation in the past with `util.ContextualError`. 

if we return `error` instead of `*util.ContextualError` then we need to check each error return where we may have a `util.ContextualError` to avoid adding further context and making the log line silly.

Basically I am hoping to avoid this becoming pervasive in `main` https://github.com/slackhq/nebula/blob/master/main.go#L225-L230

Instead of returning the error I guess we could handle logging in the `New*` functions and return an `ok`, thoughts?

pki.go
nbrownus: Yeah this felt odd to do but I was matching existing behavior https://github.com/slackhq/nebula/blob/master/main.go#L66-L71

pki.go
nbrownus: @wadey had the genius idea of just using helper functions so I moved to that.

PR #953, opened 2023-08-08
Reduce what `HostInfo.ConnectionState` keeps track of

nbrownus:
The `ConnectionState` struct doesn't need the entire `CertState` object, this reworks things slightly to make it so we only need to keep track of the local certificate.

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -406,7 +406,7 @@ func (n *connectionManager) shouldSwapPrimary(current, primary *HostInfo) bool {
 	}
 
 	certState := n.intf.pki.GetCertState()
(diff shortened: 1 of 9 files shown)


wadey (dismissed):
I think we had to do this before because things would expect the cs to already be set and to be able to lock around it as soon as the hostinfo entered the pending host map: https://github.com/slackhq/nebula/pull/535

I think with the refactor to remove pending Hostmap this shouldn't be a problem, but we should ensure we have removed the race conditions that caused us to need to fix #535 in the first place before we merge this.

wadey (approved):

PR #954, opened 2023-08-09
Simplify getting a hostinfo or starting a handshake with one

nbrownus:
This PR is based on #953. The base branch will be changed to `master` when it merges.

The goal here is to simplify getting a hostinfo and starting a handshake, paving the way to invert how handshake state is tracked so that we don't have to keep all of it around in a fully formed hostinfo.

This should also resolve the last remaining racy issues with interactions between the pending and main hostmap.

There are 2 trouble spots here. 
1. `HandshakeManager.allocateIndex` acquires a main hostmap read lock and a pending hostmap write lock while it holds a hostinfo lock. I haven't found anywhere where that order is violated yet but it's worth some extra eyes.
2. The big lie correction in `handshake_ix.go` is inherently racy, I think it's a little better now than it was before but solving it entirely requires inverting the interactions with `HandshakeManager`.

### DIFF
--- a/inside.go
+++ b/inside.go
@@ -44,7 +44,10 @@ func (f *Interface) consumeInsidePacket(packet []byte, fwPacket *firewall.Packet
 		return
 	}
 
-	hostinfo := f.getOrHandshake(fwPacket.RemoteIP)
+	hostinfo, ready := f.getOrHandshake(fwPacket.RemoteIP, func(h *HostInfo) {
+		h.unlockedCachePacket(f.l, header.Message, 0, packet, f.sendMessageNow, f.cachedPacketMetrics)
+	})
+
 	if hostinfo == nil {
 		f.rejectInside(packet, out, q)
 		if f.l.Level >= logrus.DebugLevel {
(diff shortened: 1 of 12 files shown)


inside.go
```diff
@@ -44,7 +44,10 @@ func (f *Interface) consumeInsidePacket(packet []byte, fwPacket *firewall.Packet
 		return
 	}
 
-	hostinfo := f.getOrHandshake(fwPacket.RemoteIP)
+	hostinfo, ready := f.getOrHandshake(fwPacket.RemoteIP, func(h *HostInfo) {
+		h.unlockedCachePacket(f.l, header.Message, 0, packet, f.sendMessageNow, f.cachedPacketMetrics)
+	})
```
wadey, line 49: creating this closure for every packet on the hot patch shouldn't be an issue right? Hopefully it doesn't create garbage, at least if not used?

wadey (approved):
I think overall this looks good. I do see some races that could still happen due to the transition from handshakeManager to main hostmap, but I dont think there is an easy way to solve them unless we only had one map that contained both pending and active.

inside.go
nbrownus: In a further branch I have a handshake benchmark and pprof does not show any allocations here.

wadey (approved):

PR #955, opened 2023-08-15
Clean up a hostinfo to reduce memory usage

nbrownus:
This removes a lot of state from `HostInfo` to reduce our memory footprint for active tunnels.


handshake_ix.go
```diff
@@ -350,22 +340,6 @@ func ixHandshakeStage2(f *Interface, addr *udp.Addr, via *ViaSender, hostinfo *H
 	}
 
 	ci := hostinfo.ConnectionState
-	if ci.ready {
```
wadey, line 353: if we remove this ci.ready check, I think we need a different double check after the `hh.Lock()` to see if a different thread already completed the handshake for this host? How do we protected about late handshake packets coming in?

handshake_ix.go
```diff
@@ -140,9 +140,6 @@ func ixHandshakeStage1(f *Interface, addr *udp.Addr, via *ViaSender, packet []by
 		},
 	}
 
-	hostinfo.Lock()
-	defer hostinfo.Unlock()
```
wadey, line 144: The race detector might get upset about this, even if its okay in theory. I guess we can see what happens.

handshake_manager.go
```diff
@@ -64,10 +64,44 @@ type HandshakeManager struct {
...
+		hh.packetStore = append(hh.packetStore, &cachedPacket{t, st, f, tempPacket})
+		if l.Level >= logrus.DebugLevel {
+			hh.hostinfo.logger(l).
+				WithField("length", len(hh.packetStore)).
+				WithField("stored", true).
+				Debugf("Packet store")
+		}
+
+	} else if l.Level >= logrus.DebugLevel {
+		m.dropped.Inc(1)
```
wadey, line 93: This metric increment should be outside the DebugLevel check.

handshake_manager.go
```diff
@@ -108,41 +167,35 @@ func (c *HandshakeManager) NextOutboundHandshakeTimerTick(now time.Time) {
...
-	if hostinfo.HandshakeComplete {
-		// Ensure we don't exist in the pending hostmap anymore since we have completed
-		c.DeleteHostInfo(hostinfo)
+func (hm *HandshakeManager) handleOutbound(vpnIp iputil.VpnIp, lighthouseTriggered bool) {
+	hh := hm.queryVpnIp(vpnIp)
+	if hh == nil {
 		return
 	}
+	hh.Lock()
+	defer hh.Unlock()
```
wadey, line 176: I think everytime we do a lock like this, we need to double check afterwards if another thread beat us to finishing out this entry, right? (or add a comment about why no-one could have beat us to it).

outside.go
```diff
@@ -455,9 +455,6 @@ func (f *Interface) handleRecvError(addr *udp.Addr, h *header.H) {
 		return
 	}
 
-	hostinfo.Lock()
-	defer hostinfo.Unlock()
```
wadey, line 459: Why don't we need this anymore? Shouldn't be at least doublecheck that someone else didn't already handle a recv_error for this hostinfo on a different thread?

wadey (commented):
First read through, mostly concerns about the locking changes

handshake_ix.go
wadey: It looks like its fine, at least running `make smoke-docker-race` shows no race detector complaints during the smoke test.

handshake_ix.go
nbrownus: I expect in this version it would fail when trying to read a noise message. The next PR in the series would put these calls to `ixHandshake*` within the pending hostmap lock to avoid these races.

wadey (dismissed):

wadey (approved):

PR #958, opened 2023-08-21
smoke-test: don't assume docker needs sudo

wadey:
Let the host deal with this detail if necessary


JohnMaguire (approved):
CI is passing, so this LGTM.

PR #981, opened 2023-09-22
update to go1.22

wadey:
Since the first minor version update has already been released, we can probably feel comfortable updating to go1.21. This version now enforces that the go version on the system is compatible with the version specified in go.mod, so we can remove the old logic around checking the minimum version in the Makefile.

- https://go.dev/doc/go1.21#tools

> To improve forwards compatibility, Go 1.21 now reads the go line in a go.work or go.mod file as a strict minimum requirement: go 1.21.0 means that the workspace or module cannot be used with Go 1.20 or with Go 1.21rc1. This allows projects that depend on fixes made in later versions of Go to ensure that they are not used with earlier versions. It also gives better error reporting for projects that make use of new Go features: when the problem is that a newer Go version is needed, that problem is reported clearly, instead of attempting to build the code and printing errors about unresolved imports or syntax errors.

Also, switch off of `go-version-file` in the Github Actions, because it doesn't work great with the new `go mod tidy` format that go 1.22 does. See: 

- https://github.com/actions/setup-go/pull/460
- https://github.com/actions/setup-go/issues/457

We have to make a few fixes for gvisor for go1.22 compatibility, I merged in the fixes from #1093 to resolve those.


brad-defined:
https://tip.golang.org/doc/go1.21#windows
> As [announced](https://tip.golang.org/doc/go1.20#windows) in the Go 1.20 release notes, Go 1.21 requires at least Windows 10 or Windows Server 2016; support for previous versions has been discontinued.

I think the only tricky thing about this is whether we want to drop support for older versions of Windows.

wadey:
Sounds good, let’s move this to Nebula 1.9 at least

wadey:
> EDIT: attempting to patch to build on Alpinelinux

@andypost why do you need to patch out CGO_ENABLED ? Nebula shouldn't require CGO for standard builds.

wadey:
Updated to point to go1.22

.github/workflows/gofmt.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
     - uses: actions/setup-go@v5
       with:
-        go-version-file: 'go.mod'
+        go-version: '1.22'
```
nbrownus, line 21: Why move back to declaring the version here?

nbrownus (approved):

Issue #1005, opened 2023-11-03
Feature request: publish firewall hash to Prometheus


britcey:
Hi all, 

It'd be great if the firewall hash were published with the Prometheus stats - it'd be a useful mechanism to alert when a client's firewall rules change.

`level=info msg="Firewall started" firewallHash=3775881c4dfa7c63e3d343e747ad66b0ddf6a72fde1ce3b2eea43436175f101f`

The gotcha is that it'd need to be a float64 (we'd want it as the value, not as a label);  blackbox_exporter uses a FNV-1  hash for that sort of thing:

https://github.com/prometheus/blackbox_exporter/blob/master/prober/utils.go#L136

https://pkg.go.dev/hash/fnv

Would it be as simple as updating firewall.go to add a FNV-1 hash function alongside GetRuleHash() and adding another metrics.GetOrRegisterGauge in EmitStats with that hash value?

britcey:
It appears it is indeed that simple:

https://github.com/slackhq/nebula/pull/1010

PR #1010, opened 2023-11-10
Added firewall.rules.hash metric

britcey:
Added a FNV-1 hash of the firewall rules as a Prometheus value.

### DIFF
--- a/firewall.go
+++ b/firewall.go
@@ -279,16 +279,16 @@ func (f *Firewall) GetRuleHash() string {
 	return hex.EncodeToString(sum[:])
 }
 
-// GetRuleHashFNV returns a int64 FNV-1 hash representation the rules, for use as a metric value
-func (f *Firewall) GetRuleHashFNV() int64 {
-	h := fnv.New64a()
+// GetRuleHashFNV returns a uint32 FNV-1 hash representation the rules, for use as a metric value
+func (f *Firewall) GetRuleHashFNV() uint32 {
+	h := fnv.New32a()
 	h.Write([]byte(f.rules))
-	return int64(h.Sum64())
+	return h.Sum32()
 }
 
 // GetRuleHashes returns both the sha256 and FNV-1 hashes, suitable for logging
 func (f *Firewall) GetRuleHashes() string {
-	return "SHA:" + f.GetRuleHash() + ",FNV:" + strconv.FormatInt(f.GetRuleHashFNV(), 10)
+	return "SHA:" + f.GetRuleHash() + ",FNV:" + strconv.FormatUint(uint64(f.GetRuleHashFNV()), 10)
 }
 
 func AddFirewallRulesFromConfig(l *logrus.Logger, inbound bool, c *config.C, fw FirewallInterface) error {
--- a/firewall.go
+++ b/firewall.go
@@ -279,11 +279,16 @@ func (f *Firewall) GetRuleHash() string {
 	return hex.EncodeToString(sum[:])
 }
 
-// GetRuleHashFNV returns a float64 FNV-1 hash representation the rules, suitable for use as a Prometheus value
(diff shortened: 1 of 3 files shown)


britcey:
Closes issue #1005

nbrownus:
Thanks for the PR! My primary concern is that there is no direct correlation back to the log line. I think this could be resolved in two ways.

1. Include the FNV hash in the log lines
2. Include the sha256 hash as a label

Ideally we would do both here but option 2 is more difficult so lets add the FNV hash to the log lines.

firewall.go
```diff
@@ -278,6 +279,13 @@ func (f *Firewall) GetRuleHash() string {
 	return hex.EncodeToString(sum[:])
 }
 
+// GetRuleHashFNV returns a float64 FNV-1 hash representation the rules, suitable for use as a Prometheus value
+func (f *Firewall) GetRuleHashFNV() float64 {
+	h := fnv.New32a()
```
nbrownus, line 284: Metrics are `int64` so we should use `fnv.New64a` and return a `int64` instead of a `float64`, will help avoid the already unlikely possibility of a colliding hash.

britcey:
> Thanks for the PR! My primary concern is that there is no direct correlation back to the log line. I think this could be resolved in two ways.
> 
> 1. Include the FNV hash in the log lines
> 2. Include the sha256 hash as a label
> 
> Ideally we would do both here but option 2 is more difficult so lets add the FNV hash to the log lines.

Option 2 would also increase cardinality, creating a new timeseries every time that label changed, so including the FNV hash in the log is the way to go, for sure.  I'll make that happen.

firewall.go
britcey: Just so I'm clear - your particular implementation is using int64 for the metric values?  Because the Prometheus data model is float64.

nbrownus: Yep, https://pkg.go.dev/github.com/rcrowley/go-metrics#Gauge

britcey: n/m, I see that go-metrics does indeed use int64 - https://pkg.go.dev/github.com/rcrowley/go-metrics?utm_source=godoc#Gauge

britcey:
This work for the logs?

Startup:

`level=info msg="Firewall started" firewallHashes="SHA:8cdd862ba57eca683ac4fb32fcd8f727743df3ba0c92b267c3e9ca128e43be8d,FNV:-4874977699962996863"`

after a reload:

`level=info msg="New firewall has been installed" firewallHashes="SHA:3775881c4dfa7c63e3d343e747ad66b0ddf6a72fde1ce3b2eea43436175f101f,FNV:7449190936218615427" oldFirewallHashes="SHA:8cdd862ba57eca683ac4fb32fcd8f727743df3ba0c92b267c3e9ca128e43be8d,FNV:-4874977699962996863" rulesVersion=1`

firewall.go
wadey: The go-metrics integration with prometheus does convert it back to a float64 though. So if we generate an int64 it will be lossy: https://github.com/deathowl/go-metrics-prometheus/blob/f2a1482ba35b474b8a55c052f7333c614b93d36b/prometheusmetrics.go#L156-L157

I think the best bet would be to go back to fnv.New32a, treat it internally as a `uint32`, but the gauge can cast it to `int64` (and then it won't be lossy when prom integration converts it to a float64, since a uint32 should fit in there safely).

nbrownus: Yep, agreed and that is super frustrating. Sorry for the run around.

britcey: No problem - back to New32a and using it as a uint32 internally.

britcey:
Anything needed here?

johnmaguire (approved):

johnmaguire:
@britcey Thanks for your contribution!

PR #1012, opened 2023-11-13
Multipath unsafe routes

forfuncsake:
This changeset adds support for discovering weighted multipath routes from the system route table on linux.


forfuncsake:
While this was a working implementation, there are a couple of open questions about how we should treat same-weight routes. TCP connections would likely need to remain sticky to a tunnel/route (where possible), particularly if using NAT on the host that is forwarding to the unsafe network, if the route became unavailable, we could consider sending a reset back to force the client to start a new TCP session over the new path, but obviously there are also a lot of other corner cases.

Secondarily, the feature was broken by #1004 and the use case for which I was experimenting with this became a bit too complex.

Closing for now, we can revisit if there is a future need.

PR #1024, opened 2023-11-24
Update Arch Linux package link

dumbasPL:
The community repository was merged into extra during the Arch Linux [Git migration](https://archlinux.org/news/git-migration-announcement/) and no longer exists

PS: I've noticed there is a package for nebula in [Alpine Linux](https://pkgs.alpinelinux.org/packages?name=nebula), Wouldn't it be a good idea to add it to the list as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,20 @@ Check the [releases](https://github.com/slackhq/nebula/releases/latest) page for
     ```
     $ sudo pacman -S nebula
     ```
+
 - [Fedora Linux](https://src.fedoraproject.org/rpms/nebula)
     ```
     $ sudo dnf install nebula
+    ```
+
+- [Debian Linux](https://packages.debian.org/source/stable/nebula)
+    ```
+    $ sudo apt install nebula
+    ```
+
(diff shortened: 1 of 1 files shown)


nbrownus:
Thanks for the update!

> PS: I've noticed there is a package for nebula in [Alpine Linux](https://pkgs.alpinelinux.org/packages?name=nebula), Wouldn't it be a good idea to add it to the list as well?

Certainly, do you mind adding it to the PR?

dumbasPL:
Sorry for the delay, I forgot about this PR. I've also added a Debian link since that's a popular server distro.

Note about Alpine: unfortunately, they don't have a nice "stable" link. The current one without a branch specified points to the `edge` branch. The only potential problem with this is that users might get confused about the version available without noticing that they are looking at the edge (testing) branch. Since Alpine isn't a rolling release distro, the stable version will sometimes be a few versions behind. The alternative would be to put a link the the current stable branch but that also means that it will need to be updated every time Alpine releases a new version.

nbrownus (approved):

PR #1030, opened 2023-11-29
SIGHUP is only useful when config was loaded from a file

forfuncsake:
Have (*config.C).CatchHUP() return early when there is no file path available from which to reload.
This will allow wrapping service to manage their own signal trapping (which is particularly important if they've used config from a string.


nbrownus:
We had discussed removing `C.CatchHUP()` from the `config` package entirely and having the `cmd` package handle the call to `C.ReloadConfig()`. I think this is the more correct approach for anyone using an alternate config loading path (likely as a library) and presumably would enjoy having more control of the signal handling.

Happy to merge but would also like to come back and do the other work as after 1.8.

PR #1032, opened 2023-12-01
cmd/nebula-cert: fix clobbered error

alrs:
This picks up a dropped `err` clobbered by the use of `:=` in an `if` block.


cmd/nebula-cert/ca.go
```diff
@@ -212,8 +212,9 @@ func ca(args []string, out io.Writer, errOut io.Writer, pr PasswordReader) error
 		return fmt.Errorf("error while signing: %s", err)
 	}
 
+	var b []byte
 	if *cf.encryption {
-		b, err := cert.EncryptAndMarshalSigningPrivateKey(curve, rawPriv, passphrase, kdfParams)
+		b, err = cert.EncryptAndMarshalSigningPrivateKey(curve, rawPriv, passphrase, kdfParams)
 		if err != nil {
 			return fmt.Errorf("error while encrypting out-key: %s", err)
 		}
```
nbrownus, line 220: Nice catch, the github suggestion thing is a bit weird but I think it would be more clear if this block prepared `b` in either case and let the code lower down handle the file write/error

```suggestion
	var b []byte
	if *cf.encryption {
		b, err = cert.EncryptAndMarshalSigningPrivateKey(curve, rawPriv, passphrase, kdfParams)
		if err != nil {
			return fmt.Errorf("error while encrypting out-key: %s", err)
		}
	} else {
		b = cert.MarshalSigningPrivateKey(curve, rawPriv)
	}
	err = os.WriteFile(*cf.outKeyPath, b, 0600)
	if err != nil {
		return fmt.Errorf("error while writing out-key: %s", err)
	}
```

nbrownus (approved):

PR #1037, opened 2023-12-08
Push Docker images as part of the release workflow

### LINKED ISSUE #25: Docker image

johnmaguire:
Closes #25.

## Description

This PR adds a Github Actions workflow to build and push Docker images whenever a new release is tagged. The image is pushed with a version tag, as well as with the `latest` tag. This allows users to choose between manual updates or automatic updates by pinning to the `latest` tag.

With this image, you can provide Nebula connectivity to your Docker host while running Nebula inside a container:

```
docker run \
    --name nebula \
    --network host \
    --cap-add NET_ADMIN \
    --volume ./config:/etc/nebula \
    --rm \
    nebulaoss/nebula
```

In the example above, you must create a `config` directory containing a valid `config.yml`. Here is the same configuration, but with docker-compose:

```
version: '2'
services:
  nebula:
    image: nebulaoss/nebula
    network_mode: host
    cap_add:
      - NET_ADMIN
    volumes:
      - ./config:/etc/nebula
    restart: unless-stopped
```

This image also allows overriding the args passed to Nebula:

```
❯ docker run nebulaoss/nebula -version
Version: 0.0.11
```

This can be used to allow using a config directory instead of a config file, for example.

## Github Actions Details

If `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` are not set, every step in the `build-docker` job of the release workflow is skipped. This avoids a "failed" status in forks that want to create repo-local releases without pushing to Docker. The tradeoff is that it could indicate a false success when the required variables are not configured.

### Variables

Required:
 - `DOCKERHUB_USERNAME` (variable)
 - `DOCKERHUB_TOKEN` (secret) - must be read/write
 
Optional:
 - `DOCKER_REPO_NAME` (variable) - defaults to "nebulaoss/nebula"
 - `DOCKER_REPO_TAG` (variable) - defaults to "latest" (e.g. can be overridden to "unstable")

## Todo
- [x] Configure Docker secrets in this repo
- [ ] Update `nebulaoss/nightly` with its own Docker secrets and repo


Makefile
```diff
@@ -30,6 +30,13 @@ ifndef BUILD_NUMBER
 	endif
 endif
 
+ifndef DOCKER_IMAGE_NAME
+	DOCKER_IMAGE_NAME = nebula
+endif
+ifndef DOCKER_IMAGE_TAG
+	DOCKER_IMAGE_TAG = lates
```
ruzko, line 37: Should this be `DOCKER_IMAGE_TAG = latest` instead?

johnmaguire: It sure should, thanks for that!

johnmaguire:
I just noticed that this branch had old code, and was not up-to-date with my personal repo's branch. I just pushed an update.

Makefile
```diff
@@ -151,6 +160,9 @@ build/nebula-%.tar.gz: build/%/nebula build/%/nebula-cert
 build/nebula-%.zip: build/%/nebula.exe build/%/nebula-cert.exe
 	cd build/$* && zip ../nebula-$*.zip nebula.exe nebula-cert.exe
 
+docker/%: build/%/nebula build/%/nebula-cert .FORCE
```
johnmaguire, line 163:
```suggestion
docker/%: build/%/nebula build/%/nebula-cert
```

nitpick

johnmaguire (commented):
@wadey The changes look great. I really like the distroless approach! Thanks for all the help here!

Makefile
wadey: added, thanks!

Makefile
```diff
@@ -22,6 +22,13 @@ ifndef BUILD_NUMBER
 	endif
 endif
 
+ifndef DOCKER_IMAGE_REPO
+	DOCKER_IMAGE_REPO = nebulaoss/nebula
+endif
+ifndef DOCKER_IMAGE_TAG
+	DOCKER_IMAGE_TAG = latest
+endif
+
```
wadey, line 31:
```suggestion
DOCKER_IMAGE_REPO ?= nebulaoss/nebula
DOCKER_IMAGE_TAG ?= latest
```

if we define it this way, you should still be able to override it.

wadey (approved):

PR #1044, opened 2023-12-17
Fix re-entrant `GetOrHandshake` issues

nbrownus:
@wadey got a deadlock with v1.8.0 and was able to pull the stack trace. `HandshakeManager.GetOrHandshake` can be re-entrant via `HandshakeManager.StartHandshake` through a call to `hm.lightHouse.QueryServer()`. Aside from the double read lock on the main hostmap not being great, the `ConnectionManager` go routine had fired between the 1st and 2nd calls to `HandshakeManager.GetOrHandshake` and was waiting on a write lock for the main hostmap while blocking any future read locks. 

This is fixed by adding a channel and handling the actual lighthouse queries in a go routine. Should also speed up the hot path when many handshakes are occurring.

There is also a case when a tunnel is being tested and is using a relay for a double read lock in `ConnectionManager`.

This is fixed by turning the test packet into a traffic decision result and handling outside of the read lock.

My primary concern is in handling the `QueryServer` writes on a nonblocking buffered channel. I think we will want to block when full but I am leaving as nonblocking for now to review.

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -74,6 +74,8 @@ type LightHouse struct {
 	// IP's of relays that can be used by peers to access me
 	relaysForMe atomic.Pointer[[]iputil.VpnIp]
 
+	queryChan chan iputil.VpnIp
+
 	calculatedRemotes atomic.Pointer[cidr.Tree4[[]*calculatedRemote]] // Maps VpnIp to []*calculatedRemote
 
 	metrics           *MessageMetrics
@@ -110,6 +112,7 @@ func NewLightHouseFromConfig(ctx context.Context, l *logrus.Logger, c *config.C,
 		nebulaPort:   nebulaPort,
 		punchConn:    pc,
 		punchy:       p,
+		queryChan:    make(chan iputil.VpnIp, c.GetUint32("handshakes.query_buffer", 64)),
 		l:            l,
 	}
 	lighthouses := make(map[iputil.VpnIp]struct{})
@@ -138,6 +141,8 @@ func NewLightHouseFromConfig(ctx context.Context, l *logrus.Logger, c *config.C,
 			l.WithError(err).Error("failed to reload lighthouse")
 		}
 	})
+
+	h.startQueryWorker()
 
 	return &h, nil
 }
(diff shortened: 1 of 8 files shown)


lighthouse.go
```diff
@@ -456,29 +461,18 @@ func (lh *LightHouse) Query(ip iputil.VpnIp, f EncWriter) *RemoteList {
...
-	lighthouses := lh.GetLighthouses()
-	lh.metricTx(NebulaMeta_HostQuery, int64(len(lighthouses)))
-	nb := make([]byte, 12, 12)
-	out := make([]byte, mtu)
-	for n := range lighthouses {
-		f.SendMessageToVpnIp(header.LightHouse, 0, n, query, nb, out)
+	default:
+		if lh.l.Level >= logrus.DebugLevel {
+			lh.l.WithField("vpnIp", ip).Debug("Lighthouse query buffer was full, dropping request")
+		}
```
wadey, line 475: I wonder if this should be higher than debug, since without debug logs on it would be hard to tell this is happening and that you need to increase the buffer.

wadey (approved):
approved with a small comment

lighthouse.go
nbrownus: I think it might be better to just make this a blocking write to a buffered channel

wadey (approved):

lighthouse.go
```diff
@@ -456,30 +461,12 @@ func (lh *LightHouse) Query(ip iputil.VpnIp, f EncWriter) *RemoteList {
...
-	}
-
-	lighthouses := lh.GetLighthouses()
-	lh.metricTx(NebulaMeta_HostQuery, int64(len(lighthouses)))
-	nb := make([]byte, 12, 12)
-	out := make([]byte, mtu)
-	for n := range lighthouses {
-		f.SendMessageToVpnIp(header.LightHouse, 0, n, query, nb, out)
-	}
+	lh.queryChan <- ip
```
brad-defined, line 469: The channel consumer in startQueryWorker() calls lh.ifce.SendMessageToVpnIp() calls f.getOrHandshake() calls hm.StartHandshake() calls hm.lightHouse.QueryServer()

This call chain means that the channel consumer may also attempt to write to the channel. If the channel is full, this write operation will block forever, as the only thing that could make room in the channel is blocked waiting for room in the channel.

brad-defined (changes requested):
I might have misread it, but I think the implementation may block when the channel is full.

lighthouse.go
brad-defined: The call stack can only happen if QueryWorker doesn't have a tunnel to some lighthouse. With an early exit in QueryServer for lighthouse IP's, then QueryWorker will never try to write to the channel that it's reading from.

brad-defined (approved):

wadey (approved):

PR #1054, opened 2023-12-26
Support inlined sshd host key

ccat3z:
Make `sshd.host_key` accept inlined key like `pki.ca`.

``` yaml
sshd:
  host_key: |
    -----BEGIN OPENSSH PRIVATE KEY-----
    ...
```


johnmaguire (approved):

johnmaguire:
Thanks for this change - this has always been a tad annoying! :-)

PR #1071, opened 2024-02-05
`local_cidr` moved to always be evaluated

nbrownus:
`local_cidr` was added to firewall rules recently and the current implementation only looks at `local_cidr` if the `host`, `groups`, and `cidr` do not already match. This greatly limits the ability to control what an unsafe router allows for traffic flows.

This change moves `local_cidr` to its own evaluation group that will be checked after doing host identity matching (`host`, `groups`, or `cidr`). If the local certificate has any subnets then all rules will have the local certificates ip address as the `local_cidr` by default to block any unintended exposure via an unsafe route.

This is a breaking change and will require anyone that uses unsafe routes to change their firewall rules to include a `local_cidr` for the traffic they want to allow.

### DIFF
--- a/cidr/tree4.go
+++ b/cidr/tree4.go
@@ -144,7 +144,7 @@ func (tree *Tree4[T]) MostSpecificContains(ip iputil.VpnIp) (ok bool, value T) {
 
 type eachFunc[T any] func(T) bool
 
-// EachContains will call a function, passing the value, for each entry until the function returns false or the search is complete
+// EachContains will call a function, passing the value, for each entry until the function returns true or the search is complete
 // The final return value will be true if the provided function returned true
 func (tree *Tree4[T]) EachContains(ip iputil.VpnIp, each eachFunc[T]) bool {
 	bit := startbit
--- a/cidr/tree4.go
+++ b/cidr/tree4.go
@@ -142,15 +142,22 @@ func (tree *Tree4[T]) MostSpecificContains(ip iputil.VpnIp) (ok bool, value T) {
 	return ok, value
 }
 
-// Match finds the most specific match
-// TODO this is exact match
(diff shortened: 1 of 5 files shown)


firewall_test.go
```diff
@@ -226,123 +204,159 @@ func TestFirewall_Drop(t *testing.T) {
...
 				InvertedGroups: map[string]struct{}{"nope": {}},
 				Name:           "good-host",
 			},
 		}
 		for n := 0; n < b.N; n++ {
-			ft.match(firewall.Packet{Protocol: firewall.ProtoTCP, LocalPort: 100, LocalIP: ip}, true, c, cp)
+			ft.match(firewall.Packet{Protocol: firewall.ProtoTCP, LocalPort: 10}, true, c, cp)
 		}
 	})
+	//
```
brad-defined, line 306: intentionally commented out?

cidr/tree4.go
```diff
@@ -142,15 +142,22 @@ func (tree *Tree4[T]) MostSpecificContains(ip iputil.VpnIp) (ok bool, value T) {
...
 }
 
-// Match finds the most specific match
-// TODO this is exact match
-func (tree *Tree4[T]) Match(ip iputil.VpnIp) (ok bool, value T) {
+type eachFunc[T any] func(T) bool
+
+// EachContains will call a function, passing the value, for each entry until the function returns true or the search is complete
+// The final return value will be true if the provided function returned true
+func (tree *Tree4[T]) EachContains(ip iputil.VpnIp, each eachFunc[T]) bool {
```
brad-defined, line 149: Nit: `EachContains` sounds like each and every element must contain. I would call this one `AnyContains`, FWIW.

brad-defined (approved):

PR #1077, opened 2024-02-19
Increase default `lookup_timeout` from 250ms to 5s

maggie44:
I get a lot of errors in the logs like these, particularly when starting the service for the first time:

```
ERRO[0000] DNS resolution failed for static_map host     error="lookup example.com: i/o timeout" hostname=example.com network=ip4
```

When using multiple lighthouses some of them resolve ok, others timeout, and eventually they all resolve on future loops. The DNS resolution is working but the timeout is sometimes being reached before it has a time to finish. 

The timeout for these requests is current set at `250ms`. This is extremely low and can't see any reason why.

Here are some example defaults from elsewhere for some precedent:

https://github.com/istio/istio/blob/ac901c3ed1a2455705709bd5e81df781d7a63083/pilot/pkg/util/network/ip.go#L145
https://github.com/tailscale/tailscale/blob/a4a909a20b0f868de4870294e200e803f61589f7/ipn/localapi/debugderp.go#L161

This PR raises the default timeout to 5s.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -33,7 +33,7 @@ static_host_map:
   #network: ip4
 
   # lookup_timeout is the DNS query timeout.
-  #lookup_timeout: 250ms
+  #lookup_timeout: 5s
 
 lighthouse:
   # am_lighthouse is used to enable lighthouse functionality for a node. This should ONLY be true on nodes
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -383,7 +383,7 @@ func getStaticMapCadence(c *config.C) (time.Duration, error) {
 }
 
 func getStaticMapLookupTimeout(c *config.C) (time.Duration, error) {
-	lookupTimeout := c.GetString("static_map.lookup_timeout", "250ms")
+	lookupTimeout := c.GetString("static_map.lookup_timeout", "5s")
 	d, err := time.ParseDuration(lookupTimeout)
 	if err != nil {
 		return 0, err


SamSirry:
Hi @johnmaguire 
I came here after noticing numerous error entries in the journal for "DNS resolution failed for static_map host".
While I can adjust the timeout value in the config file (static_map.lookup_timeout), my concern is that an unharmful timeout event still appears in the log as an "error", and it's still cluttering the log while the logging level is set to "error", which draws unnecessary attention.

My suggestion is to report an "error" only when all static hosts are unresolvable/unreachable. Otherwise, a "warning" should be enough.

I believe the relevant code is around this line: https://github.com/slackhq/nebula/blob/50b24c102ebc01aae21f0bcfc2dfff08f0611fcf/remote_list.go#L122

What are your thoughts on this?

johnmaguire:
@SamSirry Hey there, sorry for the delay. For future reference, it's difficult to keep track of questions/feedback in comments of closed issues - it's probably best to file your own issue and we can follow up and close it if necessary.

I think you raise a good point that the log level here may be too high. I plan on raising this with the rest of the team next week to get their thoughts.

johnmaguire:
@maggie44 I understood the proposal to be changing the log level to a warning unless all DNS names for a given static host were unresolvable, in which case an error should be logged. (e.g. an internal DNS name only available on LAN and an external public DNS name)

I agree that any single unreachable static host is an error - it is not only used for lighthouses. It is possible to run a Lighthouse-less network through the static_host_map, for example.

PR #1091, opened 2024-03-13
Exploration of making config loading more errorful

jasikpark:
This PR pulls in https://github.com/siriusa51/goresult to allow returning an error from config-loading functions while also sticking with our existing behavior of setting a default value for options that fail to parse.

This has the advantage of allowing us to bubble up parse errors and to eventually make the config loading more strict, to provide better feedback when loading your `config.yml` and to better reflect the actual loaded config in nebula w.r.t. the on-disk representation.


johnmaguire:
I don't feel very comfortable with the import on siriusa51/goresult this adds. For one thing, Result types are not really idiomatic Go. Normally you would return two values instead. Secondly, this repo looks like a personal project that isn't actively maintained.

PR #1098, opened 2024-03-26
Add support for SSH CAs

### LINKED ISSUE #1076: Support access using SSH certificates + roles

1.8.2

N/A

### Describe the Bug

Users can be allowed to acces nebula SSH interface by having nebula trust certificates issues by an SSH certificate authority.

This feature should be accompanied by a set of options to trust a specific user name or users from a specific group - specified in the certificate.

One advantage of using an SSH CA is that you can also sign the host key by the CA and thus avoid the dangers of trusting the host key on first connect. 

A configuration could look like this

```
sshd:
  enabled: true
  listen: 127.0.0.1:2222
  host_key: /etc/ssh/ssh_host_ed25519_key
  certificate_authorities:
    - allow_groups:
        - sysadmins
        - team-dev1
      keys:
        - /etc/ssh/user_ca_key.pub
```

Some examples on how to setup ssh CA 
- https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Certificate-based_Authentication
- https://www.hashicorp.com/blog/managing-ssh-access-at-scale-with-hashicorp-vault

This issue is somewhat related to https://github.com/slackhq/nebula/issues/1051 . 

### Logs from affected hosts

N/A

N/A

johnmaguire:
This adds support for configuring SSH CAs. The SSH server will accept authentication for any valid certificate signed by an SSH CA, assuming that the user they connect with matches a principal in the certificate.

This would close #1076 if merged.

### DIFF
--- a/sshd/server.go
+++ b/sshd/server.go
@@ -1,6 +1,7 @@
 package sshd
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"net"
@@ -15,8 +16,11 @@ type SSHServer struct {
 	config *ssh.ServerConfig
 	l      *logrus.Entry
 
+	certChecker *ssh.CertChecker
+
 	// Map of user -> authorized keys
 	trustedKeys map[string]map[string]bool
+	trustedCAs  []ssh.PublicKey
 
 	// List of available commands
 	helpCommand *Command
@@ -31,6 +35,7 @@ type SSHServer struct {
 
 // NewSSHServer creates a new ssh server rigged with default commands and prepares to listen
 func NewSSHServer(l *logrus.Entry) (*SSHServer, error) {
+
 	s := &SSHServer{
 		trustedKeys: make(map[string]map[string]bool),
 		l:           l,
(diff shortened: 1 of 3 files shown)


sshd/server.go
```diff
@@ -31,15 +35,52 @@ type SSHServer struct {
...
+
+			return &ssh.Permissions{
+				// Record the public key used for authentication.
+				Extensions: map[string]string{
+					"fp":   fp,
+					"user": c.User(),
+				},
+			}, nil
+
+		},
```
forfuncsake, line 79: Could just reference the existing function here:

```suggestion
		UserKeyFallback: s.matchPubKey,
```

sshd/server.go
```diff
@@ -31,15 +35,52 @@ type SSHServer struct {
...
 		conns:       make(map[int]*session),
 	}
 
+	cc := ssh.CertChecker{
+		IsUserAuthority: func(auth ssh.PublicKey) bool {
+			for _, ca := range s.trustedCAs {
+				if bytes.Equal(ca.Marshal(), auth.Marshal()) {
+					return true
+				}
+				fmt.Println("didn't pass ca check")
```
forfuncsake, line 52: remove debug print?

sshd/server.go
forfuncsake: If you prefer to inline it, the original function no longer has any call sites and can be removed.

johnmaguire: Thanks - I preferred inlining it so all the auth checking code was in one place. Oops.

johnmaguire:
In #1076, it was requested that this support a concept of "groups" as SSH certificate principals. In other words, when creating a certificate, you would specify groups (e.g. `dev`, `ops`, `support`.) In the Nebula config file, you'd have a list of `allowed_princpials` which would only allow certificates containing one of these principals. I imagine this would be a per-CA option (and should alternatively allow for an "any" rule, omitting the list.)

This could get a little awkward though - the current implementation requires that the username you connect with match a principal in the certificate. Ultimately however, the debug SSH server has no concept of permissions, so this username is only used for logging purposes. We could update the log line to record both the "key ID" and all principals, in the case of an SSH cert.

wadey (approved):

PR #1099, opened 2024-03-28
Fix "any" firewall rules for unsafe_routes

johnmaguire:
When `default_local_cidr_any` is set to false, firewall rules that don't include a `local_ip` are treated as though they have a `local_ip` set to the Nebula IP address of the host. This ensures that firewall rules targeting ports on the host don't inadvertently allow traffic to unsafe_routes.

However, the conditional was a big over-eager, and treated explicit `0.0.0.0/0` the same as missing `local_ip`. This was a bug - we want users to be able to create "any" rules explicitly.


nbrownus (approved):
Nice catch!

PR #1132, opened 2024-04-30
Remove Arch nebula.service file

Antiz96:
As done for Fedora in https://github.com/slackhq/nebula/pull/1128, Arch Linux now also ships [their own systemd unit file](https://gitlab.archlinux.org/archlinux/packaging/packages/nebula/-/commit/08ea867ca9946a3a4fba5847c7150a1bf5b23af7) (as seen with @johnmaguire).


johnmaguire (approved):
Thanks for your help with this @Antiz96!

PR #1141, opened 2024-05-09
Added support for DNS SRV

QZAiXH:
Despite the fact that the operator did not provide me with a public IP, I am still able to indirectly map and obtain a public IP and port through the use of the [natmap](https://github.com/heiher/natmap) tool. By entering the acquired public IP and port into DNS SRV records, nebula can resolve these records and access my temporary public IP and port. This feature is incredibly useful to me.

### DIFF
--- a/remote_list.go
+++ b/remote_list.go
@@ -5,8 +5,10 @@ import (
 	"context"
 	"net"
 	"net/netip"
+	"regexp"
 	"sort"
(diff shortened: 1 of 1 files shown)


QZAiXH:
Could @nbrownus and @wadey please take a look at my PR? This feature is very useful for those without a public IP address.

PR #1148, opened 2024-05-22
upgrade to yaml.v3 

wadey:
The main nice fix here is that maps unmarshal into `map[string]any`
instead of `map[any]any`, so it cleans things up a bit.

The only gotcha I see for our uses is that the Yaml 1.1 bool types like `yes` won't automatically be set as a bool inside of `map[string]any`. So I added a little helper `config.AsBool` to keep backwards compat there.

- https://ubuntu.com/blog/api-v3-of-the-yaml-package-for-go-is-available


config/config.go
```diff
@@ -266,6 +266,22 @@ func (c *C) GetBool(k string, d bool) bool {
 	return v
 }
 
+func AsBool(v any) (value bool, ok bool) {
+	switch x := v.(type) {
+	case bool:
+		return x, true
+	case string:
+		switch x {
```
nbrownus, line 274: Does this work for `Yes`?

wadey: ah this is what they are doing in v3 if you unmarshal into a Bool:

- https://github.com/go-yaml/yaml/commit/01167db92b45abedd6aa8435e46d1e75097dcc4b

So we can probably copy that.

This is what v2 did:

- https://github.com/go-yaml/yaml/blob/v2.4.0/resolve.go#L37-L42

wadey: Existing in v1.9.5: https://github.com/slackhq/nebula/blob/v1.9.5/config/config.go#L257-L262

nbrownus (approved):

Issue #917, opened 2023-07-18
Feature request: Indicate when a lighthouse is unreachable at the ERROR log level


maggie44:
### What version of `nebula` are you using?

1.7.2

### What operating system are you using?

Mac

### Describe the Bug

When a node is started but the lighthouse is unavailable/unreachable there is no warning at log level 'ERROR'. At level INFO I see:

```
INFO[0007] Handshake timed out                           durationNs=7101752416 handshake="map[stage:1 style:ix_psk0]"
```

I think there should be some sort of indicator at the ERROR log level when the lighthouse cannot be reached. It is a breaking issue and without info here we assume things are connected. 

When enabling INFO I get far too much information streamed in to the console just to be able to see timeouts to the lighthouse. 

Example logging at INFO, without starting a lighthouse:

```
./nebula -config config.yml
INFO[0000] Firewall rule added                           firewallRule="map[caName: caSha: direction:outgoing endPort:0 groups:[] host:any ip: localIp: proto:0 startPort:0]"
INFO[0000] Firewall rule added                           firewallRule="map[caName: caSha: direction:incoming endPort:0 groups:[] host:any ip: localIp: proto:1 startPort:0]"
INFO[0000] Firewall rule added                           firewallRule="map[caName: caSha: direction:incoming endPort:443 groups:[laptop home] host: ip: localIp: proto:6 startPort:443]"
INFO[0000] Firewall started                              firewallHash=58e2e90dab9bc5d50b9e164e6b1b0cb09c5168fc46847e3b1b3a5026d4d3de63
WARN[0000] interface name must be utun[0-9]+ on Darwin, ignoring 
INFO[0000] Main HostMap created                          network=192.168.100.10/24 preferredRanges="[]"
INFO[0000] punchy enabled                               
INFO[0000] Loaded send_recv_error config                 sendRecvError=always
INFO[0000] Nebula interface is active                    boringcrypto=false build=1.7.2 interface=utun5 network=192.168.100.10/24 udpAddr="[::]:4242"
INFO[0000] Handshake message sent                        handshake="map[stage:1 style:ix_psk0]" initiatorIndex=2960355161 localIndex=2960355161 remoteIndex=0 udpAddrs="[100.64.22.11:4242]" vpnIp=192.168.100.1
[...]

PR #1150, opened 2024-05-27
Change 'Handshake timed out' log message to type: `Error`

maggie44:
Closes: https://github.com/slackhq/nebula/issues/917


johnmaguire:
Unfortunately, "Handshake timed out" occurs frequently in regular Nebula use even when there is no configuration or operational error in Nebula (e.g. due to network conditions, on either end.) Changing this log to "Error" would create too much noise in the logs.

Issue #328, opened 2020-11-02
Support for HSM (Hardware Security Module) or PKCS#11


dbeecham:
This is similar to issue https://github.com/slackhq/nebula/issues/51, but not immediately equivalent. It would be neat to be able to put the CA certificate key on a HSM module like the YubiHSM. If `nebula-cert` could have support for PKCS#11, it could store the private key on any PKCS#11 module such as the YubiHSM. Nebula doesn't strictly need to use x509 certificates for this, it could just store the private key pair on the HSM, but if it did use x509, that support would be automatic since we could use openssl/gnutls to generate the certificates on the HSM or on Hashicorp Vault ourselves.

cwichura:
Seconded.  PKCS#11 bindings would support many HSMs.  But for full support, would need (configurable?) hooks to have it call out to log in to a networked HSM cluster, such as Luna SA.

kevin-fibich:
I think this issue should to be extended to a change/clarification of nebulas solution design, since a couple of problems arise from nebulas current approach to certificates (nebula style certificates).

IMHO and from a design perspective, nebula should not deal with any technical details or specific implementations regarding the PKI system. Instead it should define an API signature of what functionality it requires in regards to Authentication/Authorization and make the specific implementation plugable. The current implementation could of course serve as a default module and retain things like the nebula-cert tool but why do we limit nebulas design to that specific implementation? This will only create drag down the line because nebula then needs to chase functionality that already exists elsewhere, be it HSM support, certificate rollout, revocation support etc. If we instead externalise these issues and make the implementations plugable it would become much easier to benefit from the value already created by other solutions and at the same time enable devs to focus on actual nebula functionality. Dropping the whole PKI thing for a PSK or even no auth at all? Or maybe use some next-gen quantum key distribution system? As long as the implementation can satisfy the requirements such an API would present, i believe nebula does not benefit from enforcing a specific one. Some use-cases could even call for multiple implementations at the same time, like using certs like today but getting unsafe_routes via a bgp peering with the lighthouses or some k8s CNI.

The only drawback i see is that this could increase complexity and someone with little experience could set up nebula in a way that is not as secure as that person expects.

dumbasPL:
It's been over 2 years, any progress on this?

johnmaguire:
Hi @dumbasPL - this issue is a feature request, not a tracking issue. I don't know of anyone working on this feature. The best way to show your support is to react to the original issue with a 👍 - this allows maintainers to sort issues by votes.

PR #1153, opened 2024-05-29
add PKCS11 support

JackDoanRivian:
Adds PKCS11 support for the P256 curve. It could probably be generalized for curve 25519 as well, but this is what our HSMs support, and therefore what I could easily test.

I've tested _very thoroughly_ against the difficult-to-obtain Microchip TA100, and did a quick verification with a Yubikey 5C.

Here's some quick examples of how this change can be exercised on a TA100. I'll follow up with Yubikey instructions when time permits:

This change requires building with CGO, so here's the exact invocation we used (adjust as needed for your target architecture of course):

```
CC=aarch64-linux-gnu-gcc CGO_ENABLED=1 GOARCH=arm64 go build -buildvcs=false -ldflags="${LDFLAGS}" -o nebula.arm64.${COMMIT_HASH} ./cmd/nebula
CC=aarch64-linux-gnu-gcc CGO_ENABLED=1 GOARCH=arm64 go build -buildvcs=false -ldflags="${LDFLAGS}" -o nebula-cert.arm64.${COMMIT_HASH} ./cmd/nebula-cert
```
Unfortunately static linking with musl doesn't work, because musl doesn't implement `dlopen` for statically-linked binaries. :disappointed: 

```
./nebula-cert keygen -out-pub test.pem -pkcs11 'pkcs11:slot-id=0;object=device;token=00ABC;type=private?module-path=/tmp/libcryptoauth.so'
```

A config file's `pki` section would look like this:
```
pki:
  ca: /path/to/ca.pem
  cert: /path/to/cert.pem
  key: 'pkcs11:slot-id=0;object=device;token=00ABC;type=private?module-path=/tmp/libcryptoauth.so'
```

Please let me know what you think! I'm very open to making adjustments to make this fit in better with the existing code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,8 +41,7 @@ ALL_LINUX = linux-amd64 \
 	linux-mips-softfloat \
 	linux-riscv64 \
 	linux-loong64 \
-	linux-amd64-pkcs11 \
-	linux-arm64-pkcs11
+	linux-amd64-pkcs11
 
 ALL_FREEBSD = freebsd-amd64 \
 	freebsd-arm64
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,9 @@ ALL_LINUX = linux-amd64 \
 	linux-mips64le \
 	linux-mips-softfloat \
 	linux-riscv64 \
-        linux-loong64
+	linux-loong64 \
+	linux-amd64-pkcs11 \
+	linux-arm64-pkcs11
 
 ALL_FREEBSD = freebsd-amd64 \
 	freebsd-arm64
@@ -116,6 +118,9 @@ bin-freebsd-arm64: build/freebsd-arm64/nebula build/freebsd-arm64/nebula-cert
 bin-boringcrypto: build/linux-$(shell go env GOARCH)-boringcrypto/nebula build/linux-$(shell go env GOARCH)-boringcrypto/nebula-cert
 	mv $? .
 
+bin-pkcs11: build/linux-$(shell go env GOARCH)-pkcs11/nebula build/linux-$(shell go env GOARCH)-pkcs11/nebula-cert
+	mv $? .
+
 bin:
 	go build $(BUILD_ARGS) -ldflags "$(LDFLAGS)" -o ./nebula${NEBULA_CMD_SUFFIX} ${NEBULA_CMD_PATH}
 	go build $(BUILD_ARGS) -ldflags "$(LDFLAGS)" -o ./nebula-cert${NEBULA_CMD_SUFFIX} ./cmd/nebula-cert
@@ -133,6 +138,13 @@ build/linux-mips-softfloat/%: LDFLAGS += -s -w
 # boringcrypto
 build/linux-amd64-boringcrypto/%: GOENV += GOEXPERIMENT=boringcrypto CGO_ENABLED=1
 build/linux-arm64-boringcrypto/%: GOENV += GOEXPERIMENT=boringcrypto CGO_ENABLED=1
+
+# pkcs11
+build/linux-amd64-pkcs11/%: BUILD_ARGS += -tags pkcs11
+build/linux-amd64-pkcs11/%: CGO_ENABLED=1
+build/linux-arm64-pkcs11/%: BUILD_ARGS += -tags pkcs11
+build/linux-arm64-pkcs11/%: CGO_ENABLED=1
+build/linux-arm64-pkcs11/%: GOENV += CC=aarch64-linux-gnu-gcc
 
 build/%/nebula: .FORCE
 	GOOS=$(firstword $(subst -, , $*)) \
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,7 @@ ALL_LINUX = linux-amd64 \
 	linux-mips64le \
 	linux-mips-softfloat \
 	linux-riscv64 \
-	linux-loong64 \
-	linux-amd64-pkcs11
+	linux-loong64
 
 ALL_FREEBSD = freebsd-amd64 \
 	freebsd-arm64
@@ -64,7 +63,7 @@ ALL = $(ALL_LINUX) \
 e2e:
 	$(TEST_ENV) go test -tags=e2e_testing -count=1 $(TEST_FLAGS) ./e2e
 
-e2ev: TEST_FLAGS = -v
+e2ev: TEST_FLAGS += -v
 e2ev: e2e
 
 e2evv: TEST_ENV += TEST_LOGS=1
@@ -97,7 +96,7 @@ release-netbsd: $(ALL_NETBSD:%=build/nebula-%.tar.gz)
 
 release-boringcrypto: build/nebula-linux-$(shell go env GOARCH)-boringcrypto.tar.gz
 
-BUILD_ARGS = -trimpath
+BUILD_ARGS += -trimpath
 
 bin-windows: build/windows-amd64/nebula.exe build/windows-amd64/nebula-cert.exe
 	mv $? .
@@ -117,8 +116,9 @@ bin-freebsd-arm64: build/freebsd-arm64/nebula build/freebsd-arm64/nebula-cert
 bin-boringcrypto: build/linux-$(shell go env GOARCH)-boringcrypto/nebula build/linux-$(shell go env GOARCH)-boringcrypto/nebula-cert
 	mv $? .
 
-bin-pkcs11: build/linux-$(shell go env GOARCH)-pkcs11/nebula build/linux-$(shell go env GOARCH)-pkcs11/nebula-cert
-	mv $? .
+bin-pkcs11: BUILD_ARGS += -tags pkcs11
+bin-pkcs11: CGO_ENABLED = 1
+bin-pkcs11: bin
 
 bin:
 	go build $(BUILD_ARGS) -ldflags "$(LDFLAGS)" -o ./nebula${NEBULA_CMD_SUFFIX} ${NEBULA_CMD_PATH}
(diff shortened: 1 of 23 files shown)


cert/cert.go
```diff
@@ -693,6 +694,9 @@ func (nc *NebulaCertificate) CheckRootConstrains(signer *NebulaCertificate) erro
 
 // VerifyPrivateKey checks that the public key in the Nebula certificate and a supplied private key match
 func (nc *NebulaCertificate) VerifyPrivateKey(curve Curve, key []byte) error {
+	if nc.Pkcs11Backed {
+		return nil //todo!
```
JackDoanRivian, line 698: need to implement this; shouldn't be too tricky

Makefile
```diff
@@ -134,6 +139,13 @@ build/linux-mips-softfloat/%: LDFLAGS += -s -w
 build/linux-amd64-boringcrypto/%: GOENV += GOEXPERIMENT=boringcrypto CGO_ENABLED=1
 build/linux-arm64-boringcrypto/%: GOENV += GOEXPERIMENT=boringcrypto CGO_ENABLED=1
 
+# pkcs11
+build/linux-amd64-pkcs11/%: BUILD_ARGS += -tags pkcs11
+build/linux-amd64-pkcs11/%: CGO_ENABLED=1
+build/linux-arm64-pkcs11/%: BUILD_ARGS += -tags pkcs11
+build/linux-arm64-pkcs11/%: CGO_ENABLED=1
+build/linux-arm64-pkcs11/%: GOENV += CC=aarch64-linux-gnu-gcc
```
JackDoan, line 147: The current test-linux runner doesn't have an aarch64 compiler, which is making runs fail :/

Leaving this comment to track, and in the meantime, removing it from the `all` target

JackDoanRivian, line 147: The current test-linux runner doesn't have an aarch64 compiler, which is making runs fail :/

Leaving this comment to track, and in the meantime, removing it from the `all` target

jasikpark:
Related: https://github.com/slackhq/nebula/issues/328

numinit:
How tricky would it be to add support for `ca` and `sign`?

JackDoanRivian:
@numinit not that bad for `sign`, but I (perhaps incorrectly?) assume that most folks who want a hardware-backed key for their CA are operating Nebula "at scale", and therefore also want semi-custom signing logic, and so therefore they don't want to use the CLI to sign their client certs.

For `ca`, I could add code to generate a CA cert from the key pointed to by a pkcs11 URL, but I don't think it'd be reasonable to add key generation support, since most HSMs (at least the ones I've used) really push you to use their own tooling for creating/managing keys.

numinit:
I think that's basically all I'd expect; just symmetry between the sign, CA and keygen commands :-)

numinit:
@JackDoanRivian this ought to be a good start :-)

https://github.com/numinit/nebula/commit/fbdb4b8491cf12263c0a45c70e96ea637e0211f5

JackDoanRivian:
@numinit this looks great! Is it okay with you if I cherrypick this? You'll have to sign the Salesforce CLA for this to land once I do.

numinit:
Updated so the test suite passes.

https://github.com/numinit/nebula/tree/pkcs11-v1.9.3

numinit:
@JackDoanRivian Go for it, CLA signed. I've been integration testing it and it works great: https://github.com/numinit/nixpkcs

numinit:
I made minor fixes to the existing tests, but probably need to write some more CLI tests.

Regardless, a smoketest of the whole process with multiple Nebula nodes using libsoftokn works great.

https://github.com/numinit/nixpkcs/blob/master/test.nix#L221

numinit:
Running into some trouble with tpm2-pkcs11 using softhsm under QEMU. I'll debug a bit.

> Error: error while getting public key with PKCS#11: asn1: syntax error: sequence truncated

cmd/nebula-cert/sign.go
```diff
@@ -176,19 +186,37 @@ func signCert(args []string, out io.Writer, errOut io.Writer, pr PasswordReader)
...
-		var pubCurve cert.Curve
 		pub, _, pubCurve, err = cert.UnmarshalPublicKey(rawPub)
 		if err != nil {
 			return fmt.Errorf("error while parsing in-pub: %s", err)
 		}
 		if pubCurve != curve {
 			return fmt.Errorf("curve of in-pub does not match ca")
 		}
+	} else if isP11 {
+		pub, err = p11Client.GetPubKey()
```
JackDoanRivian, line 216: this could be a good spot to check that out pkcs11 public key matches the one in the CA-cert we're using

numinit:
One more cherry-pick: https://github.com/numinit/nebula/commit/98dbced19b74f86b4bf4da4c64fe559c40818c07

numinit:
@JackDoanRivian Ack, I missed that comment. Good example that runs in the NixOS test suite is here:

https://github.com/numinit/nixpkcs/blob/master/nixos/tests/nebula.nix#L41

Config specializations (I test this with both nssdb and TPM2) are here:

https://github.com/numinit/nixpkcs/blob/master/flake.nix#L51

Ultimately nixpkcs creates all those declaratively, the imperative commands and environment are buried in that source, though. :-)

johnmaguire:
Finally got some time to test this on macOS with an old school YubiKey 4 (USB A!) and I am able to get it working. A couple notes:

- IIUC, P256 curve is required
- The Makefile currently doesn't build for anything but Linux. I will include a patch at the end of this comment that worked for me. This patch removes it as a distro-specific build target though, and I think that may be important for your use case?

And here's how I configured it:

1. `brew install opensc yubico-piv-tool`
2. `ykman piv reset`
```
WARNING! This will delete all stored PIV data and restore factory settings. Proceed? [y/N]: y
Resetting PIV data...
Reset complete. All PIV data has been cleared from the YubiKey.
Your YubiKey now has the default PIN, PUK and Management Key:
	PIN:	123456
	PUK:	12345678
	Management Key:	010203040506070801020304050607080102030405060708
```

3. `ykman piv keys generate --algorithm ECCP256 9a pubkey.pem`
```
Enter a management key [blank to use default key]:
/opt/homebrew/Cellar/ykman/5.5.1/libexec/lib/python3.12/site-packages/yubikit/piv.py:160: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  return algorithms.TripleDES(management_key)
Private key generated in slot 9A (AUTHENTICATION), public key written to pubkey.pem.
```

4. `pkcs11-tool --module /opt/homebrew/lib/libykcs11.dylib --list-slots`
```
Available slots:
Slot 0 (0x0): Yubico Yubikey 4 OTP+U2F+CCID
  token label        : YubiKey PIV #5599240
  token manufacturer : Yubico (www.yubico.com)
  token model        : YubiKey YK4
  token flags        : login required, rng, token initialized, PIN initialized
  hardware version   : 1.0
  firmware version   : 4.34
  serial num         : 5599240
  pin min/max        : 6/64
```

4. `make bin-pkcs11` (This required a code change to Makefile!)
```
go build -tags cgo,pkcs11 -trimpath -ldflags "-X main.Build=1.9.3-upstream-master-1-g40cfd00-dirty" -o ./nebula "./cmd/nebula"
go build -tags cgo,pkcs11 -trimpath -ldflags "-X main.Build=1.9.3-upstream-master-1-g40cfd00-dirty" -o ./nebula-cert ./cmd/nebula-cert
```

5. Build the URL using url-encoded values from step 4 by hand: `pkcs11:model=YubiKey%20YK4;manufacturer=Yubico%20%28www.yubico.com%29;serial=5599240;token=YubiKey%20PIV%20%235599240;id=%01?module-path=/opt/homebrew/lib/libykcs11.dylib&pin-value=123456`
6. `./nebula-cert keygen -curve P256 -pkcs11 "pkcs11:model=YubiKey%20YK4;manufacturer=Yubico%20%28www.yubico.com%29;serial=5599240;token=YubiKey%20PIV%20%235599240;id=%01?module-path=/opt/homebrew/lib/libykcs11.dylib&pin-value=123456" -out-pub pkcs11.pub`
7.  CA must use the P256 curve: `nebula-cert ca -ips '100.100.0.0/24' -name 'Test CA for PKCS11' -curve P256`
9. Sign the cert: `nebula-cert sign -in-pub pkcs11.pub -ip '100.100.0.1/24' -name pkcs11`
10. Fill out the `pki` section of the config:
 ```
pki:
  ca: /Users/jmaguire/src/nebula/ca.crt
  cert: /Users/jmaguire/src/nebula/pkcs11.crt
  key: pkcs11:model=YubiKey%20YK4;manufacturer=Yubico%20%28www.yubico.com%29;serial=5599240;token=YubiKey%20PIV%20%235599240;id=%01?module-path=/opt/homebrew/lib/libykcs11.dylib&pin-value=123456
```

Here's how I patched the Makefile:

```
diff --git a/Makefile b/Makefile
index 1b75dd5..bba5a4c 100644
--- a/Makefile
+++ b/Makefile
@@ -117,8 +117,9 @@ bin-freebsd-arm64: build/freebsd-arm64/nebula build/freebsd-arm64/nebula-cert
 bin-boringcrypto: build/linux-$(shell go env GOARCH)-boringcrypto/nebula build/linux-$(shell go env GOARCH)-boringcrypto/nebula-cert
        mv $? .

-bin-pkcs11: build/linux-$(shell go env GOARCH)-pkcs11/nebula build/linux-$(shell go env GOARCH)-pkcs11/nebula-cert
-       mv $? .
+bin-pkcs11: BUILD_ARGS += -tags pkcs11
+bin-pkcs11: CGO_ENABLED = 1
+bin-pkcs11: bin

 bin:
        go build $(BUILD_ARGS) -ldflags "$(LDFLAGS)" -o ./nebula${NEBULA_CMD_SUFFIX} ${NEBULA_CMD_PATH}
```

cert/cert.go
JackDoanRivian: this actually isn't needed bc of how pubkey lookup works

johnmaguire (approved):

PR #1162, opened 2024-06-10
cleanup SK_MEMINFO vars

wadey:
We had to manually define these types before, but the latest release of `golang.org/x/sys` adds these definitions:

- https://github.com/golang/sys/commit/6dfb94eaa3bd0fcaa615f58e915f7214ce078beb

Since we just updated with this PR, we can clean this up now:

- https://github.com/slackhq/nebula/pull/1161

### DIFF
--- a/udp/udp_linux.go
+++ b/udp/udp_linux.go
@@ -27,25 +27,6 @@ type StdConn struct {
 	batch int
 }
 
-var x int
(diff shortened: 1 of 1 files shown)


johnmaguire (approved):
Each replacement looks appropriate.

PR #1163, opened 2024-06-13
Fix logging in `cachePacket`

jasikpark:
We were previously logging "Packet store" when the packet is dropped, the counter and logging should agree.


jasikpark:
I was wrong, the debug string is the same whether it was stored or not, and the `stored` variable tracks that status.

PR #1166, opened 2024-06-17
[fix] windows reload should check via

sohunjug:
When using this library for secondary development as a dependency, or when invoking a reload through sshd, if you modify unsafe_route corresponding to via, please delete the old information.


wadey:
Line 26 a at the top of this method says:

```
// Via is ignored since that is only consumed within nebula itself
```

So is that comment incorrect?

sohunjug:
> Line 26 a at the top of this method says:
> 
> ```
> // Via is ignored since that is only consumed within nebula itself
> ```
> 
> So is that comment incorrect?

windows 
```
route print
``` 
when i want change unsafe_route: via, reload will create multi routes with different via

nbrownus:
Digging into the situation here, Windows is treated differently than the other operating systems and I'm not sure there is a good reason for it. The destination for the route is recorded as the `Via` instead of the nebula tun device which leads to duplicated routes when the `Via` for an `unsafe_route` changes and `nebula` is reloaded.

The ideal fix here would be have windows install the route destination as the nebula tun device instead, which would match the other OSs behavior in silently erroring on installing a duplicate route, unless Windows is happy to add the duplicate route, in which case I would prefer we add a specific check for Windows only on equating `Via`.

In the end I suspect this is largely a vanity issue since `nebula` still needs to internally determine the unsafe router for a packet outside its own vpn networks.

PR #1173, opened 2024-07-05
WIP: switch most everything to netip in prep for ipv6 in the overlay

nbrownus:
Currently a work in progress


calculated_remote.go
```diff
@@ -1,63 +1,79 @@
...
+	b = ip.As4()
+	intIp := binary.BigEndian.Uint32(b[:])
+
+	return &Ip4AndPort{(maskIp & mask) | (intIp & ^mask), c.port}
+}
 
-	return &Ip4AndPort{Ip: uint32(masked), Port: c.port}
+func (c *calculatedRemote) apply6(ip netip.Addr) *Ip4AndPort {
+	//TODO: IPV6-WORK
+	panic("AAHHHHHH")
```
IanVS, line 66: image

control.go
```diff
@@ -132,7 +130,7 @@ func (c *Control) ListHostmapIndexes(pendingMap bool) []ControlHostInfo {
 }
 
 // GetHostInfoByVpnIp returns a single tunnels hostInfo, or nil if not found
-func (c *Control) GetHostInfoByVpnIp(vpnIp iputil.VpnIp, pending bool) *ControlHostInfo {
+func (c *Control) GetHostInfoByVpnIp(vpnIp netip.Addr, pending bool) *ControlHostInfo {
```
nbrownus, line 133: Should we unmap for the caller here or expect they have unmapped the addr?

hostmap.go
```diff
@@ -133,19 +132,19 @@ func (rs *RelayState) CopyRelayForIdxs() []uint32 {
...
 }
 
-func (rs *RelayState) RemoveRelay(localIdx uint32) (iputil.VpnIp, bool) {
+func (rs *RelayState) RemoveRelay(localIdx uint32) (netip.Addr, bool) {
 	rs.Lock()
 	defer rs.Unlock()
 	r, ok := rs.relayForByIdx[localIdx]
 	if !ok {
-		return iputil.VpnIp(0), false
+		return netip.Addr{}, false
```
nbrownus, line 140: Do we want to return unspecified here or invalid?

iputil/util.go
```diff
@@ -1,93 +0,0 @@
-package iputil
-
-import (
-	"encoding/binary"
-	"fmt"
-	"net"
-	"net/netip"
-)
-
-type VpnIp uint32
```
JackDoanRivian, line 10: leaving a note to consider bringing this back as `type VpnIp netip.addr` mostly as documentation, so we don't forget

allow_list_test.go
```diff
@@ -143,3 +143,19 @@ func TestLocalAllowList_AllowName(t *testing.T) {
 	assert.Equal(t, true, al.AllowName("eth0"))
 	assert.Equal(t, true, al.AllowName("ens5"))
 }
+
+func string2nip(s string) netip.Addr {
+	ip, err := netip.ParseAddr(s)
```
JackDoanRivian, line 148: probably doesn't matter, but netip.MustParseAddr exists and does this for you

control.go
JackDoanRivian: as best as I can tell, this is only called in test code, so I think we can err on the side of calling `Unmap` internally here.

dns_server.go
```diff
@@ -80,7 +82,11 @@ func parseQuery(l *logrus.Logger, m *dns.Msg, w dns.ResponseWriter) {
 			}
 		case dns.TypeTXT:
 			a, _, _ := net.SplitHostPort(w.RemoteAddr().String())
-			b := net.ParseIP(a)
+			b, err := netip.ParseAddr(a)
+			if err != nil {
+				return
```
JackDoanRivian, line 87: consider l.Debugf?

e2e/router/router.go
```diff
@@ -697,41 +685,25 @@ func (r *R) FlushAll() {
...
-		port, err := strconv.Atoi(sPort)
-		if err != nil {
-			panic(err)
-		}
-
-		r.outNat[c.GetUDPAddr()+":"+fromAddr] = net.UDPAddr{
-			IP:   net.ParseIP(sHost),
-			Port: port,
-		}
+		r.outNat[c.GetUDPAddr().String()+":"+fromAddr.String()] = toAddr
```
JackDoanRivian, line 706: idk if you wanna mess with it, but I think we have the opportunity to use an AddrPort over a string here

nbrownus: This is an internal NAT for the e2e tests, both sides are `netip.AddrPort`s. Could move this to a struct containing two `netip.AddrPort`s to avoid the strings

hostmap.go
JackDoanRivian: The default/empty value makes sense imo -- also, this function isn't ever called :D

udp/udp_rio_windows.go
```diff
@@ -295,17 +298,15 @@ func (u *RIOConn) WriteTo(buf []byte, addr *Addr) error {
...
-func (u *RIOConn) LocalAddr() (*Addr, error) {
+func (u *RIOConn) LocalAddr() (netip.AddrPort, error) {
 	sa, err := windows.Getsockname(u.sock)
 	if err != nil {
-		return nil, err
+		return netip.AddrPort{}, err
 	}
 
 	v6 := sa.(*windows.SockaddrInet6)
-	return &Addr{
```
JackDoanRivian, line 305: does golang have a `htons` equivalent?

control_test.go
```diff
@@ -91,27 +98,29 @@ func TestControl_GetHostInfoByVpnIp(t *testing.T) {
...
+		CurrentRemote:          remote1,
+		CurrentRelaysToMe:      []netip.Addr{},
+		CurrentRelaysThroughMe: []netip.Addr{},
 	}
 
 	// Make sure we don't have any unexpected fields
 	assertFields(t, []string{"VpnIp", "LocalIndex", "RemoteIndex", "RemoteAddrs", "Cert", "MessageCounter", "CurrentRemote", "CurrentRelaysToMe", "CurrentRelaysThroughMe"}, thi)
-	test.AssertDeepCopyEqual(t, &expectedInfo, thi)
+	assert.EqualValues(t, &expectedInfo, thi)
+	//TODO: netip.Addr reuses global memory for zone identifiers which breaks our "no reused memory check" here
```
nbrownus, line 118: Not sure what to do with this one

hostmap.go
```diff
@@ -666,13 +665,21 @@ func (i *HostInfo) CreateRemoteCIDR(c *cert.NebulaCertificate) {
 		return
 	}
 
-	remoteCidr := cidr.NewTree4[struct{}]()
+	remoteCidr := new(bart.Table[struct{}])
 	for _, ip := range c.Details.Ips {
-		remoteCidr.AddCIDR(&net.IPNet{IP: ip.IP, Mask: net.IPMask{255, 255, 255, 255}}, struct{}{})
+		//TODO: IPV6-WORK what to do when ip is invalid?
```
nbrownus, line 670: Need to determine failure strategy

control_test.go
JackDoanRivian: Can we write a function to check everything but the zone?

Or, for the purpose of testing, we could (somehow?) ensure that each netip.Addr we check like this has a unique zone, but that sort of sucks

lighthouse.go
```diff
@@ -182,25 +172,40 @@ func (lh *LightHouse) GetUpdateInterval() int64 {
...
+			if err != nil {
+				return util.NewContextualError("Unable to parse port in lighthouse.advertise_addrs entry", m{"addr": rawAddr, "entry": i + 1}, err)
+			}
+
+			if port == 0 {
+				port = int(lh.nebulaPort)
 			}
 
-			if ip4 := fIp.To4(); ip4 != nil && lh.myVpnNet.Contains(fIp) {
+			//TODO: we could technically insert all returned ips instead of just the first one if a dns lookup was used
```
nbrownus, line 200: Do we want to include this in the change here?

udp/udp_linux.go
```diff
@@ -146,12 +146,23 @@ func (u *StdConn) ListenOut(r EncReader, lhf LightHouseHandlerFunc, cache *firew
 		//metric.Update(int64(n))
 		for i := 0; i < n; i++ {
 			if u.isV4 {
-				udpAddr.IP = names[i][4:8]
+				ip, _ = netip.AddrFromSlice(names[i][4:8])
+				//TODO: IPV6-WORK what is not ok?
```
nbrownus, line 150: Need to determine what to do here

iputil/util.go
wadey: or at least `type VpnIp=netip.addr`

udp/udp_rio_windows.go
JackDoanRivian: does not, lgtm

wadey (approved):

gaissmai:
>     * `bart` appears to be pretty slow when compared with our old `cidr.Tree`. We should look at performance carefully and determine if we need to resurrect that code.

huch, really? I'm the author of `bart` and I'm pretty sure `bart` is one of the fastest go libraries for IPv4/IPv6 longest-prefix-match operations. Do you have benchmarks?

PR #1176, opened 2024-07-10
Ipv6 overlay 

### LINKED ISSUE #6: IPv6 for overlay network?
Is it possible to use IPv6 for the internal overlay network? Our IPv4 RFC1918 allocation strategy has been...not good.

nbrownus:
This is a very early proof of concept. It is 100% **not** backward compatible with released nebula versions in its current form

Moving over to https://github.com/nbrownus/nebula/tree/ipv6-overlay-final for the time being

Closes #6


cert/cert.proto
```diff
@@ -19,8 +19,8 @@ message RawNebulaCertificateDetails {
     string Name = 1;
 
     // Ips and Subnets are in big endian 32 bit pairs, 1st the ip, 2nd the mask
-    repeated uint32 Ips = 2;
-    repeated uint32 Subnets = 3;
+    RawNetwork Ip = 2;
+    repeated RawNetwork Subnets = 3;
```
wadey, line 23: For a compatible version, I think we can just deprecate the old versions of these fields and make the new Replacement be optional. Then we could have the code prefer the new fields but fall back to the old ones if missing.

JackDoanRivian: I like this approach as well

handshake_ix.go
```diff
@@ -99,20 +99,7 @@ func ixHandshakeStage1(f *Interface, addr netip.AddrPort, via *ViaSender, packet
...
-		if f.l.Level > logrus.DebugLevel {
-			e = e.WithField("cert", remoteCert)
-		}
-
-		e.Info("Invalid vpn ip from host")
-		return
-	}
-
-	vpnIp = vpnIp.Unmap()
+	vpnIp := remoteCert.Details.Ip.Addr()
```
JackDoanRivian, line 102: If  `Ip` is somehow blank/invalid, do we need a check for that here to mirror the old way?

outside.go
```diff
@@ -303,22 +305,99 @@ func (f *Interface) handleEncrypted(ci *ConnectionState, addr netip.AddrPort, h
...
+	if incoming {
+		fp.RemoteIP, _ = netip.AddrFromSlice(data[8:24])
+		fp.LocalIP, _ = netip.AddrFromSlice(data[24:40])
+	} else {
+		fp.LocalIP, _ = netip.AddrFromSlice(data[8:24])
+		fp.RemoteIP, _ = netip.AddrFromSlice(data[24:40])
+	}
+
+	//TODO: whats a reasonable number of extension headers to attempt to parse?
+	//https://www.ietf.org/archive/id/draft-ietf-6man-eh-limits-00.html
```
JackDoanRivian, line 337: If I'm reading this correctly, the last bullet of 3.2.2 says we can set a limit of 104 bytes for the "maximum length of the IPv6 header chain". Perhaps that's the way to go, until we observe problems?

JackDoanRivian: Cisco has a nice paper here that explains what they do with extension headers: https://www.cisco.com/en/US/technologies/tk648/tk872/technologies_white_paper0900aecd8054d37d.html

JackDoanRivian: https://datatracker.ietf.org/doc/html/rfc7872#page-3 This RFC measured the drop rate of IPv6 packets with _any_ extension headers circa 2016, with dismal results. It seems that attempting to deal with them at all is considered noble.

outside.go
```diff
@@ -303,22 +305,99 @@ func (f *Interface) handleEncrypted(ci *ConnectionState, addr netip.AddrPort, h
...
+				return fmt.Errorf("ipv6 packet was too small")
+			}
+			fp.Protocol = uint8(proto)
+			fp.RemotePort = binary.BigEndian.Uint16(data[offset : offset+2])
+			fp.LocalPort = binary.BigEndian.Uint16(data[offset+2 : offset+4])
+			fp.Fragment = false
+			return nil
+
+		case layers.IPProtocolIPv6Fragment:
+			//TODO: can we determine the protocol?
```
JackDoanRivian, line 371: Kind of? Subsequent fragment packets don't contain their TCP/UDP headers. We can keep a list of all the fragment Identification Numbers with their protocols and port numbers, but that could get out of hand quickly.

outside.go
```diff
@@ -303,22 +305,99 @@ func (f *Interface) handleEncrypted(ci *ConnectionState, addr netip.AddrPort, h
...
+	}
+
+	//TODO: whats a reasonable number of extension headers to attempt to parse?
+	//https://www.ietf.org/archive/id/draft-ietf-6man-eh-limits-00.html
+	proto := layers.IPProtocol(data[6])
+	offset := 40
+	for i := 0; i < 24; i++ {
+		switch proto {
+		case layers.IPProtocolICMPv6:
+			//TODO: we need a new protocol in config language "icmpv6"
```
JackDoanRivian, line 343: We may also want to think about dropping NDP and perhaps also RS/RA traffic, but allowing things like ping and error messages. NDP in particular should never occur on a nebula network (I think?)

outside.go
JackDoanRivian: Because ipv6 only allows the host to fragment, support for fragments at all could be less critical.

Another ipv6-specific problem is because of extension headers, I think it's possible for port numbers to be missing from the first packet of a fragmented message :(

cert/cert.proto
```diff
@@ -35,6 +35,12 @@ message RawNebulaCertificateDetails {
     Curve curve = 100;
 }
 
+message RawNetwork {
```
JackDoanRivian, line 38: Because submessages are LEN types, it could make sense to use a string here, which would let us use netip.Addr.MarshalBinary.

I'm gonna ignore `bits` for now because it needs to be encoded somehow anyway

A submessage:
```
[LEN+id] [length] [ [hi id][hi varint] [lo id][lo varint] ] = 
min 1 + 1 + 2(1+1)  = 6, max 1 + 1 + 2(1+10) = 24
```

A `bytes` for address:
```
[LEN+id] [length] [16-byte address]  = 
1 + 1 + 16 = 18
```

A submessage can make sense if we think hi/lo of the v6 address are likely to be "small numbers", but if they're not, we could end up wasting bytes compared to a string-ier representation.

nebula.proto
```diff
@@ -23,13 +23,18 @@ message NebulaMeta {
...
-  uint32 VpnIp = 1;
+  Ip VpnIp = 1;
   repeated Ip4AndPort Ip4AndPorts = 2;
   repeated Ip6AndPort Ip6AndPorts = 4;
-  repeated uint32 RelayVpnIp = 5;
+  repeated Ip RelayVpnIp = 5;
   uint32 counter = 3;
 }
 
+message Ip {
```
JackDoanRivian, line 33: same comment as above, "what if we used `bytes`?"

cert/cert.proto
nbrownus: Current goals are to get the innards of ipv6 working, figure backwards/forwards compatibility will take the most time so I am ignoring it for now.

cert/cert.proto
nbrownus: I dont anticipate this to be the final types, `fixed64 Hi` and `uint64 Lo` is probably correct though. I agree avoiding the submessage is ideal as well. Avoiding protobuf would also be pretty cool :D

nebula.proto
nbrownus: I dislike the unbounded nature of `bytes`. 

As mentioned in the other comment, I had noodled on something along these lines
```
message NebulaMetaDetails {
  fixed64 VpnIpHi = 1;
  uint64 VpnIpLo = 2;
  ...
}
```

Stuff to noodle on at least

IanVS:
Don't worry, It is still being actively developed.  This was simply an early proof-of-concept PR that was used to get some feedback from other team members.

Issue #915, opened 2023-07-11
Support socks5 inbound


aa51513:
Nebula is a pretty good tool, I just can't put it down

But recently I encountered a problem: I want to run nebula on a machine without root permissions and network privileges, and I failed

I try to choose not to create a tun interface, nebula can run normally, but I can't do anything except ping the nebula network IP

I would like to ask whether it can support socks5 inbound, and then pass the traffic through nebula's virtual Lan to reach the destination

The data flow is probably like this：
`visitor—>socks5_client—>nebula(socks5_server)—>another_nebula_node`

I noticed that there is already a mature socks5 server implemented by golang on github, which fully supports tcp and udp
[socks5_server_by_golang](https://github.com/wzshiming/socks5)

nbrownus:
Closing in favor of #1014

PR #1179, opened 2024-07-14
Support UDP/TCP port fowarding to a host without setting up a tun

### LINKED ISSUE #1014: Feature request: Support UDP/TCP port fowarding to a host without setting up a tun
Allow binding another Nebula node's port as a local port on the machine. This would allow tun-less access to that host/port combo.

### Discussed in https://github.com/slackhq/nebula/discussions/1011

<div type='discussions-op-text'>

<sup>Originally posted by **nick008a** November 12, 2023</sup>
Feature request : support UDP/TCP port fowarding without setting up a tun

This will allow nebula to work without root, and one could further set up a socks proxy with this to enable #915</div>

<div type='discussions-op-text'>

<sup>Originally posted by **nick008a** November 12, 2023</sup>

i was looking for something like ngrok which allows you tunnel a port to a remote machine (a local machine can access 127.0.0.1:8080 to talk to, say, 127.0.0.1:8081 on a remote machine without setting up a tun)</div>

cre4ture:
This implements #1014.
It also allows to use remote SOCKS proxy (or `ssh` server) to enable the forwarding of more than a single port at a time.
This would solve the needs for #915.

Build-Status twin PR: https://github.com/cre4ture/nebula/pull/1

How to be used can be read in the example `config.yml`:

```yml
# by using port port forwarding (port tunnels) its possible to establish connections
# from/into the nebula-network without using a tun/tap device and thus without requiring root access
# on the host. port forwarding is only supported when setting "tun.user" is set to true and thus
# a user-tun instead of a real one is used.
# IMPORTANT: For incoming tunnels, don't forget to also open the firewall for the relevant ports.
port_forwarding:
  outbound:
  # format of local and remote address: <host/ip>:<port>
  #- local_address: 127.0.0.1:3399
  #  remote_address: 192.168.100.92:4499
     # format of protocols lists (yml-list): [tcp], [udp], [tcp, udp]
  #  protocols: [tcp, udp]
  inbound:
  # format of forward_address: <host/ip>:<port>
  #- port: 5599
  #  forward_address: 127.0.0.1:5599
     # format of protocols lists (yml-list): [tcp], [udp], [tcp, udp]
  #  protocols: [tcp, udp]

```

~Till now I only did some basic manual tests with netcat~. They resulted in the expected behavior.
Further tests where done by @sybrensa and also by myself.
- [x] outgoing TCP tested intensively
- [x] ~usafe_routes~ FAILS, but potential fix for the user-tun available here: https://github.com/cre4ture/nebula/tree/feature/enable_unsafe_routes_for_user_tun
- [x] incoming TCP tested intensively
- [x] incoming UDP tested intensively (I used iperf3. Datarate was reduced, but works)
- [x] outgoing UDP tested intensively
- more performance tuning (different PR I guess)

~These points are still open:~
- [x] namings like "tunnel" vs. "forwarding", "ingoing" vs. "inbound", ...  Please help me in finding the right terminology as I'm not an expert in this topic.
- [x] mutexes needed? I honestly did't yet explicitly check for potential concurrent accesses. this is ~a TODO~ DONE.
- [x] clean shutdown - Currently it gives some errors when shutting down. this is ~also TODO~ DONE.
- [x] memory leaks - ~I think~ DONE.
- [x] performance tunings - DONE, but could be more
- [x] 2nd iteration of performance tunings - DONE, but could be more
- [x] reloadable (SIGHUP) port forwarding configuration

I did file copy tests. ~What I got was 1/2 the rate and around 4 times more CPU usage compared to the case with kernel-tun.
I tried multiple different improvements on my code, but I fear its actually more a limitation of the gvisor netstack :-/~
Any help or idea is welcome.

Update: My file copy tests achieve now around 90% of the speed at 160% of CPU load of nebula. This is a significant improvment and acceptable (at least for my useases). Thanks @akernet for the support here.


akernet:
Thanks a ton for taking this project on!

I took a stab at this before finding your changes here, it's [here](https://github.com/slackhq/nebula/compare/master...akernet:nebula:push-ynqokntmuslt) if you want to take a look.

In particular [e764eb](https://github.com/slackhq/nebula/commit/e764eb4f5cfe46b8375baa820d6fc9726a58b032) adds a script that sets up a two device loopback tunnel (without root) and runs an iperf3 speed test.

I was able to improve throughput by ~7x by adding some buffering to the `UserDevice` ([a16fdb](https://github.com/slackhq/nebula/commit/a16fdb5b0792a19e24ba504fb8a7dc8f585a0dc0)) which you might want to try too. With this change it is running at 1.6 Gbit/s with a 1300 MTU which makes it quite usable to me at least. 

I still think there are several improvements left on the table:
 - Re-use allocations in this buffering by using something like `sync.Pool`, 
 - Or even better, avoid these copies all together, although that would require some bigger changes in the rest of Nebula.
 - Depending on how the internal locking of gvisor works you could probably gain some by running several net stacks in parallel that handle their individual set of connections (distributing packets based on their source address+destination port). This would not improve single socket connections though.

cre4ture:
> Thanks a ton for taking this project on!
> 
> I took a stab at this before finding your changes here, it's [here](https://github.com/slackhq/nebula/compare/master...akernet:nebula:push-ynqokntmuslt) if you want to take a look.
> 
> In particular [e764eb](https://github.com/slackhq/nebula/commit/e764eb4f5cfe46b8375baa820d6fc9726a58b032) adds a script that sets up a two device loopback tunnel (without root) and runs an iperf3 speed test.
> 
> I was able to improve throughput by ~7x by adding some buffering to the `UserDevice` ([a16fdb](https://github.com/slackhq/nebula/commit/a16fdb5b0792a19e24ba504fb8a7dc8f585a0dc0)) which you might want to try too. With this change it is running at 1.6 Gbit/s with a 1300 MTU which makes it quite usable to me at least.
> 
> I still think there are several improvements left on the table:
> 
>     * Re-use allocations in this buffering by using something like `sync.Pool`,
> 
>     * Or even better, avoid these copies all together, although that would require some bigger changes in the rest of Nebula.
> 
>     * Depending on how the internal locking of gvisor works you could probably gain some by running several net stacks in parallel that handle their individual set of connections (distributing packets based on their source address+destination port). This would not improve single socket connections though.

Hey, cool. I could use some support especially regarding the performance and the automated testing. But also for other golang specifics as I'm rather unexperienced there.

I tried to do some performance profiling with the help of `pprof`.
This results where leading me to problems in gvisor itself, as I can read it from this diagram:
grafik

It seems that its doing some dynamic memory allocation. But from a first glance into the relevant source-code I could not see the reason for it. Do you have an idea? I'm actually already failing to upgrade the gvisior to the latest version. something is wrong with it such that it complains about multiple packages in a directrory. But how can this be undiscovered by the maintainters of gvisor? I have the feeling that I'm doing something wrong... :-/

I will for sure try your idea with the buffered queues. 
I didn't yet look in detail into your branch. But it seems that we had some similar approach using the gvisor stack.
I would be glad, if you can do a review of my code. It seems that the maintainers of the repo where so far bussy with other stuff. ;-)

johnmaguire:
Hi @cre4ture - this PR looks very interesting! I wanted to let you know that the maintainers best suited to review this PR are currently working on #6 which requires some major rework of Nebula, so it will probably be a bit before we are able to dig in. That being said, thanks for the contribution!

cre4ture:
@akernet I added your test and afterwards also the performance improvement via cherry-picking. Hope this is fine for you :-)

Your performance improvement lead to a improvment of x6 for the test in your testscript-commit on my hardware. I will do a further test where I will use this on a real-world example (the file-copy test with two seperate machines) to confirm this improvment. I'm a bit concerned because the pprof profiling directed my in some different direct. At least this is how I interpret it right now. I will net you know.

cre4ture:
@akernet 

it seems to me that the improvement with the buffered pipes has no impact on the my testing scenario with the file-copy between two different machines. I'm using a "private cloud" server storage called ["garage"](https://garagehq.deuxfleurs.fr/). The two machines are my laptop and a NUC PC ("server"). These are my results:

```
1. test-case: official released binary - kernel tun:
95.5 MB/second
CPU load on server side:
88 % nebula
71 % garage

2. test-case: local build with buffered pipes - gvisor tun+stack:
52.8 MB/second
CPU load server side:
197 % nebula
34 % garage

2. test-case: local build no buffered pipes - gvisor tun+stack:
52.3 MB/second
CPU load server side:
195 % nebula
35 % garage

3. test-case: local build with the buffered pipes - but kernel tun (as in 1. test-case):
99.5 MB/second
94% nebula
72% garage
```

The last test shall demonstrate the the locally compiled binary has comparable performance as the release binary.
So what exactly this means is not yet clear to me. It seems that the buffered pipes do not have a significant impact for this test-scenarios. But I can't reason about it.

akernet:
> I would be glad, if you can do a review of my code. It seems that the maintainers of the repo where so far bussy with other stuff. ;-)

I'm also new to go but I'll try to find some time in the next days! :)

>I added your test and afterwards also the performance improvement via cherry-picking. Hope this is fine for you :-)

Ofc!

>I'm using a "private cloud" server storage called ["garage"](https://garagehq.deuxfleurs.fr/)

Cool, I've been looking at garage for the past weeks so nice to see that you are using it with Nebula!

>It seems that the buffered pipes do not have a significant impact for this test-scenarios. But I can't reason about it.

Yeah this is a bit strange. How is the system load overall, is it close to max? The buffering should not make Nebula more efficient when it comes to CPU, in fact it's probably gonna be more costly due to the copying. What it does is decoupling the outside and inside (gvisor) parts, allowing them to run in parallel. If the system is at limits already this is unlikely to help, however. I'll try to do some benchmarking in CPU limited scenarios too, gvisor is always gonna use some additional resources but the 66% of your pprof run seems a bit high.

cre4ture:
Update regarding performance:
Yesterday I could achieve a significant improvement for my filecopy test-scenario:
```
"performance 5":

89.8 MB/second
cpu-load server side:
163 % CPU nebula
63 % CPU garage
```

This brings the copy-speed to almost the reference (99MB/second) - only 10% difference remaining.
The CPU-load of nebula is only + ~60% now.

So overall we are at around 170% difference. Which is significant compared to the initial ~ 400%.

@akernet I think the main difference comes from a further improvement of your performance tuning "buffering to UserDevice". I achieved it by addtionally avoiding data-copy and dynamic allocation steps (i think).

port-forwarder/fwd_tcp.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (cf ForwardConfigOutgoingTcp) SetupPortForwarding(
+	tunService *service.Service,
+	l *logrus.Logger,
+) (io.Closer, error) {
+	localTcpListenAddr, err := net.ResolveTCPAddr("tcp", cf.localListen)
+	if err != nil {
+		return nil, err
+	}
+	remoteTcpAddr, err := net.ResolveTCPAddr("tcp", cf.remoteConnect)
```
akernet, line 51: I would recommend that you do this resolution when needed. For IPs it doesn't really matter, but host lookups might change after starting Nebula.

Another thing that I don't think we have to cover in this PR is looking up IPs from the Nebula network instead of using DNS

port-forwarder/fwd_tcp.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (pt *PortForwardingOutgoingTcp) handleClientConnection(localConnection *net.TCPConn) {
+	err := pt.handleClientConnection_intern(localConnection)
+	if err != nil {
+		pt.l.Debugf("Closed TCP client connection %s. Err: %v",
+			localConnection.LocalAddr().String(), err)
+	}
+}
+
+func (pt *PortForwardingOutgoingTcp) handleClientConnection_intern(localConnection net.Conn) error {
```
akernet, line 98: Could we remove this function? Looks a bit redundant

port-forwarder/fwd_tcp.go
```diff
@@ -0,0 +1,230 @@
...
+
+	dataTransferHandler := func(from, to net.Conn) error {
+
+		name := fmt.Sprintf("%s -> %s", from.LocalAddr().String(), to.LocalAddr().String())
+
+		defer from.Close()
+		defer to.Close()
+		defer cancel()
+		// give communication in opposite direction time to finish as well
+		defer time.Sleep(time.Millisecond * 300)
```
akernet, line 119: I think you want to avoid sleeps like this one if possible

port-forwarder/fwd_tcp.go
```diff
@@ -0,0 +1,230 @@
...
+		from.SetDeadline(time.Time{})
+		buf := make([]byte, 1*(1<<20))
+		for {
+			select {
+			case <-dataTransferCtx.Done():
+				return nil
+			default:
+			}
+
+			rn, r_err := from.Read(buf)
```
akernet, line 132: I wonder if this could be replaced with `io.Copy()`

port-forwarder/fwd_tcp.go
```diff
@@ -0,0 +1,230 @@
...
+
+type ForwardConfigIncomingTcp struct {
+	ForwardConfigIncoming
+}
+
+func (cfg ForwardConfigIncomingTcp) ConfigDescriptor() string {
+	return fmt.Sprintf("inbound.tcp.%d.%s", cfg.port, cfg.forwardLocalAddress)
+}
+
+type PortForwardingOutgoingTcp struct {
```
akernet, line 31: Suggestion: setup a `context.WithCancel()` here that you then pass down to established sockets (creating a new derived context for all of them), that way live connections will be automatically canceled at shutdown or config changes

port-forwarder/fwd_tcp.go
```diff
@@ -0,0 +1,230 @@
...
+		// give communication in opposite direction time to finish as well
+		defer time.Sleep(time.Millisecond * 300)
+
+		// no write/read timeout
+		to.SetDeadline(time.Time{})
+		from.SetDeadline(time.Time{})
+		buf := make([]byte, 1*(1<<20))
+		for {
+			select {
+			case <-dataTransferCtx.Done():
```
akernet, line 127: I think you can handle this in a separate goroutine and call `from.Close()` and `to.Cancel()`

akernet (commented):
Really nice! Cool that you got the buffer reuse to work

examples/config.yml
```diff
@@ -206,6 +206,10 @@ relay:
 
 # Configure the private interface. Note: addr is baked into the nebula certificate
 tun:
+  # When tun is set to "user" then a userspace tun is used instead of an official kernel supported one.
+  # This way nebula can be used without root privileges and offers port forwarding from and into
+  # the nebula network. "tun.disabled" needs to be false for this use-case.
+  user: false
```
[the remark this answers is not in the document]
cre4ture: This was the quickest way to make it run. But its true, its not very intuitive. I will give it a try. Thanks

examples/config.yml
```diff
@@ -368,3 +372,22 @@ firewall:
...
+port_forwarding:
+  outbound:
+  # format of local and remote address: <host/ip>:<port>
+  #- local_address: 127.0.0.1:3399
+  #  remote_address: 192.168.100.92:4499
+     # format of protocols lists (yml-list): [tcp], [udp], [tcp, udp]
+  #  protocols: [tcp, udp]
+  inbound:
+  # format of forward_address: <host/ip>:<port>
+  #- port: 5599
```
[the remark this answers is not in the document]
cre4ture: Will it ever be usefull to provide an IP-address additionally to the port here? I'm not an expert, but from what I understood, the only relevant ip-address to specify here is the own nebula-ip. This was the reason to reduce this setting just to the port. I would suggest to call it `listen_port`. Maybe this way its best?

port-forwarder/fwd_tcp.go
cre4ture: Ah, I see. Your right. I never used it, but nebula has an experimental DNS, right?

port-forwarder/fwd_tcp.go
cre4ture: I agree, especially in other languages like C++ where the default is synchronous execution. With golang I wasn't so sure whether its a real problem as the goroutines seems to compare to the async sleeps of other languages.

I honestly do not know it this is actually really needed at all. Because this is intented to catch the case where the receiving side is closed before the sending side could send all requested data. I do not have a testing possibility for this. Do you know if this is needed, or how to test it ideally?

port-forwarder/fwd_tcp.go
cre4ture: it can. I once tried this. But it didn't improve the performance. Thats why I went on with other experiments. But now, as we solved the issue to a big extent, I will introduce it again as it simplifies the code. Thanks for pointing it out.

cre4ture: I added a variant with io.Copy().
Problem: It seems to slightly decrease performance. So I'm wondering if I better should keep the original implementation.

port-forwarder/fwd_tcp.go
cre4ture: Thanks for pointing out. I realized that it is actually even redundant. Thus I completely removed it.

port-forwarder/fwd_tcp.go
cre4ture: I like to keep it. There is a tiny but significant advantage of keeping it: The logging for the error case needs to be implemented only once instead of twice when merging the function into the caller function. I renamed the function to make this a bit more clear.

cmd/nebula/main.go
```diff
@@ -52,16 +59,61 @@ func main() {
 		os.Exit(1)
 	}
 
-	ctrl, err := nebula.Main(c, *configTest, Build, l, nil)
-	if err != nil {
-		util.LogWithContextIfNeeded("Failed to start", err, l)
-		os.Exit(1)
-	}
+	userspace_tun := c.GetBool("tun.user", false)
+	if userspace_tun {
```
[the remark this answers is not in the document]
cre4ture: I had a quick look. But this is not easily possible. We instanciate the `service`  class a few lines below with service.New(). Inside of this, the `nebula.Main()` is called. So we would need to refactor the instaciation of the `service`. I currently don't see the big value to justify this.

examples/config.yml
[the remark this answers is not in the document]
cre4ture: done :-)

examples/config.yml
```diff
@@ -368,3 +372,22 @@ firewall:
...
+
+# by using port port forwarding (port tunnels) its possible to establish connections
+# from/into the nebula-network without using a tun/tap device and thus without requiring root access
+# on the host. port forwarding is only supported when setting "tun.user" is set to true and thus
+# a user-tun instead of a real one is used.
+# IMPORTANT: For incoming tunnels, don't forget to also open the firewall for the relevant ports.
+port_forwarding:
+  outbound:
+  # format of local and remote address: <host/ip>:<port>
+  #- local_address: 127.0.0.1:3399
```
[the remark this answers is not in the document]
cre4ture: changed accordingly. thanks

examples/config.yml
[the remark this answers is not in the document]
cre4ture: changed to `listen_port`. thanks

port-forwarder/fwd_tcp.go
cre4ture: done. thanks

port-forwarder/fwd_tcp.go
cre4ture: I removed it.

port-forwarder/fwd_tcp.go
cre4ture: thanks. done. I also added a test for this specific feature.

cre4ture:
@akernet please re-review. and it seems that you need to also sign the cla as I cherry picked from your branch to honor your contribution. :-)

examples/config.yml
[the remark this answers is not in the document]
akernet: I'm not sure we want to support it here but I believe that you can have more than one IP assigned to a Nebula certificate/host so it kinda makes sense to leave it configurable.

cre4ture:
@johnmaguire Can you approve another CI run on this branch? It should be green accroding to the run in my fork: https://github.com/cre4ture/nebula/pull/1

cre4ture:
@johnmaguire thanks for running the CI. It has detected a instability that I do not fully understand. It seems that the tests get messed up due to using the same port  numers. But I explicitly made each test using its own port number. Do you know if the CI runns the same test multiple times in parallel? Otherwise I can't really explaint this.
For now I added some random number generator for additionally getting differences in the used port numbers. But its clear that this just reduces the chance for an issue and not completely solves it.

Anyway, could yoe please run the CI another time?
Thanks.

johnmaguire:
Hi @cre4ture - I don't believe that CI runs the same test in parallel multiple times. There is a test matrix which runs the tests on various OS. Looking at the failed CI, it appears it succeeded on windows-latest, macos-latest, linux with boringcrypto, but failed on ubuntu-linux. Oftentimes this is indicative of some kind of intermittent test failure (i.e. flaky test, maybe a race condition.)

> It seems that the tests get messed up due to using the same port numers.

I admittedly only took a very quick look, but I didn't find a log message that suggested a requested port was in use. Perhaps I missed it. Can you share what you're looking at?

cre4ture:
> Hi @cre4ture - I don't believe that CI runs the same test in parallel multiple times. There is a test matrix which runs the tests on various OS in individual containers - so ports should not conflict. Looking at the failed CI, it appears it succeeded on windows-latest, macos-latest, linux with boringcrypto, but failed on ubuntu-linux. Oftentimes this is indicative of some kind of intermittent test failure (i.e. flaky test, maybe a race condition.)
> 
> > It seems that the tests get messed up due to using the same port numers.
> 
> I admittedly only took a very quick look, but I didn't find a log message that suggested a requested port was in use. Perhaps I missed it. Can you share what you're looking at?

@johnmaguire  I'm also not fully sure about this. An error message like you suggest would make it clear. But I can't explain why the failing test complains about not fitting ca-certificates otherwise.
What I'm looking at is this:
```
Invalid certificate from host" error="certificate validation failed: could not find ca for the certificate
```

and also this:

```
"Invalid certificate from host" error="certificate validation failed: certificate is expired"
```

But for the second one, I have an explanation meanwhile: the generated certificate for the test-services is only valid for 5 minutes. as the relevant test runns into a timeout after 10 minutes, its clear that this error occurs after 5 minutes.

Still, the first error I can't explain. The ca-key-pair is also generated freshly each time the test runs. From the ca-key-pair, the signed service key-pair and certificate is generated each time anew. So there should not be the possibility that these do not match. Expect, if one consideres that tests run in parallel an they accidentially connect with the wrong test-partner.

But I'm open for other ideas. Any help is welcome.

johnmaguire:
FYI we recently merged #1181 which caused conflicts. One conflict is caused by returning `*gonet.TCPConn` in lieu of `net.Conn`. Is it necessary to return the former, or we can continue to return the interface? Thanks!

cre4ture:
> FYI we recently merged #1181 which caused conflicts. One conflict is caused by returning `*gonet.TCPConn` in lieu of `net.Conn`. Is it necessary to return the former, or we can continue to return the interface? Thanks!

I merged the changes from main. It seems that the interface is OK.
Please tell me if you prefer a rebase, and/or a squash commit/merge.

udp/udp_linux.go
```diff
@@ -315,6 +321,10 @@ func (u *StdConn) getMemInfo(meminfo *[unix.SK_MEMINFO_VARS]uint32) error {
 
 func (u *StdConn) Close() error {
 	//TODO: this will not interrupt the read loop
```
cre4ture, line 323: @johnmaguire I think I have the test stable now. At least on linux. The issue was a unclean shutdown of the nebula service at the end of each test which caused confusion in the following tests.

The change here in this file solved the issue.

service/service.go
```diff
@@ -237,15 +243,49 @@ func (s *Service) Listen(network, address string) (net.Listener, error) {
...
+		Port:     uint16(addr.Port),
+		LinkAddr: "",
+	}, nil, ipv4.ProtocolNumber)
+}
+
 func (s *Service) Wait() error {
-	return s.eg.Wait()
+	err := s.eg.Wait()
+
+	s.ipstack.Destroy()
```
cre4ture, line 262: Adding this line made the windows test stable. They no run 200+ in a row without issue.

udp/udp_rio_windows.go
```diff
@@ -323,6 +329,14 @@ func (u *RIOConn) Close() error {
 	windows.PostQueuedCompletionStatus(u.rx.iocp, 0, 0, nil)
 	windows.PostQueuedCompletionStatus(u.tx.iocp, 0, 0, nil)
 
+	u.rx.mu.Lock()                   // for waiting till active reader is done
```
cre4ture, line 332: Had to run the test 1000 times to achieve stable reproduction on my windows 11. Is fixed with this change.

cre4ture:
@johnmaguire after extensive testing and a few findings I dare to claim that the tests are now stable on windows and linux. Can you please approve another workflow run on this PR? And have a look about what is still do be done to merge this?

ExplodingDragon:
@cre4ture Any new changes? Currently it seems to be incompatible with the master branch.

cre4ture:
@ExplodingDragon  thanks for the notification. I merged the changes in.

maggie44:
Is it possible to also add the ability to map two cidrs against each other? 

   - local_cidr: 10.0.0.0/24
      remote_cidr: 192.168.100.0/24

cre4ture:
> Is it possible to also add the ability to map two cidrs against each other?
> 
>     * local_cidr: 10.0.0.0/24
>       remote_cidr: 192.168.100.0/24

Hello maggie44,

thanks for your interesting question. I can't really answer that without further studying the documentation myself.
If nebula supports it, then its probably only possible with a real tun and thus root access.
I think this would be a good starting point for you:
https://nebula.defined.net/docs/config/tun/#tunroutes

hope this helps you a step forward.

maggie44:
I meant in relation to this PR. Based on the example in the initial post it is configured like this:

```
port_forwarding:
  outbound:
  # format of local and remote address: <host/ip>:<port>
  #- local_address: 127.0.0.1:3399
  #  remote_address: 192.168.100.92:4499
     # format of protocols lists (yml-list): [tcp], [udp], [tcp, udp]
  #  protocols: [tcp, udp]
```

I was querying whether it could also be configured like this:

```
port_forwarding:
  outbound:
  #- local_address: 10.0.0.0/24:3399
  #  remote_address: 192.168.100.0/24:4499
     # format of protocols lists (yml-list): [tcp], [udp], [tcp, udp]
  #  protocols: [tcp, udp]
```

But I see now that the local address would need an interface? Which would defeat the object of removing the TUN device.

maggie44:
Best I can think of right now would be to have TCP forwarding from a local IP, like `127.0.0.1:3399`, routed through to an IP based on path. This is very much an API related feature. For example:

```
http://127.0.0.1:3399/192.168.100.0
```

Where `192.168.100.0` is a path. Traffic routes in to `127.0.0.1:3399` which avoids requiring the TUN, then base on path `192.168.100.0` proxies traffic to the required host. That would allow opening TCP connections to an unlimited amount of connected hosts (I'm intentionally avoiding talking about UDP in relation to this, as the paths wouldn't be compatible). 

Straying off the original intent of this PR though, initially I was thinking of it only as a CIDR mapping which would have been a smaller change.

cre4ture:
@maggie44 Aha. I see the purpose of your question more clear now.
Actually I recalled that in the Feature request ( #1014) something similar like this is already mentioned which links to the feature request of having a SOCKS-Proxy (#915).
By following this approach, I thing you can achieve what you like. Just install a SOCKS proxy or similar on the remote host.
This way, one can probably also achieve the port-range "forwarding" request of @saket424 in the feature request.

maggie44:
Thanks @cre4ture.

Here is the sort of things I had in mind. Looks like it's possible with the gVisor `Service` already in the main branch:

```
func startHTTPReverseProxy(ctx context.Context, svc *service.Service) error {
    mux := http.NewServeMux()
    server := &http.Server{
        Addr:    ":3399",
        Handler: mux,
    }

    // Create a custom transport with the custom dial context
    transport := &http.Transport{
        DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
            return svc.DialContext(ctx, network, addr)
        },
    }

    mux.HandleFunc("/proxy/", func(w http.ResponseWriter, r *http.Request) {
        // Extract the target from query parameters
        target := r.URL.Query().Get("target")
        if target == "" {
            http.Error(w, "missing target parameter", http.StatusBadRequest)
            return
        }

        // Parse the target URL
        targetURL, err := url.Parse(target)
        if err != nil {
            http.Error(w, fmt.Sprintf("invalid target parameter: %s", err), http.StatusBadRequest)
            return
        }

        // Create a reverse proxy
        proxy := &httputil.ReverseProxy{
            Rewrite: func(r *httputil.ProxyRequest) {
                // Use only the target's path and query parameters. SetURL sets the majority of the URL fields.
                // Path and RawQuery replace the path to override the default which is to append the request
                // path to the target.
                r.SetURL(targetURL)
                r.Out.URL.Path = targetURL.Path
                r.Out.URL.RawQuery = targetURL.RawQuery
            },
            Transport: transport,
        }

        proxy.ServeHTTP(w, r)
    })

    go func() {
        <-ctx.Done()
        log.Println("shutting down server")
        _ = server.Shutdown(context.Background())
    }()

    log.Printf("starting server on %s", server.Addr)
    return server.ListenAndServe()
}
```

niowniow:
First of all, thank you for this contribution. It works very well for me.

I just wanted to note that port forwarding from the same node to itself does not work with this solution. I know that it's unnecessary because I can directly connect to the port on the same node, but by accident it happened to me. The following does not work on my machine with nebula/dial_address 192.168.100.3.

```
port_forwarding:
  enable_without_rules: false
  outbound:
  - listen_address: "127.0.0.1:5558"
    dial_address: "192.168.100.3:8880"
    protocols: [tcp, udp]
  inbound:
  - listen_port: 8880
    dial_address: 0.0.0.0:8880
    protocols: [tcp, udp]
```

I don't know enough details of how it works, so I can't debug it. It tries to connect but gets stuck.
```
wget http://localhost:5558/
--2024-12-19 10:05:30--  http://localhost:5558/
Resolving localhost (localhost)... 127.0.0.1
Connecting to localhost (localhost)|127.0.0.1|:5558... connected.
HTTP request sent, awaiting response...
```

Feel free to ignore, since I can bypass the issue by connecting directly to the port on the same machine.

cre4ture:
> dial_address

Hi @niowniow,

thanks for testing the implementation on this PR and your feedback.
I assume that "192.168.100.3" is your own IP in the nebula network. Right?

So, I can't really tell if this is supported or not as it depends on a part of the code that this PR doesn't touch. I think.
But did you check if dialing to "127.0.0.1" instead of "0.0.0.0" does help?
I currently getting the impression that there is the mixup of listen and dial IP.

cre4ture:
@johnmaguire @akernet 
Hello guys, is there any idea from your side when this PR can be reviewed from your side?

johnmaguire:
Hi @cre4ture, there's been a big ongoing push on #6 that has prevented us from looking too closely at this PR. I'm optimistic that we can start working through a backlog of PRs that deserve reviews soon.

ExplodingDragon:
@cre4ture Any new changes? Currently it seems to be incompatible with the master branch.

There appears to be a potential memory leak. (Version: 1.9.5 , Patch:  03630fbf3517a636777de0e96cfe2e2c1ea58086)

After restarting Nebula, the memory usage has changed.

ExplodingDragon:
> @ExplodingDragon I merged in the changes from master. In a first glance I could not detect any change that could cause a memory leak in the relevant code of this PR. Could you check if the leak is fixed with the latest master merge?

@cre4ture I've updated to the latest version (52623820c2be9571bb46acd16f5afd7811fe6542) and merged this PR, but can confirm the memory leak issue still persists (memory usage increased by 200%+ in ~6 hours, from 20M to 80M). Nodes without this PR merged not exhibit the problem. I'll analyze the details later.

<details>
<summary>Config</summary>

```yaml
pki:
  ca: /etc/nebula/ca.crt
  cert: /etc/nebula/host.crt
  key: /etc/nebula/host.key

static_host_map:
  "10.250.254.254": ["<secret>"]
  "10.250.254.253": ["<secret>"]
  "10.250.254.252": ["<secret>"]
  "10.250.254.251": ["<secret>"]


lighthouse:
  interval: 10
  hosts:
    - "10.250.254.254"

listen:
  host: "::"
  port: 3920

punchy:
  punch: true
  respond: true
  delay: 1s
  respond_delay: 5s

sshd:
  enabled: false

relay:
  relays:
    - "10.250.254.253"
    - "10.250.254.251"
  am_relay: false
  use_relays: true

tun:
  disabled: true

port_forwarding:
  enable_without_rules: true
  inbound:
    - listen_port: 22
      dial_address: "127.0.0.1:22"
      protocols: [tcp, udp]

    - listen_port: 80
      dial_address: "127.0.0.1:80"
      protocols: [tcp, udp]

    - listen_port: 443
      dial_address: "127.0.0.1:443"
      protocols: [tcp, udp]

logging:
  level: warn
  format: text

firewall:
  outbound_action: drop
  inbound_action: drop

  conntrack:
    tcp_timeout: 12m
    udp_timeout: 3m
    default_timeout: 10m

  outbound:
    - port: any
      proto: any
      host: any

  inbound:
    - port: any
      proto: icmp
      host: any

    - port: 22
      proto: tcp
      host: any

    - port: 80
      proto: tcp
      host: any

    - port: 443
      proto: tcp
      host: any
```

</details>

cre4ture:
@ExplodingDragon I extended the automated tests to do some stress testing and could identify one leaking go-routine per TCP connection. This is fixed, can you try to validate if this fixes your observation?

ExplodingDragon:
@cre4ture  all right, i think it's fixed. thanks.

nbrownus:
@cre4ture and @akernet thank you for all the work you've put into this PR! It certainly is a feature we have discussed and wished for many times over the years. 

One of the big questions that comes up when discussing this is: should this be included in the main `nebula` binary or released under a different binary, say `nebula-proxy`/`nebula-pf`/etc.

As for review, I will start going through it this month. At the moment we are trying to shore up the last bits of "ipv6 in the overlay" which as I'm sure you've seen through the merges of `master` has been quite a large change.

cre4ture:
Thanks for the status update. I'm looking forward to your review.

Regarding this point:

> should this be included in the main `nebula` binary or released under a different binary, say `nebula-proxy`/`nebula-pf`/etc

In general I'm open to change it. But I can't see any advantage by now.
I think it should stay in the nebula binary as the usage of it is the same.
It just needs a modified and extended configuration.

Idea: As one usually would call it as a non-root user, maybe the caller would want also force-disable the regular tun by setting a command line argument (instead/additionally of doing so in the config).

ExplodingDragon:
@nbrownus Any updates on this? I've been using this PR on 20+ devices for a while now, and everything looks good. Would be great to get this merged upstream.

Issue #1183, opened 2024-07-16
🐛 BUG:  after dns changed, connection lost forever


yylt:
### What version of `nebula` are you using? (`nebula -version`)

1.9.3

### What operating system are you using?

linux

### Describe the Bug

After the DNS change, the connection was not successful, and the log is as follows:

```
Jul 16 08:20:31 bpf nebula[762]: level=info msg="DNS results changed for host list" newSet="map[112.193.70.10:11010:{}]" origSet="&map[112.193.74.11:11010:{}]"
Jul 16 08:20:31 bpf nebula[762]: level=info msg="Handshake message sent" handshake="map[stage:1 style:ix_psk0]" initiatorIndex=4237819112 localIndex=4237819112 remoteIndex=0 udpAddrs=
"[112.193.74.11:11010 112.193.70.10:11010]" vpnIp=10.99.99.1
Jul 16 08:20:38 bpf nebula[762]: level=info msg="Handshake timed out" durationNs=6689486915 handshake="map[stage:1 style:ix_psk0]" initiatorIndex=4237819112 localIndex=4237819112 remo
teIndex=0 udpAddrs="[112.193.74.11:11010 112.193.70.10:11010]" vpnIp=10.99.99.1
Jul 16 08:20:41 bpf nebula[762]: level=info msg="Handshake message sent" handshake="map[stage:1 style:ix_psk0]" initiatorIndex=1848068562 localIndex=1848068562 remoteIndex=0 udpAddrs=
"[112.193.74.11:11010 112.193.70.10:11010]" vpnIp=10.99.99.1
Jul 16 08:20:48 bpf nebula[762]: level=info msg="Handshake timed out" durationNs=6589887970 handshake="map[stage:1 style:ix_psk0]" initiatorIndex=1848068562 localIndex=1848068562 remo
teIndex=0 udpAddrs="[112.193.74.11:11010 112.193.70.10:11010]" vpnIp=10.99.99.1
Jul 16 08:20:51 bpf nebula[762]: level=info msg="Handshake message sent" handshake="map[stage:1 style:ix_psk0]" initiatorIndex=860811292 localIndex=860811292 remoteIndex=0 udpAddrs="[
112.193.74.11:11010 112.193.70.10:11010]" vpnIp=10.99.99.1

```

The available address is `112.193.70.10`, It seems that only the first address is being used, which is the previous one and is currently unavailable.

### Logs from affected hosts

```

```

### Config files from affected hosts

```
static_host_map:
[...]

PR #1184, opened 2024-07-18
not update lighthouse ips if dns lookup failed

yylt:


remote_list.go
```diff
@@ -119,14 +119,17 @@ func NewHostnameResults(ctx context.Context, l *logrus.Logger, d time.Duration,
...
 			for {
+				lookupSuccess = true
 				netipAddrs := map[netip.AddrPort]struct{}{}
 				for _, hostPort := range r.hostnames {
 					timeoutCtx, timeoutCancel := context.WithTimeout(ctx, r.lookupTimeout)
 					addrs, err := net.DefaultResolver.LookupNetIP(timeoutCtx, r.network, hostPort.name)
 					timeoutCancel()
 					if err != nil {
 						l.WithFields(logrus.Fields{"hostname": hostPort.name, "network": r.network}).WithError(err).Error("DNS resolution failed for static_map host")
+						lookupSuccess = false
```
johnmaguire, line 132: I think for correctness, the default would need to be false and toggle this to `true` when err == nil. This is because if we try to resolve multiple hostnames for a given Nebula IP, and one fails and another succeeds, we'd like to use the results from the successful query.

johnmaguire:
Can you please describe the purpose of this change? In #1183 you stated it resolved some issues. Can you describe the issues this change resolved?

My understanding of the code here is that if DNS resolution fails, it may be seen as a "different" list of IPs than resolved previously. However, the results of this function are prepended to the overall cache here: https://github.com/slackhq/nebula/blob/16eaae306afe95ac9876e333fc1636c98cd999a4/lighthouse.go#L565-L569

In other words, I don't think a failure to resolve will end up _removing_ any previously successfully resolved queries. Given that that is the case, I'm not sure I understand what this PR is attempting to fix.

yylt:
> Can you please describe the purpose of this change? In #1183 you stated it resolved some issues. Can you describe the issues this change resolved?您能否描述一下此更改的目的？在#1183 中，您表示它解决了一些问题。您能描述一下此更改解决的问题吗？
> 
> My understanding of the code here is that if DNS resolution fails, it may be seen as a "different" list of IPs than resolved previously. However, the results of this function are prepended to the overall cache here: 我对这里代码的理解是，如果 DNS 解析失败，它可能会被视为与之前解析的“不同”IP 列表。然而，该函数的结果被添加到此处的整体缓存中：
> 
> https://github.com/slackhq/nebula/blob/16eaae306afe95ac9876e333fc1636c98cd999a4/lighthouse.go#L565-L569
> 
> In other words, I don't think a failure to resolve will end up _removing_ any previously successfully resolved queries. Given that that is the case, I'm not sure I understand what this PR is attempting to fix.换句话说，我认为解决失败不会最终删除任何以前成功解决的查询。鉴于这种情况，我不确定我是否理解此 PR 试图解决的问题。

I apologize, the PR and [issue](https://github.com/slackhq/nebula/issues/1183) mentioned different situation, both related to DDNS, so they were linked together. Here is log section and why submit this change.

```
Jul 12 03:24:25 node nebula[2115492]: level=error msg="DNS resolution failed for static_map host" error="lookup foo.bar.com: i/o timeout" hostname=foo.bar.com network=ip4
Jul 12 03:24:25 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[]" origSet="&map[112.193.70.10:11010:{}]"
Jul 12 03:24:55 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[112.193.70.10:11010:{}]" origSet="&map[]"
Jul 12 03:38:55 node nebula[2115492]: level=error msg="DNS resolution failed for static_map host" error="lookup foo.bar.com: i/o timeout" hostname=foo.bar.com network=ip4
Jul 12 03:38:55 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[]" origSet="&map[112.193.70.10:11010:{}]"
Jul 12 03:39:25 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[112.193.70.10:11010:{}]" origSet="&map[]"
Jul 12 03:43:55 node nebula[2115492]: level=error msg="DNS resolution failed for static_map host" error="lookup foo.bar.com: i/o timeout" hostname=foo.bar.com network=ip4
Jul 12 03:43:55 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[]" origSet="&map[112.193.70.10:11010:{}]"
Jul 12 03:44:25 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[112.193.70.10:11010:{}]" origSet="&map[]"
Jul 12 04:05:55 node nebula[2115492]: level=error msg="DNS resolution failed for static_map host" error="lookup foo.bar.com: i/o timeout" hostname=foo.bar.com network=ip4
Jul 12 04:05:55 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[]" origSet="&map[112.193.70.10:11010:{}]"
Jul 12 04:06:25 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[112.193.70.10:11010:{}]" origSet="&map[]"
Jul 12 04:30:25 node nebula[2115492]: level=error msg="DNS resolution failed for static_map host" error="lookup foo.bar.com: i/o timeout" hostname=foo.bar.com network=ip4
Jul 12 04:30:25 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[]" origSet="&map[112.193.70.10:11010:{}]"
Jul 12 04:30:55 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[112.193.70.10:11010:{}]" origSet="&map[]"
Jul 12 04:31:55 node nebula[2115492]: level=error msg="DNS resolution failed for static_map host" error="lookup foo.bar.com: i/o timeout" hostname=foo.bar.com network=ip4
Jul 12 04:31:55 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[]" origSet="&map[112.193.70.10:11010:{}]"
Jul 12 04:32:25 node nebula[2115492]: level=info msg="DNS results changed for host list" newSet="map[112.193.70.10:11010:{}]" origSet="&map[]"
```

Based on the above, it appears that DNS queries are fluctuating—sometimes there is an ip, and other times times out. 

This might be related to UDP traffic in the environment. However, since this is happening in virtual machine, I'm not sure how to investigate further, so handle the error if dns query failed.

johnmaguire:
@yylt Thanks for explaining. Indeed it looks like half of the DNS queries in the log are failing. That being said, the failing lookups should not be overriding the successful lookups in the cache Nebula uses to find hosts (which is a separate list from the one referenced in `origSet`.)

One way to verify this is to enable the [debug SSH server](https://nebula.defined.net/docs/config/sshd/#sshdenabled) and then issue the `list-lighthouse-addrmap` following a failed DNS query. You should still see the IP address present.

yylt:
Since the `nebula` at this node is already on the version that merged `the PR`, the issue has not recurred up to now, which has been around 50 days. 

Additionally, the DDNS changes approximately once a week, requiring SSH to be opened and checked after address change. It is unlikely to be feasible in the short term.

johnmaguire:
@yylt Understood that you're unable to test currently, but this PR doesn't appear to solve any issues and the code as written creates undesirable behaviors (such as a single DNS lookup failure causing successful queries to be discarded.) Thanks for the contribution, but I'm going to close the PR.

PR #1186, opened 2024-07-23
honor remote_allow_ilst in hole punch response

wadey:
When we receive a "hole punch notification" from a Lighthouse, we send a hole punch packet to every remote of that host, even if we don't include those remotes in our "remote_allow_list". Change the logic here to check if the remote IP is in our allow list before sending the hole punch packet.

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -1125,12 +1125,18 @@ func (lhh *LightHouseHandler) handleHostPunchNotification(n *NebulaMeta, vpnIp i
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


nbrownus (approved):

johnmaguire:
@JackDoan / @nbrownus  Do you know if the cert v2 work has made this PR unnecessary?

@wadey Can you fix the merge conflict if not? Thanks!

wadey:
fixed merge conflicts :+1:

nbrownus (approved):

PR #1188, opened 2024-07-31
Make some Nebula state programmatically available via control object

brad-defined:
Adds some hooks to the Control object that expose some lower level Nebula state. Gets a host certificate by IP, attempts to establish a tunnel to a peer, prints tunnel info, and queries a lighthouse.

### DIFF
--- a/control.go
+++ b/control.go
@@ -129,8 +129,7 @@ func (c *Control) ListHostmapIndexes(pendingMap bool) []ControlHostInfo {
 	}
 }
 
-// GetCertByVpnIp returns a single tunnels hostInfo, or nil if not found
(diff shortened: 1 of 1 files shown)


control.go
```diff
@@ -129,6 +129,43 @@ func (c *Control) ListHostmapIndexes(pendingMap bool) []ControlHostInfo {
 	}
 }
 
+// GetCertByVpnIp returns a single tunnels hostInfo, or nil if not found
```
nbrownus, line 132: No hostinfo here

brad-defined: PR updated

nbrownus (approved):

Issue #1189, opened 2024-08-04
🐛 BUG: mips-softfloat endianness is incorrect for port numbers in the static host map.


rawdigits:
### What version of `nebula` are you using? (`nebula -version`)

1.9.2

### What operating system are you using?

Linux

### Describe the Bug

I have been pulling my hair out trying to get nebula working on a router and while debugging with tcpdump, I saw that it was trying to handshake with the lighthouses using an incorrect port number. It should be trying to use 4242 but it is trying to reach port 37392 on both of them.

I _think_ this is the result of something being read in the wrong order due to an endianness mismatch, which results in the port number being wrong.

Nebula believes that it is reaching out to 4242, but tcpdump sees otherwise.

### Logs from affected hosts

```

INFO[0000] Handshake message sent                        handshake="map[stage:1 style:ix_psk0]" initiatorIndex=1642101477 localIndex=1642101477 remoteIndex=0 udpAddrs="[REMOVED:4242]" vpnIp=REMOVED

... udp port 37392 unreachable, length 317
```

### Config files from affected hosts

```

```

rawdigits:
Just a quick followup - I manually modified the port number in the static host map to be 37392, and the handshake packet now arrives to 4242. This is definitely an endianness mismatch.

rawdigits:
I finally tested rolling back (I built it manually) and 0.3.2 is working for me. This was definitely a bug introduced later.

rawdigits:
I built against nebula 1.8.2 and it works, so I think the changes to support v4 only is the issue. I'm assuming this line.

https://github.com/perfecto25/nebula/blame/82fafdf96808f379243b4ff073c4bd76673de0ae/udp/udp_linux.go#L229

rawdigits:
confirmed, this fixes the issue in a one-off build.

```
diff --git a/udp/udp_linux.go b/udp/udp_linux.go
index ef07243..ddbc48b 100644
--- a/udp/udp_linux.go
+++ b/udp/udp_linux.go
@@ -220,7 +220,7 @@ func (u *StdConn) writeTo6(b []byte, ip netip.AddrPort) error {
        rsa.Addr = ip.Addr().As16()
        port := ip.Port()
        // Little Endian -> Network Endian
-       rsa.Port = (port >> 8) | ((port & 0xff) << 8)
+       rsa.Port = port
 
        for {
                _, _, err := unix.Syscall6(
[...]

PR #1194, opened 2024-08-12
udp: fix endianness for port

wadey:
If the host OS is already big endian, we were swapping bytes when we shouldn't have. Use the Go helper to make sure we do the endianness correctly

Fixes: #1189

### DIFF
--- a/udp/udp_linux.go
+++ b/udp/udp_linux.go
@@ -218,9 +218,7 @@ func (u *StdConn) writeTo6(b []byte, ip netip.AddrPort) error {
 	var rsa unix.RawSockaddrInet6
 	rsa.Family = unix.AF_INET6
 	rsa.Addr = ip.Addr().As16()
-	port := ip.Port()
-	// Little Endian -> Network Endian
-	rsa.Port = (port >> 8) | ((port & 0xff) << 8)
+	binary.BigEndian.PutUint16((*[2]byte)(unsafe.Pointer(&rsa.Port))[:], ip.Port())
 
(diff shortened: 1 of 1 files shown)


JackDoanRivian (approved):
This made me curious how golang itself was handling portnumber-endianness, so I peeked into syscall_linux.go, this is very similar to what they do:

```
func (sa *SockaddrInet4) sockaddr() (unsafe.Pointer, _Socklen, error) {
	if sa.Port < 0 || sa.Port > 0xFFFF {
		return nil, 0, EINVAL
	}
	sa.raw.Family = AF_INET
	p := (*[2]byte)(unsafe.Pointer(&sa.raw.Port))
	p[0] = byte(sa.Port >> 8)
	p[1] = byte(sa.Port)
	sa.raw.Addr = sa.Addr
	return unsafe.Pointer(&sa.raw), SizeofSockaddrInet4, nil
}
```

except the use of binary.Bigendian makes it nice and obvious what's happening.

nbrownus (approved):

wadey:
> This made me curious how golang itself was handling portnumber-endianness, so I peeked into syscall_linux.go, this is very similar to what they do:

@JackDoanRivian I still don't understand how that works if the host OS is big or little endian, it seems like they don't do anything to handle that? I'm still confused about that. It does also make me wonder how our code in 1.8 also worked.

wadey:
> I still don't understand how that works if the host OS is big or little endian

ah ok I understand it now

PR #1197, opened 2024-08-13
make boringcrypto: add checklinkname flag for go1.23

wadey:
Starting with go1.23, we need to set -checklinkname=0 when building for boringcrypto because we need to use go:linkname to access `newGCMTLS`.

Note that this does break builds when using a go version less than go1.23.0. We can probably assume that someone using this Makefile and manually building is using the latest release of Go though.

See:

- https://go.dev/doc/go1.23#linker


wadey:
needs #1197

wadey:
merged into #1198

PR #1198, opened 2024-08-19
build with go1.23

wadey:
This doesn't change our go.mod, which still only requires go1.22 as a minimum. It only changes our builds to use go1.23 so we have the latest improvements.


wadey:
Needs #1197

nbrownus (approved):

PR #1216, opened 2024-09-13
V2 certificate format

nbrownus:
WIP based on cert-interface branch


cert/cert_v2.asn1
```diff
@@ -0,0 +1,38 @@
...
+    -- signature(details + curve + publicKey) using the appropriate method for curve
+    signature OCTET STRING
+}
+
+Details ::= SEQUENCE {
+    name Name,
+
+    -- At least 1 ipv4 or ipv6 address must be present if isCA is false
+    networks SEQUENCE OF Network,
+    unsafeNetworks SEQUENCE OF Network OPTIONAL,
```
johnmaguire, line 25: Very happy to see this getting renamed from `-ips` and `-subnets` which were a bit confusing (many people believed subnets was essentially the CIDR part of `-ips`.) That being said, calling them both "networks" might also be confusing. I think I'd lean towards simply calling this `unsafeRoutes` to match the config option.

cert/cert_v2.go
JackDoanRivian, line 1: thoughts on each cert variety going in its own subpackage for namespacing and stuff?

cert/cert_v2.asn1
nbrownus: I'm not opposed to calling these `unsafeRoutes`, it doesn't quite feel right though since this field is describing a network that can be used to configure unsafe routes on another host.

cert/cert_v2.asn1
```diff
@@ -0,0 +1,38 @@
+Nebula DEFINITIONS AUTOMATIC TAGS ::= BEGIN
+
+Name ::= UTF8String (SIZE (1..253))
+Time ::= INTEGER (0..18446744073709551615) -- uint64 maximum
```
JackDoanRivian, line 4: this is seconds-since-epoch, right?

nebula.proto
```diff
@@ -62,6 +62,7 @@ message NebulaHandshakeDetails {
   uint32 ResponderIndex = 3;
   uint64 Cookie = 4;
   uint64 Time = 5;
+  uint32 CertVersion = 8;
```
JackDoanRivian, line 65: thoughts on making this an enum to match cert.go?

cert/pem.go
```diff
@@ -30,19 +30,24 @@ func UnmarshalCertificateFromPEM(b []byte) (Certificate, []byte, error) {
...
 	case CertificateV2Banner:
-		//TODO
-		panic("TODO")
+		c, err = unmarshalCertificateV2(p.Bytes, nil)
 	default:
 		return nil, r, ErrInvalidPEMCertificateBanner
 	}
+
+	if err != nil {
+		return nil, r, err
```
JackDoanRivian, line 46: semantics changed a little bit here -- we used to not return the pem-remainder on err, now we do. It's probably fine? It's more consistent with the default/`ErrInvalidPEMCertificateBanner` branch.

cert/cert_v2.asn1
nbrownus: Yes

nebula.proto
nbrownus: It would be a cyclical import for the `cert` package or else we'd have to do silly type conversions between the two packages.

I will say using `cert.Version*` for protocol level stuff gave me the ick.

cert/cert_v2.asn1
```diff
@@ -0,0 +1,38 @@
...
+    networks SEQUENCE OF Network OPTIONAL,
+    unsafeNetworks SEQUENCE OF Network OPTIONAL,
+    groups SEQUENCE OF Name OPTIONAL,
+    isCA BOOLEAN DEFAULT false,
+
+    -- ASN.1 time formats are all string representations so we use our own uint64 limited one
+    notBefore Time,
+    notAfter Time,
+
+    issuer OCTET STRING OPTIONAL,
```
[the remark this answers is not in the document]
nbrownus: We are reusing the structure for CAs and they don't have an issuer. I need to put a comment describing the application behavior

lighthouse.go
```diff
@@ -995,74 +1044,96 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, vpnIp netip.Addr, a
...
+			panic("unsupported version")
 		}
-		n.Details.RelayVpnIp = append(n.Details.RelayVpnIp, relays...)
 	}
 }
 
-func (lhh *LightHouseHandler) handleHostQueryReply(n *NebulaMeta, vpnIp netip.Addr) {
-	if !lhh.lh.IsLighthouseIP(vpnIp) {
+func (lhh *LightHouseHandler) handleHostQueryReply(n *NebulaMeta, reqHostinfo *HostInfo) {
+	//TODO: this is kind of dumb
```
JackDoanRivian, line 1098: it is dumb, but it raises some good questions about multi-IP stuff. Are we going to require users to put _each_ IP for a lighthouse in their config file? If we do, does it make sense to group them?

```
lighthouse:
  hosts:
    - 10.0.0.1  # the old way
    - [fc00::2, 10.0.0.2]  # a new way
```
similar ideas apply wrt the static host map

overlay/tun_linux.go
```diff
@@ -324,32 +357,35 @@ func (t *tun) Activate() error {
...
+	if err = ioctl(t.ioctlFd, unix.SIOCSIFFLAGS, uintptr(unsafe.Pointer(&ifrf))); err != nil {
+		return fmt.Errorf("failed to run tun device: %s", err)
 	}
-	t.deviceIndex = link.Attrs().Index
 
-	if err = t.setDefaultRoute(); err != nil {
-		return err
+	//set route MTU
+	for i := range t.vpnNetworks {
+		if err = t.setDefaultRoute(t.vpnNetworks[i]); err != nil {
```
JackDoanRivian, line 378: this still races and errors sometimes on my proxmox host. Not sure why. On hosts that use systemd-networkd I haven't observed any issues.

lighthouse.go
nbrownus: From a user perspective I would say no. I think it would be better to layer in learned lighthouse vpn addrs but as a stop gap I have included a `IsAnyLighthouseIP` (which the name is wrong ugh) that checks if any of the addrs are a lighthouse addr.

cmd/nebula-cert/sign.go
```diff
@@ -146,37 +170,57 @@ func signCert(args []string, out io.Writer, errOut io.Writer, pr PasswordReader)
...
-	network, err := netip.ParsePrefix(*sf.ip)
-	if err != nil {
-		return newHelpErrorf("invalid ip definition: %s", *sf.ip)
-	}
-	if !network.Addr().Is4() {
-		return newHelpErrorf("invalid ip definition: can only be ipv4, have %s", *sf.ip)
-	}
+	if *sf.networks != "" {
+		for _, rs := range strings.Split(*sf.networks, ",") {
+			//TODO: error on duplicates? Mainly only addr matters, having two of the same addr in the same or different prefix space is strange
```
nbrownus, line 175: Also no 4in6

cmd/nebula-cert/sign.go
```diff
@@ -146,37 +170,57 @@ func signCert(args []string, out io.Writer, errOut io.Writer, pr PasswordReader)
...
+	if *sf.unsafeNetworks != "" {
+		//TODO: error on duplicates?
+		for _, rs := range strings.Split(*sf.unsafeNetworks, ",") {
 			rs := strings.Trim(rs, " ")
 			if rs != "" {
-				s, err := netip.ParsePrefix(rs)
+				n, err := netip.ParsePrefix(rs)
 				if err != nil {
-					return newHelpErrorf("invalid subnet definition: %s", rs)
+					return newHelpErrorf("invalid -unsafe-networks definition: %s", rs)
```
nbrownus, line 206: Not `-unsafe-networks` just `network`

nbrownus: And no 4in6

cmd/nebula-cert/sign.go
```diff
@@ -243,18 +274,85 @@ func signCert(args []string, out io.Writer, errOut io.Writer, pr PasswordReader)
...
+		if len(v4Networks) != 1 {
+			return newHelpErrorf("invalid -networks definition: v1 certificates can only have a single ipv4 address")
 		}
-	} else {
-		c, err = t.SignPkcs11(caCert, curve, p11Client)
-		if err != nil {
-			return fmt.Errorf("error while signing with PKCS#11: %w", err)
+
+		if version == cert.Version1 {
+			// If we are asked to mint a v1 certificate only then we cant just ignore any v6 addresses
```
nbrownus, line 289: We arent ignoring them

cert/cert_v1.go
```diff
@@ -173,181 +170,163 @@ func (nc *certificateV1) VerifyPrivateKey(curve Curve, key []byte) error {
...
-	s += fmt.Sprintf("\t\tPublic key: %x\n", nc.details.PublicKey)
-	s += fmt.Sprintf("\t\tCurve: %s\n", nc.details.Curve)
-	s += "\t}\n"
-	fp, err := nc.Fingerprint()
-	if err == nil {
-		s += fmt.Sprintf("\tFingerprint: %s\n", fp)
+func (c *certificateV1) String() string {
+	b, err := json.MarshalIndent(c.marshalJSON(), "", "\t")
+	if err != nil {
+		return "<error marshalling certificate>"
```
nbrownus, line 213: #1253 extends this with the error and asks if we'd rather `panic`

cert/cert_v2.go
```diff
@@ -0,0 +1,621 @@
...
+	"net/netip"
+	"slices"
+	"time"
+
+	"golang.org/x/crypto/cryptobyte"
+	"golang.org/x/crypto/cryptobyte/asn1"
+	"golang.org/x/crypto/curve25519"
+)
+
+//TODO: should we avoid hex encoding shit on output? Just let it be base64?
```
nbrownus, line 23: Remove

cert/cert_v2.go
```diff
@@ -0,0 +1,621 @@
...
+	return c.signature
+}
+
+func (c *certificateV2) UnsafeNetworks() []netip.Prefix {
+	return c.details.unsafeNetworks
+}
+
+func (c *certificateV2) Fingerprint() (string, error) {
+	b := make([]byte, len(c.rawDetails)+1+len(c.publicKey))
+	//TODO: double check this, panic on empty raw details
```
nbrownus, line 129: How should we treat this situation

cert/cert_v2.go
```diff
@@ -0,0 +1,621 @@
...
+	b[len(c.rawDetails)] = byte(c.curve)
+	copy(b[len(c.rawDetails)+1:], c.publicKey)
+	copy(b[len(c.rawDetails)+1+len(c.publicKey):], c.signature)
+	sum := sha256.Sum256(b)
+	return hex.EncodeToString(sum[:]), nil
+}
+
+func (c *certificateV2) CheckSignature(key []byte) bool {
+	b := make([]byte, len(c.rawDetails)+1+len(c.publicKey))
+	//TODO: double check this, panic on empty raw details
```
nbrownus, line 140: How should we treat this situation

cert/cert_v2.go
```diff
@@ -0,0 +1,621 @@
...
+		return fmt.Errorf("public key in cert and private key supplied don't match")
+	}
+
+	return nil
+}
+
+func (c *certificateV2) String() string {
+	b, err := json.MarshalIndent(c.marshalJSON(), "", "\t")
+	if err != nil {
+		return "<error marshalling certificate>"
```
nbrownus, line 220: #1253 does not give this the same treatment as v1 certs but it should match whatever we decide there

cert/cert_v2.go
```diff
@@ -0,0 +1,621 @@
...
+	return string(b)
+}
+
+func (c *certificateV2) MarshalForHandshakes() ([]byte, error) {
+	var b cryptobyte.Builder
+	// Outermost certificate
+	b.AddASN1(asn1.SEQUENCE, func(b *cryptobyte.Builder) {
+
+		// Add the cert details which is already marshalled
+		//TODO: panic on nil rawDetails
```
nbrownus, line 231: Panic or error?

cert/cert_v2.go
```diff
@@ -0,0 +1,621 @@
...
+func (c *certificateV2) marshalForSigning() ([]byte, error) {
+	d, err := c.details.Marshal()
+	if err != nil {
+		//TODO: annotate?
+		return nil, err
+	}
+	c.rawDetails = d
+
+	b := make([]byte, len(c.rawDetails)+1+len(c.publicKey))
+	//TODO: double check this
```
nbrownus, line 360: Needs another set of eyes to verify

cert/cert_v2.go
```diff
@@ -0,0 +1,621 @@
...
+
+	// Maybe grab the public key
+	var rawPublicKey cryptobyte.String
+	if len(publicKey) > 0 {
+		rawPublicKey = publicKey
+	} else if !input.ReadOptionalASN1(&rawPublicKey, nil, TagCertPublicKey) {
+		return nil, ErrBadFormat
+	}
+
+	//TODO: Assert public key length
```
nbrownus, line 491: Do we want to?

cert/sign.go
```diff
@@ -24,27 +29,62 @@ type TBSCertificate struct {
...
-func (t *TBSCertificate) sign(signer Certificate, curve Curve, key []byte, client *pkclient.PKClient) (Certificate, error) {
+// SignWith does the same thing as sign, but uses the function in `sp` to calculate the signature.
+// You should only use SignWith if you do not have direct access to your private key.
+func (t *TBSCertificate) SignWith(signer Certificate, curve Curve, sp SignerLambda) (Certificate, error) {
 	if curve != t.Curve {
 		return nil, fmt.Errorf("curve in cert and private key supplied don't match")
 	}
 
 	//TODO: make sure we have all minimum properties to sign, like a public key
+	//TODO: we need to verify networks and unsafe networks (no duplicates, max of 1 of each version for v2 certs
```
nbrownus, line 87: Which checks are we missing that we want to include?

cert/sign.go
```diff
@@ -67,10 +107,55 @@ func (t *TBSCertificate) sign(signer Certificate, curve Curve, key []byte, clien
...
+	if err != nil {
+		return nil, err
+	}
+
+	sig, err := sp(certBytes)
+	if err != nil {
+		return nil, err
+	}
+
+	//TODO: check if we have sig bytes?
```
nbrownus, line 141: Do we want an extra test in case `sp` returns no error but nil `sig`?

cmd/nebula-cert/print.go
```diff
@@ -49,20 +49,19 @@ func printCert(args []string, out io.Writer, errOut io.Writer) error {
 	var qrBytes []byte
 	part := 0
 
+	var jsonCerts []cert.Certificate
```
nbrownus, line 52: Main thing to note here is that we don't have a human string version anymore. Its always json. Do we want this?

cmd/nebula-cert/sign.go
```diff
@@ -146,37 +170,57 @@ func signCert(args []string, out io.Writer, errOut io.Writer, pr PasswordReader)
...
-		for _, rs := range strings.Split(*sf.subnets, ",") {
+	var v4UnsafeNetworks []netip.Prefix
+	var v6UnsafeNetworks []netip.Prefix
+	if *sf.unsafeNetworks == "" && *sf.subnets != "" {
+		// Pull up deprecated -subnets flag if needed
+		*sf.unsafeNetworks = *sf.subnets
+	}
+
+	if *sf.unsafeNetworks != "" {
+		//TODO: error on duplicates?
```
nbrownus, line 200: Dupe check

connection_manager.go
```diff
@@ -235,11 +235,11 @@ func (n *connectionManager) migrateRelayUsed(oldhostinfo, newhostinfo *HostInfo)
 			index = existing.LocalIndex
 			switch r.Type {
 			case TerminalType:
-				relayFrom = n.intf.myVpnNet.Addr()
-				relayTo = existing.PeerIp
+				relayFrom = n.intf.myVpnAddrs[0]
```
nbrownus, line 238: I think we will want to find the matching network here instead of using the first vpn addr like in #1261

connection_manager.go
```diff
@@ -235,11 +235,11 @@ func (n *connectionManager) migrateRelayUsed(oldhostinfo, newhostinfo *HostInfo)
...
+				relayTo = existing.PeerAddr
 			case ForwardingType:
-				relayFrom = existing.PeerIp
-				relayTo = newhostinfo.vpnIp
+				relayFrom = existing.PeerAddr
+				relayTo = newhostinfo.vpnAddrs[0]
```
nbrownus, line 242: I think we will want to find the matching network here instead of using the first vpn addr like in https://github.com/slackhq/nebula/pull/1261

connection_manager.go
```diff
@@ -253,45 +253,64 @@ func (n *connectionManager) migrateRelayUsed(oldhostinfo, newhostinfo *HostInfo)
...
+			index, err = AddRelay(n.l, newhostinfo, n.hostMap, r.PeerAddr, nil, r.Type, Requested)
 			if err != nil {
 				n.l.WithError(err).Error("failed to migrate relay to new hostinfo")
 				continue
 			}
 			switch r.Type {
 			case TerminalType:
-				relayFrom = n.intf.myVpnNet.Addr()
-				relayTo = r.PeerIp
+				relayFrom = n.intf.myVpnAddrs[0]
```
nbrownus, line 263: I think we will want to find the matching network here instead of using the first vpn addr like in https://github.com/slackhq/nebula/pull/1261

connection_manager.go
```diff
@@ -253,45 +253,64 @@ func (n *connectionManager) migrateRelayUsed(oldhostinfo, newhostinfo *HostInfo)
...
+				relayTo = r.PeerAddr
 			case ForwardingType:
-				relayFrom = r.PeerIp
-				relayTo = newhostinfo.vpnIp
+				relayFrom = r.PeerAddr
+				relayTo = newhostinfo.vpnAddrs[0]
```
nbrownus, line 267: I think we will want to find the matching network here instead of using the first vpn addr like in https://github.com/slackhq/nebula/pull/1261

connection_manager.go
```diff
@@ -313,7 +332,7 @@ func (n *connectionManager) makeTrafficDecision(localIndex uint32, now time.Time
 		return closeTunnel, hostinfo, nil
 	}
 
-	primary := n.hostMap.Hosts[hostinfo.vpnIp]
+	primary := n.hostMap.Hosts[hostinfo.vpnAddrs[0]]
```
nbrownus, line 335: I think we will want to find the matching network here instead of using the first vpn addr like in https://github.com/slackhq/nebula/pull/1261

connection_manager.go
```diff
@@ -407,21 +426,24 @@ func (n *connectionManager) shouldSwapPrimary(current, primary *HostInfo) bool {
 	// If we are here then we have multiple tunnels for a host pair and neither side believes the same tunnel is primary.
 	// Let's sort this out.
 
-	if current.vpnIp.Compare(n.intf.myVpnNet.Addr()) < 0 {
+	//TODO: current.vpnIp should become an array of vpnIps
+	if current.vpnAddrs[0].Compare(n.intf.myVpnAddrs[0]) < 0 {
```
nbrownus, line 430: I think we will want to find the matching network here instead of using the first vpn addr like in https://github.com/slackhq/nebula/pull/1261

connection_manager.go
```diff
@@ -407,21 +426,24 @@ func (n *connectionManager) shouldSwapPrimary(current, primary *HostInfo) bool {
 	// If we are here then we have multiple tunnels for a host pair and neither side believes the same tunnel is primary.
 	// Let's sort this out.
 
-	if current.vpnIp.Compare(n.intf.myVpnNet.Addr()) < 0 {
+	//TODO: current.vpnIp should become an array of vpnIps
```
nbrownus, line 429: Remove TODO

connection_manager.go
```diff
@@ -407,21 +426,24 @@ func (n *connectionManager) shouldSwapPrimary(current, primary *HostInfo) bool {
...
+	if current.vpnAddrs[0].Compare(n.intf.myVpnAddrs[0]) < 0 {
 		// Only one side should flip primary because if both flip then we may never resolve to a single tunnel.
 		// vpn ip is static across all tunnels for this host pair so lets use that to determine who is flipping.
 		// The remotes vpn ip is lower than mine. I will not flip.
 		return false
 	}
 
-	certState := n.intf.pki.GetCertState()
-	return bytes.Equal(current.ConnectionState.myCert.Signature(), certState.Certificate.Signature())
+	//TODO: we should favor v2 over v1 certificates if configured to send them
```
nbrownus, line 437: Resolve TODO

connection_manager.go
```diff
@@ -407,21 +426,24 @@ func (n *connectionManager) shouldSwapPrimary(current, primary *HostInfo) bool {
...
+
+	crt := n.intf.pki.getCertificate(current.ConnectionState.myCert.Version())
+	return bytes.Equal(current.ConnectionState.myCert.Signature(), crt.Signature())
 }
 
 func (n *connectionManager) swapPrimary(current, primary *HostInfo) {
 	n.hostMap.Lock()
 	// Make sure the primary is still the same after the write lock. This avoids a race with a rehandshake.
-	if n.hostMap.Hosts[current.vpnIp] == primary {
+	if n.hostMap.Hosts[current.vpnAddrs[0]] == primary {
```
nbrownus, line 446: I think we will want to find the matching network here instead of using the first vpn addr like in https://github.com/slackhq/nebula/pull/1261

This one is a bit more tricky since we want to make assertions based on the vpn addr being used but it could be multiple

connection_manager.go
```diff
@@ -473,14 +495,16 @@ func (n *connectionManager) sendPunch(hostinfo *HostInfo) {
...
 func (n *connectionManager) tryRehandshake(hostinfo *HostInfo) {
-	certState := n.intf.pki.GetCertState()
-	if bytes.Equal(hostinfo.ConnectionState.myCert.Signature(), certState.Certificate.Signature()) {
+	crt := n.intf.pki.getCertificate(hostinfo.ConnectionState.myCert.Version())
+	if bytes.Equal(hostinfo.ConnectionState.myCert.Signature(), crt.Signature()) {
 		return
 	}
 
-	n.l.WithField("vpnIp", hostinfo.vpnIp).
+	//TODO: we should favor v2 over v1 certificates if configured to send them
```
nbrownus, line 503: Resolve TODO

connection_manager.go
```diff
@@ -473,14 +495,16 @@ func (n *connectionManager) sendPunch(hostinfo *HostInfo) {
...
+
+	n.l.WithField("vpnAddrs", hostinfo.vpnAddrs).
 		WithField("reason", "local certificate is not current").
 		Info("Re-handshaking with remote")
 
-	n.intf.handshakeManager.StartHandshake(hostinfo.vpnIp, nil)
+	n.intf.handshakeManager.StartHandshake(hostinfo.vpnAddrs[0], nil)
```
nbrownus, line 509: We likely want to know which vpn addr we were using or at least find the one we can route for this rehandshake

control.go
```diff
@@ -131,10 +131,13 @@ func (c *Control) ListHostmapIndexes(pendingMap bool) []ControlHostInfo {
 
 // GetCertByVpnIp returns the authenticated certificate of the given vpn IP, or nil if not found
 func (c *Control) GetCertByVpnIp(vpnIp netip.Addr) cert.Certificate {
-	if c.f.myVpnNet.Addr() == vpnIp {
-		return c.f.pki.GetCertState().Certificate.Copy()
+	_, found := c.f.myVpnAddrsTable.Lookup(vpnIp)
+	if found {
+		//TODO: we might have 2 certs....
```
nbrownus, line 136: Resolve TODOs

control.go
```diff
@@ -229,14 +232,14 @@ func (c *Control) CloseAllTunnels(excludeLighthouses bool) (closed int) {
 
 	shutdown := func(h *HostInfo) {
 		if excludeLighthouses {
-			if _, ok := lighthouses[h.vpnIp]; ok {
+			if _, ok := lighthouses[h.vpnAddrs[0]]; ok {
```
nbrownus, line 235: `IsAnyLighthouseAddr`

control.go
```diff
@@ -229,14 +232,14 @@ func (c *Control) CloseAllTunnels(excludeLighthouses bool) (closed int) {
...
 				return
 			}
 		}
 		c.f.send(header.CloseTunnel, 0, h.ConnectionState, h, []byte{}, make([]byte, 12, 12), make([]byte, mtu))
 		c.f.closeTunnel(h)
 
-		c.l.WithField("vpnIp", h.vpnIp).WithField("udpAddr", h.remote).
+		c.l.WithField("vpnIp", h.vpnAddrs[0]).WithField("udpAddr", h.remote).
```
nbrownus, line 242: If we land on a "used" vpn addr then log that instead

control.go
```diff
@@ -246,15 +249,15 @@ func (c *Control) CloseAllTunnels(excludeLighthouses bool) (closed int) {
 	// Grab the hostMap lock to access the Relays map
 	c.f.hostMap.Lock()
 	for _, relayingHost := range c.f.hostMap.Relays {
-		relayingHosts[relayingHost.vpnIp] = relayingHost
+		relayingHosts[relayingHost.vpnAddrs[0]] = relayingHost
```
nbrownus, line 252: Likely need to check all vpn addrs

control.go
```diff
@@ -246,15 +249,15 @@ func (c *Control) CloseAllTunnels(excludeLighthouses bool) (closed int) {
...
 	}
 	c.f.hostMap.Unlock()
 
 	hostInfos := []*HostInfo{}
 	// Grab the hostMap lock to access the Hosts map
 	c.f.hostMap.Lock()
 	for _, relayHost := range c.f.hostMap.Indexes {
-		if _, ok := relayingHosts[relayHost.vpnIp]; !ok {
+		if _, ok := relayingHosts[relayHost.vpnAddrs[0]]; !ok {
```
nbrownus, line 260: Likely need to check all vpn addrs

e2e/helpers_test.go
```diff
@@ -132,27 +148,28 @@ func deadline(t *testing.T, seconds time.Duration) doneCb {
...
 	aPacket := r.RouteForAllUntilTxTun(controlB)
 	assertUdpPacket(t, []byte("Hello from A"), aPacket, vpnIpA, vpnIpB, 90, 80)
 }
 
-func assertHostInfoPair(t *testing.T, addrA, addrB netip.AddrPort, vpnIpA, vpnIpB netip.Addr, controlA, controlB *nebula.Control) {
+func assertHostInfoPair(t *testing.T, addrA, addrB netip.AddrPort, vpnNetsA, vpnNetsB []netip.Prefix, controlA, controlB *nebula.Control) {
 	// Get both host infos
-	hBinA := controlA.GetHostInfoByVpnIp(vpnIpB, false)
-	assert.NotNil(t, hBinA, "Host B was not found by vpnIp in controlA")
+	//TODO: we may want to loop over each vpnAddr and assert all the things
```
nbrownus, line 163: Resolve TODOs

examples/config.yml
```diff
@@ -336,10 +342,13 @@ firewall:
   #   host: `any` or a literal hostname, ie `test-host`
   #   group: `any` or a literal group name, ie `default-group`
   #   groups: Same as group but accepts a list of values. Multiple values are AND'd together and a certificate would have to contain all groups to pass
-  #   cidr: a remote CIDR, `0.0.0.0/0` is any.
-  #   local_cidr: a local CIDR, `0.0.0.0/0` is any. This could be used to filter destinations when using unsafe_routes.
-  #      Default is `any` unless the certificate contains subnets and then the default is the ip issued in the certificate
-  #      if `default_local_cidr_any` is false, otherwise its `any`.
+  #   cidr: a remote CIDR, `0.0.0.0/0` is any ipv4 and `::/0` is any ipv6. //TODO: we have a problem, firewall needs to understand this and should probably allow `any` for both
```
nbrownus, line 345: Resolve TODO

firewall.go
```diff
@@ -22,7 +23,8 @@ import (
 )
 
 type FirewallInterface interface {
-	AddRule(incoming bool, proto uint8, startPort int32, endPort int32, groups []string, host string, ip, localIp netip.Prefix, caName string, caSha string) error
+	//TODO: name these better addr, localAddr. Are they vpnAddrs?
```
nbrownus, line 26: Resolve TODO

firewall.go
```diff
@@ -424,26 +432,25 @@ func (f *Firewall) Drop(fp firewall.Packet, incoming bool, h *HostInfo, caPool *
...
 		if !ok {
-			f.metrics(incoming).droppedRemoteIP.Inc(1)
+			f.metrics(incoming).droppedRemoteAddr.Inc(1)
 			return ErrInvalidRemoteIP
 		}
 	} else {
 		// Simple case: Certificate has one IP and no subnets
-		if fp.RemoteIP != h.vpnIp {
-			f.metrics(incoming).droppedRemoteIP.Inc(1)
+		//TODO: we can make this more performant
```
nbrownus, line 443: Resolve TODO

handshake_ix.go
```diff
@@ -349,16 +428,17 @@ func ixHandshakeStage2(f *Interface, addr netip.AddrPort, via *ViaSender, hh *Ha
 
 	hostinfo := hh.hostinfo
 	if addr.IsValid() {
-		if !f.lightHouse.GetRemoteAllowList().Allow(hostinfo.vpnIp, addr.Addr()) {
-			f.l.WithField("vpnIp", hostinfo.vpnIp).WithField("udpAddr", addr).Debug("lighthouse.remote_allow_list denied incoming handshake")
+		//TODO: this is kind of nonsense now
```
nbrownus, line 431: Resolve TODO

handshake_manager.go
```diff
@@ -286,17 +294,35 @@ func (hm *HandshakeManager) handleOutbound(vpnIp netip.Addr, lighthouseTriggered
...
-					theirVpnIpB := vpnIp.As4()
-
-					// Re-send the CreateRelay request, in case the previous one was lost.
 					m := NebulaControl{
 						Type:                NebulaControl_CreateRelayRequest,
 						InitiatorRelayIndex: existingRelay.LocalIndex,
-						RelayFromIp:         binary.BigEndian.Uint32(myVpnIpB[:]),
-						RelayToIp:           binary.BigEndian.Uint32(theirVpnIpB[:]),
 					}
+
```
nbrownus, line 301: This is really a test to see if we have any common networks despite the version

interface.go
```diff
@@ -417,11 +419,20 @@ func (f *Interface) emitStats(ctx context.Context, i time.Duration) {
 			f.firewall.EmitStats()
 			f.handshakeManager.EmitStats()
 			udpStats()
-			certExpirationGauge.Update(int64(f.pki.GetCertState().Certificate.NotAfter().Sub(time.Now()) / time.Second))
+			certExpirationGauge.Update(int64(f.pki.getDefaultCertificate().NotAfter().Sub(time.Now()) / time.Second))
+			//TODO: we should also report the default certificate version
```
nbrownus, line 423: Resolve TODO

lighthouse.go
```diff
@@ -516,206 +525,205 @@ func (lh *LightHouse) queryAndPrepMessage(vpnIp netip.Addr, f func(*cache) (int,
...
-}
-
-// TODO: IPV6-WORK we can delete some more of these
-func NewIp6AndPortFromNetIP(ip netip.Addr, port uint16) *Ip6AndPort {
-	ip6Addr := ip.As16()
-	return &Ip6AndPort{
-		Hi:   binary.BigEndian.Uint64(ip6Addr[:8]),
-		Lo:   binary.BigEndian.Uint64(ip6Addr[8:]),
-		Port: uint32(port),
+// TODO: IsLighthouseAddr should be sufficient, we just need to update the vpnAddrs for lighthouses after a handshake
```
nbrownus, line 717: Resolve TODO

lighthouse.go
```diff
@@ -785,65 +847,116 @@ func (lh *LightHouse) StartUpdateWorker() {
...
+			if v1Update == nil {
+				var relays []uint32
+				for _, r := range lh.GetRelaysForMe() {
+					if !r.Is4() {
+						continue
+					}
+					b := r.As4()
+					relays = append(relays, binary.BigEndian.Uint32(b[:]))
+				}
+				//TODO: assert ipv4
```
nbrownus, line 902: Resolve TODO

lighthouse.go
```diff
@@ -886,59 +999,55 @@ func (lhh *LightHouseHandler) resetMeta() *NebulaMeta {
...
 
 	// Keep the array memory around
-	details.Ip4AndPorts = details.Ip4AndPorts[:0]
-	details.Ip6AndPorts = details.Ip6AndPorts[:0]
-	details.RelayVpnIp = details.RelayVpnIp[:0]
+	details.V4AddrPorts = details.V4AddrPorts[:0]
+	details.V6AddrPorts = details.V6AddrPorts[:0]
+	details.RelayVpnAddrs = details.RelayVpnAddrs[:0]
+	details.OldRelayVpnAddrs = details.OldRelayVpnAddrs[:0]
+	//TODO: these are unfortunate
```
nbrownus, line 1006: We could likely slightly benefit by a change to the protocol in resolving this

lighthouse.go
```diff
@@ -971,21 +1092,39 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, vpnIp netip.Addr, a
...
+			b := fromVpnAddrs[0].As4()
+			n.Details.OldVpnAddr = binary.BigEndian.Uint32(b[:])
+			lhh.coalesceAnswers(useVersion, c, n)
+
+		} else if useVersion == cert.Version2 {
+			n.Details.VpnAddr = netAddrToProtoAddr(fromVpnAddrs[0])
+			lhh.coalesceAnswers(useVersion, c, n)
+
+		} else {
+			panic("unsupported version")
```
nbrownus, line 1126: Panic or err?

lighthouse.go
```diff
@@ -995,139 +1134,192 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, vpnIp netip.Addr, a
...
+				n.Details.OldRelayVpnAddrs = append(n.Details.OldRelayVpnAddrs, binary.BigEndian.Uint32(b[:]))
+			}
+
+		} else if v == cert.Version2 {
+			for _, r := range c.relay.relay {
+				n.Details.RelayVpnAddrs = append(n.Details.RelayVpnAddrs, netAddrToProtoAddr(r))
+			}
+
+		} else {
+			panic("unsupported version")
```
nbrownus, line 1182: Panic or error?

lighthouse.go
```diff
@@ -995,139 +1134,192 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, vpnIp netip.Addr, a
...
+		b := [4]byte{}
+		binary.BigEndian.PutUint32(b[:], n.Details.OldVpnAddr)
+		detailsVpnAddr = netip.AddrFrom4(b)
+		useVersion = 1
+	} else if n.Details.VpnAddr != nil {
+		detailsVpnAddr = protoAddrToNetAddr(n.Details.VpnAddr)
+		useVersion = 2
+	}
+
+	//todo hosts with only v2 certs cannot provide their ipv6 addr when contacting the lighthouse via v4?
```
nbrownus, line 1256: Resolve TODOs

lighthouse.go
```diff
@@ -995,139 +1134,192 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, vpnIp netip.Addr, a
...
+			return
+		}
+		vpnAddrB := fromVpnAddrs[0].As4()
+		n.Details.OldVpnAddr = binary.BigEndian.Uint32(vpnAddrB[:])
+
+	} else if useVersion == cert.Version2 {
+		n.Details.VpnAddr = netAddrToProtoAddr(fromVpnAddrs[0])
+
+	} else {
+		panic("unsupported version")
```
nbrownus, line 1307: Panic or error?

lighthouse.go
```diff
@@ -995,139 +1134,192 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, vpnIp netip.Addr, a
...
 
 	lhh.lh.metricTx(NebulaMeta_HostUpdateNotificationAck, 1)
-	w.SendMessageToVpnIp(header.LightHouse, 0, vpnIp, lhh.pb[:ln], lhh.nb, lhh.out[:0])
+	w.SendMessageToVpnAddr(header.LightHouse, 0, fromVpnAddrs[0], lhh.pb[:ln], lhh.nb, lhh.out[:0])
 }
 
-func (lhh *LightHouseHandler) handleHostPunchNotification(n *NebulaMeta, vpnIp netip.Addr, w EncWriter) {
-	if !lhh.lh.IsLighthouseIP(vpnIp) {
+func (lhh *LightHouseHandler) handleHostPunchNotification(n *NebulaMeta, fromVpnAddrs []netip.Addr, w EncWriter) {
+	//TODO: this is kinda stupid
```
nbrownus, line 1321: Resolve TODO

outside.go
```diff
@@ -262,25 +240,26 @@ func (f *Interface) sendCloseTunnel(h *HostInfo) {
...
 }
 
-func (f *Interface) handleHostRoaming(hostinfo *HostInfo, ip netip.AddrPort) {
-	if ip.IsValid() && hostinfo.remote != ip {
-		if !f.lightHouse.GetRemoteAllowList().Allow(hostinfo.vpnIp, ip.Addr()) {
-			hostinfo.logger(f.l).WithField("newAddr", ip).Debug("lighthouse.remote_allow_list denied roaming")
+func (f *Interface) handleHostRoaming(hostinfo *HostInfo, vpnAddr netip.AddrPort) {
+	if vpnAddr.IsValid() && hostinfo.remote != vpnAddr {
+		//TODO: this is weird now that we can have multiple vpn addrs
+		if !f.lightHouse.GetRemoteAllowList().Allow(hostinfo.vpnAddrs[0], vpnAddr.Addr()) {
```
nbrownus, line 246: Resolve TODO

outside.go
```diff
@@ -302,22 +281,122 @@ func (f *Interface) handleEncrypted(ci *ConnectionState, addr netip.AddrPort, h
...
+
+	if incoming {
+		fp.RemoteAddr, _ = netip.AddrFromSlice(data[8:24])
+		fp.LocalAddr, _ = netip.AddrFromSlice(data[24:40])
+	} else {
+		fp.LocalAddr, _ = netip.AddrFromSlice(data[8:24])
+		fp.RemoteAddr, _ = netip.AddrFromSlice(data[24:40])
+	}
+
+	//TODO: whats a reasonable number of extension headers to attempt to parse?
```
nbrownus, line 312: Resolve TODO

outside.go
```diff
@@ -302,22 +281,122 @@ func (f *Interface) handleEncrypted(ci *ConnectionState, addr netip.AddrPort, h
...
+	for i := 0; i < 24; i++ {
+		if dataLen < offset {
+			break
+		}
+
+		proto := layers.IPProtocol(data[protoAt])
+		//fmt.Println(proto, protoAt)
+		switch proto {
+		case layers.IPProtocolICMPv6:
+			//TODO: we need a new protocol in config language "icmpv6"
```
nbrownus, line 325: Resolve TODO

outside.go
```diff
@@ -302,22 +281,122 @@ func (f *Interface) handleEncrypted(ci *ConnectionState, addr netip.AddrPort, h
...
+				fp.LocalPort = binary.BigEndian.Uint16(data[offset+2 : offset+4])
+			} else {
+				fp.LocalPort = binary.BigEndian.Uint16(data[offset : offset+2])
+				fp.RemotePort = binary.BigEndian.Uint16(data[offset+2 : offset+4])
+			}
+			fp.Fragment = false
+			return nil
+
+		case layers.IPProtocolIPv6Fragment:
+			//TODO: can we determine the protocol?
```
nbrownus, line 363: Resolve TODO

overlay/tun_darwin.go
```diff
@@ -297,8 +224,89 @@ func (t *tun) Activate() error {
...
+			Len:    unix.SizeofSockaddrInet6,
+			Family: unix.AF_INET6,
+			Addr:   prefixToMask(network).As16(),
+		},
+		Lifetime: addrLifetime{
+			// never expires
+			Vltime: 0xffffffff,
+			Pltime: 0xffffffff,
+		},
+		//TODO: should we disable DAD (duplicate address detection) and mark this as a secured address?
```
nbrownus, line 297: Resolve TODO

overlay/tun_linux.go
```diff
@@ -265,22 +265,67 @@ func (t *tun) deviceBytes() (o [16]byte) {
...
+				IP:   t.vpnNetworks[i].Addr().AsSlice(),
+				Mask: net.CIDRMask(t.vpnNetworks[i].Bits(), t.vpnNetworks[i].Addr().BitLen()),
+			},
+			Label: t.vpnNetworks[i].Addr().Zone(),
+		}
+	}
+
+	//add all new addresses
+	for i := range newAddrs {
+		//todo do we want to stack errors and try as many ops as possible?
```
nbrownus, line 292: Resolve TODO

overlay/tun_linux.go
```diff
@@ -324,32 +357,35 @@ func (t *tun) Activate() error {
...
 	if err = t.addRoutes(false); err != nil {
 		return err
 	}
 
-	// Run the interface
-	ifrf.Flags = ifrf.Flags | unix.IFF_UP | unix.IFF_RUNNING
-	if err = ioctl(t.ioctlFd, unix.SIOCSIFFLAGS, uintptr(unsafe.Pointer(&ifrf))); err != nil {
-		return fmt.Errorf("failed to run tun device: %s", err)
-	}
+	//todo do we want to keep the link-local address?
```
nbrownus, line 388: Resolve TODO

overlay/tun_netbsd.go
```diff
@@ -213,7 +223,8 @@ func (t *tun) removeRoutes(routes []Route) error {
 			continue
 		}
 
-		cmd := exec.Command("/sbin/route", "-n", "delete", "-net", r.Cidr.String(), t.cidr.Addr().String())
+		//todo is this right?
```
nbrownus, line 226: Resolve TODO

overlay/tun_openbsd.go
```diff
@@ -181,8 +191,8 @@ func (t *tun) removeRoutes(routes []Route) error {
 		if !r.Install {
 			continue
 		}
-
-		cmd := exec.Command("/sbin/route", "-n", "delete", "-inet", r.Cidr.String(), t.cidr.Addr().String())
+		//todo is this right?
```
nbrownus, line 194: Resolve TODO

ssh.go
```diff
@@ -785,7 +785,8 @@ func sshPrintCert(ifce *Interface, fs interface{}, a []string, w sshd.StringWrit
 		return nil
 	}
 
-	cert := ifce.pki.GetCertState().Certificate
+	//TODO: This should return both certs
```
nbrownus, line 788: Resolve TODO

relay_manager.go
```diff
@@ -91,40 +92,60 @@ func AddRelay(l *logrus.Logger, relayHostInfo *HostInfo, hm *HostMap, vpnIp neti
 func (rm *relayManager) EstablishRelay(relayHostInfo *HostInfo, m *NebulaControl) (*Relay, error) {
 	relay, ok := relayHostInfo.relayState.CompleteRelayByIdx(m.InitiatorRelayIndex, m.ResponderRelayIndex)
 	if !ok {
-		rm.l.WithFields(logrus.Fields{"relay": relayHostInfo.vpnIp,
+		//TODO: we need to handle possibly logging deprecated fields as well
```
nbrownus, line 95: Resolve TODO

relay_manager.go
```diff
@@ -91,40 +92,60 @@ func AddRelay(l *logrus.Logger, relayHostInfo *HostInfo, hm *HostMap, vpnIp neti
...
+	if err != nil {
+		h.logger(f.l).WithError(err).Error("Failed to unmarshal control message")
+		return
+	}
+
+	var v cert.Version
+	if msg.OldRelayFromAddr > 0 || msg.OldRelayToAddr > 0 {
+		v = cert.Version1
+
+		//TODO: yeah this is junk but maybe its less junky than the other options
```
nbrownus, line 118: Resolve TODO

relay_manager.go
```diff
@@ -136,68 +157,79 @@ func (rm *relayManager) handleCreateRelayResponse(h *HostInfo, f *Interface, m *
...
 			ResponderRelayIndex: peerRelay.LocalIndex,
 			InitiatorRelayIndex: peerRelay.RemoteIndex,
-			RelayFromIp:         binary.BigEndian.Uint32(b[:]),
-			RelayToIp:           uint32(target),
 		}
+
+		if v == cert.Version1 {
+			peer := peerHostInfo.vpnAddrs[0]
+			if !peer.Is4() {
+				//TODO: log cant do it
```
nbrownus, line 181: Resolve TODO

relay_manager.go
```diff
@@ -310,11 +357,11 @@ func (rm *relayManager) handleCreateRelayRequest(h *HostInfo, f *Interface, m *N
 				f.SendMessageToHostInfo(header.Control, 0, peer, msg, make([]byte, 12), make([]byte, mtu))
 				rm.l.WithFields(logrus.Fields{
 					//TODO: IPV6-WORK another lazy used to use the req object
```
nbrownus, line 359: Resolve TODO

relay_manager.go
```diff
@@ -342,16 +389,28 @@ func (rm *relayManager) handleCreateRelayRequest(h *HostInfo, f *Interface, m *N
...
 					Type:                NebulaControl_CreateRelayResponse,
 					ResponderRelayIndex: relay.LocalIndex,
 					InitiatorRelayIndex: relay.RemoteIndex,
-					RelayFromIp:         binary.BigEndian.Uint32(fromB[:]),
-					RelayToIp:           binary.BigEndian.Uint32(targetB[:]),
 				}
+
+				if v == cert.Version1 {
+					if !h.vpnAddrs[0].Is4() {
+						//TODO: log it
```
nbrownus, line 401: Resolve TODO

relay_manager.go
```diff
@@ -360,11 +419,11 @@ func (rm *relayManager) handleCreateRelayRequest(h *HostInfo, f *Interface, m *N
 					f.SendMessageToHostInfo(header.Control, 0, h, msg, make([]byte, 12), make([]byte, mtu))
 					rm.l.WithFields(logrus.Fields{
 						//TODO: IPV6-WORK more lazy, used to use resp object
```
nbrownus, line 421: Resolve TODO

pki.go
```diff
@@ -46,16 +62,26 @@ func NewPKIFromConfig(l *logrus.Logger, c *config.C) (*PKI, error) {
...
+	return p.caPool.Load()
+}
+
+func (p *PKI) getCertState() *CertState {
 	return p.cs.Load()
 }
 
-func (p *PKI) GetCAPool() *cert.CAPool {
-	return p.caPool.Load()
+// TODO: We should remove this
```
nbrownus, line 73: Resolve TODO

pki.go
```diff
@@ -46,16 +62,26 @@ func NewPKIFromConfig(l *logrus.Logger, c *config.C) (*PKI, error) {
...
+func (p *PKI) getDefaultCertificate() cert.Certificate {
+	return p.cs.Load().GetDefaultCertificate()
+}
+
+// TODO: We should remove this
```
nbrownus, line 78: Resolve TODO

pki.go
```diff
@@ -74,33 +100,94 @@ func (p *PKI) reload(c *config.C, initial bool) error {
...
+			if currentState.v1Cert.Curve() != newState.v1Cert.Curve() {
+				return util.NewContextualError(
+					"Curve in new cert was different from old",
+					m{"new_curve": newState.v1Cert.Curve(), "old_curve": currentState.v1Cert.Curve()},
+					nil,
+				)
+			}
+
+		} else if currentState.v1Cert != nil {
+			//TODO: we should be able to tear this down
```
nbrownus, line 134: Resolve TODO

pki.go
```diff
@@ -74,33 +100,94 @@ func (p *PKI) reload(c *config.C, initial bool) error {
...
+				m{"cipher": newState.cipher},
 				nil,
 			)
 		}
 	}
 
-	p.cs.Store(cs)
+	p.cs.Store(newState)
+
+	//TODO: newState needs a stringer that does json
```
nbrownus, line 186: Resolve TODO

pki.go
```diff
@@ -198,24 +295,198 @@ func newCertStateFromConfig(c *config.C) (*CertState, error) {
 		}
 	}
 
-	nebulaCert, _, err := cert.UnmarshalCertificateFromPEM(rawCert)
+	var crt, v1, v2 cert.Certificate
+	for {
+		// Load the certificate
+		crt, rawCert, err = loadCertificate(rawCert)
+		if err != nil {
+			//TODO: check error
```
nbrownus, line 303: Resolve TODO

pki.go
```diff
@@ -198,24 +295,198 @@ func newCertStateFromConfig(c *config.C) (*CertState, error) {
...
+	if v1 != nil && v2 != nil {
+		if !slices.Equal(v1.PublicKey(), v2.PublicKey()) {
+			return nil, util.NewContextualError("v1 and v2 public keys are not the same, ignoring", nil, nil)
+		}
+
+		if v1.Curve() != v2.Curve() {
+			return nil, util.NewContextualError("v1 and v2 curve are not the same, ignoring", nil, nil)
+		}
+
+		//TODO: make sure v2 has v1s address
```
nbrownus, line 373: Resolve TODO

overlay/tun_windows.go
```diff
@@ -4,41 +4,267 @@
...
+	return nil, fmt.Errorf("TODO: multiqueue not implemented for windows")
+}
+
+func (t *winTun) Close() error {
+	// It seems that the Windows networking stack doesn't like it when we destroy interfaces that have active routes,
+	// so to be certain, just remove everything before destroying.
+	luid := winipcfg.LUID(t.tun.LUID())
+	_ = luid.FlushRoutes(windows.AF_INET)
+	_ = luid.FlushIPAddresses(windows.AF_INET)
+	/* We don't support IPV6 yet
```
nbrownus, line 242: We do now

cert/cert_v2.go
JackDoanRivian: Panics suck. It'd be nice to remove `error` from the return type, but we have an `err` to bubble up from sha-ing anyway.

cert/cert_v2.go
JackDoanRivian: IMO don't panic, return `false`. If you have nil details, your signature cannot be valid.

cert/cert_v2.go
JackDoanRivian: Error please :D

cert/cert_v2.go
JackDoanRivian: lgtm -- if you _really_ wanted, you could replace `1` with `len(byte(c.curve))` but that's silly

overlay/tun_linux.go
JackDoanRivian: IMO we shouldn't touch it, and we should nudge people to set the sysctl to disable it on their nebula interfaces

cert/pem.go
nbrownus: I agree its a change but we both agree on this Slack huddle its a good one :D

cert/sign.go
JackDoanRivian: yes, sig must be at least non-0 length. Returning blank sig, no err is a common pkcs11 failure mode

cert/sign.go
JackDoanRivian: Addressed in #1266

cmd/nebula-cert/sign.go
JackDoanRivian: Addressed in #1266

cert/cert_v2.go
JackDoan: Yes we do

lighthouse.go
JackDoan: gotta be error, letting clients make the lighthouse panic is silly IMO

overlay/tun_linux.go
JackDoan:
```
FATA failed to set default route MTU: failed to set mtu 1300 on the default route fc00::/64; invalid argument
```

overlay/tun_linux.go
JackDoanRivian: removed in #1268

overlay/tun_linux.go
JackDoanRivian: resolved in #1268

cmd/nebula-cert/sign.go
JackDoanRivian: also done in #1266

lighthouse.go
```diff
@@ -971,21 +1101,51 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, vpnIp netip.Addr, a
...
+			useVersion = lhh.lh.ifce.GetCertState().defaultVersion
+		} else {
+			crt := targetHI.GetCert().Certificate
+			useVersion = crt.Version()
+			// we can only retarget if we have a hostinfo
+			newDest, ok := findNetworkUnion(crt.Networks(), fromVpnAddrs)
+			if ok {
+				whereToPunch = newDest
+			} else {
+				//TODO this means the destination will have no addresses in common with the punch-ee
```
nbrownus, line 1132: Resolve

connection_manager.go
nbrownus: Actually `hostinfo.vpnAddrs` should only contain matching addrs, so this code should be fine

Note, we need to make `hostinfo.vpnAddrs` only contain matching addres

overlay/tun_darwin.go
JackDoanRivian: at least on Linux, my unsubstantiated opinion is that nebula should _not_ do this, but our docs should suggest appropriate sysctls and systemd-networkd settings.

JackDoanRivian:
TODO: `nebula-cert verify` only checks the first cert

maggie44:
This PR no longer exports the certificate structs; neither v1 or v2:

```
type certificateV2 struct {
	details detailsV2

	// RawDetails contains the entire asn.1 DER encoded Details struct
	// This is to benefit forwards compatibility in signature checking.
	// signature(RawDetails + Curve + PublicKey) == Signature
	rawDetails []byte
	curve      Curve
	publicKey  []byte
	signature  []byte
}

type detailsV2 struct {
	name           string
	networks       []netip.Prefix
	unsafeNetworks []netip.Prefix
	groups         []string
	isCA           bool
	notBefore      time.Time
	notAfter       time.Time
	issuer         string
}
```

```
type certificateV1 struct {
	details   detailsV1
	signature []byte
}

type detailsV1 struct {
	name           string
	networks       []netip.Prefix
	unsafeNetworks []netip.Prefix
	groups         []string
	notBefore      time.Time
	notAfter       time.Time
	publicKey      []byte
	isCA           bool
	issuer         string

	curve Curve
}
```

Before:

```

type NebulaCertificate struct {
	Details   NebulaCertificateDetails
	Signature []byte

	// the cached hex string of the calculated sha256sum
	// for VerifyWithCache
	sha256sum atomic.Pointer[string]

	// the cached public key bytes if they were verified as the signer
	// for VerifyWithCache
	signatureVerified atomic.Pointer[[]byte]
}

type NebulaCertificateDetails struct {
	Name      string
	Ips       []*net.IPNet
	Subnets   []*net.IPNet
	Groups    []string
	NotBefore time.Time
	NotAfter  time.Time
	PublicKey []byte
	IsCA      bool
	Issuer    string

	// Map of groups for faster lookup
	InvertedGroups map[string]struct{}

	Curve Curve
}
```

It is going to be a big breaking change and inconvenience for anyone who has used and uses the libraries. Would prefer to keep these exported.

maggie44:
I see now it’s switched to a factory approach, with TBSCertificate that is exported. 👍 

A few thoughts to consider:

A helper function for converting a Certificate back to a TBSCertificate would be useful to complete the factory loop and allow manipulation of existing signed certificates (as unsigned ones). 

There are also some Marshal functions like Marshal(), but without a corresponding Unmarshal(). I imagine most people are using pem encoded certs, but if Marshal() is exposed there should probably be a corresponding Unmarshal().

cert/cert_v2.go
```diff
@@ -0,0 +1,655 @@
...
+	rawDetails []byte
+	curve      Curve
+	publicKey  []byte
+	signature  []byte
+}
+
+type detailsV2 struct {
+	name           string
+	networks       []netip.Prefix
+	unsafeNetworks []netip.Prefix
```
nbdd0121, line 69: I mentioned this as a reply to a resolved comment already, but in case it's folded and not seen: I think the name `networks` and `unsafeNetworks` are confusing. The former one is an address + its on-link network prefix, while the latter is just a network prefix. For the former, only a single address is routable to the host, while all addresses in the network is routable to the host for the latter.

I would propose the name to be `addresses` (I think it's fine even though this expects an CIDR, as the input is same as what `ip address add` command expects) which clearly indicate that the address part of the network prefix is significant, and `routable_networks` (or maybe `unsafe_routable_networks`) which clearly indicates that these are CIDRs for routing only.

johnmaguire: I share concerns of confusion with the naming here. I'm also comfortable sticking with the old name which at least creates no net new confusion: https://github.com/slackhq/nebula/pull/1216#discussion_r1759131003

JackDoan: I'd also be open to the name `addresses` for the `networks` field. I think the name `networks` came up when we were considering it in the context of a CA certificate? Using that same field name for two very-similar-but-different purposes muddies the waters a little bit, but the intent was that it shows which _networks_ (as V2 certs let you have more than one!) the host participates in, as well as the actual address assignment on non-CA certs. However, the `ip address add` argument is very persuasive and probably much more obvious to new users.

Naming `unsafeNetworks` is a little trickier. When the `addresses` field is named `networks`, I think it's a pretty logical name, but `unsafeAddresses` doesn't really work. I'd propose something like `unsafeRouterFor`: the prefix `unsafe` links it mentally to `unsafe_routes` in the config, and I think `RouterFor` shows that the field expresses a capability, rather than something that might affect how "regular" Nebula traffic is routed.

maggie44: I think addresses places emphasis on the use of the ip; networks on the prefix use. But ultimately both are used and the content passed is CIDRs? Perhaps better to reflect what it is and what is passed in by the user rather than the things it’s used for (similar to before where it was called Ips). 

The bigger challenge before was having to manually change the ip when using ParseCIDR because it returned the base address instead of the provided IP, and the confusion of the certificate field ‘subnets’. Both of those are no longer an issue so bounds better.

nbdd0121: As Jack said, the issue is that the field is used for dual purposes. For a non-CA cert, `addresses` and `unsafeRouterFor` makes more sense, but for CA cert, `networks` and `unsafeNetworks` make more sense. I would say it makes more sense to cater for the non-CA cert case since there're more of them, or perhaps they can have different names for different nebula-cert commands?

control.go
nbrownus: Closed by #1319

hostmap.go
```diff
@@ -335,7 +323,9 @@ func (hm *HostMap) MakePrimary(hostinfo *HostInfo) {
 }
 
 func (hm *HostMap) unlockedMakePrimary(hostinfo *HostInfo) {
-	oldHostinfo := hm.Hosts[hostinfo.vpnIp]
+	//TODO: we may need to promote follow on hostinfos from these vpnAddrs as well since their oldHostinfo might not be the same as this one
```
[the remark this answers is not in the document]
nbrownus: Closed by #1319

PR #1224, opened 2024-09-20
Cert v2 + tun changes for Linux

JackDoanRivian:


overlay/tun_linux.go
```diff
@@ -265,22 +265,70 @@ func (t *tun) deviceBytes() (o [16]byte) {
...
-	var addr, mask [4]byte
-
-	//TODO: IPV6-WORK
-	addr = t.cidr.Addr().As4()
-	tmask := net.CIDRMask(t.cidr.Bits(), 32)
-	copy(mask[:], tmask)
-
 	s, err := unix.Socket(
-		unix.AF_INET,
+		unix.AF_INET, //todo do we ever need INET6 here?
```
nbrownus, line 331: in bsd land this needs to match the af of the routes you are trying to set, 6 for v6 and 4 for v4.

overlay/tun_linux.go
```diff
@@ -350,6 +380,7 @@ func (t *tun) Activate() error {
 	if err = ioctl(t.ioctlFd, unix.SIOCSIFFLAGS, uintptr(unsafe.Pointer(&ifrf))); err != nil {
 		return fmt.Errorf("failed to run tun device: %s", err)
 	}
+	//todo do we want to keep the link-local address?
```
nbrownus, line 383: Ideally no but in bsd land I could not find a way to avoid the auto configuration so we'd have to remove it after adding the first v6 address to the network device.

cert/cert_v2.go
```diff
@@ -231,7 +231,16 @@ func (c *certificateV2) MarshalForHandshakes() ([]byte, error) {
 		//TODO: panic on nil rawDetails
 		b.AddBytes(c.rawDetails)
 
-		// Skipping the curve and public key since those come across in a different part of the handshake
+		// Skipping public key since those come across in a different part of the handshake
+
+		//todo is curve skippable? I don't think so?
+
+		// Add the curve only if its not the default value
```
nbrownus, line 238: We can transmit the byte but we will need to assert the curve we used in handshaking matches the curve present here.

The intent of skipping these top level fields was to reassemble them from noise handshake state since they are required at that level rather than duplicating them in the packet.

overlay/tun_linux.go
```diff
@@ -265,22 +265,70 @@ func (t *tun) deviceBytes() (o [16]byte) {
...
+		}
+	}
+	return false
+}
+
+// addIPs uses netlink to add all addresses that don't exist, then it removes ones that should not be there
+func (t *tun) addIPs(link netlink.Link) error {
+	newAddrs := make([]*netlink.Addr, len(t.vpnNetworks))
+	for i := range t.vpnNetworks {
+		//todo I wish I didn't need to stringify and re-parse
```
nbrownus, line 281: Might be worth looking at `"golang.org/x/net/route"` instead of the netlink abstraction we are using today for linux

cert/cert_v2.go
JackDoanRivian: Took a stab at this, lmk if it's what you had in mind.

overlay/tun_linux.go
JackDoanRivian: Yep that's what I ended up doing

overlay/tun_linux.go
JackDoanRivian: In Linux-land we use netlink for that, and everything else is AF-independent. I left a comment in case that changes.

overlay/tun_water_windows.go
```diff
@@ -52,11 +52,13 @@ func newWaterTun(c *config.C, l *logrus.Logger, cidr netip.Prefix, _ bool) (*wat
 
 func (t *waterTun) Activate() error {
 	var err error
+	//TODO multi-ip support
```
JackDoanRivian, line 55: this is a very ugly hack

overlay/tun_openbsd.go
```diff
@@ -160,8 +170,8 @@ func (t *tun) addRoutes(logErrors bool) error {
 			// We don't allow route MTUs so only install routes with a via
 			continue
 		}
-
-		cmd := exec.Command("/sbin/route", "-n", "add", "-inet", r.Cidr.String(), t.cidr.Addr().String())
+		//todo is this right?
```
JackDoanRivian, line 173: is it okay to only use the zeroth IP for this on BSDs?

lighthouse.go
```diff
@@ -1170,6 +1170,9 @@ func (lhh *LightHouseHandler) handleHostUpdateNotification(n *NebulaMeta, vpnAdd
 		useVersion = 2
 	}
 
+	//todo hosts with only v2 certs cannot provide their ipv6 addr when contacting the lighthouse via v4?
+	//todo why do we care about the vpnip in the packet? We know where it came from, right?
```
nbrownus, line 1174: Its a great question and I have it in my internals notes as well as ditching the V4AddrPort in v2 protocol stuff.

lighthouse.go
```diff
@@ -1170,6 +1170,9 @@ func (lhh *LightHouseHandler) handleHostUpdateNotification(n *NebulaMeta, vpnAdd
 		useVersion = 2
 	}
 
+	//todo hosts with only v2 certs cannot provide their ipv6 addr when contacting the lighthouse via v4?
```
nbrownus, line 1173: They should default to protocol v2, are you seeing something I'm missing?

overlay/tun_water_windows.go
nbrownus: We can probably safely remove the fallback windows tap driver support now.

overlay/tun_openbsd.go
nbrownus: I am fine to hand wave non darwin bsds for now just to get tests passing. It looked like we may be able to reuse a lot of the darwin netlink stuff there as well.

lighthouse.go
JackDoanRivian: protocol v2 is working, this is actually a multi-IP issue in a funny hat.

If you have a LH at 10.0.0.1 and you contact it via 10.0.0.2 (who is also fc00:02), because we try to use the IP in the message instead of the hostmap, the LH will never learn the underlay IP for fc00:02, unless you also contact it via overlay-ipv6.

cert/cert_v2.go
```diff
@@ -231,7 +231,7 @@ func (c *certificateV2) MarshalForHandshakes() ([]byte, error) {
 		//TODO: panic on nil rawDetails
 		b.AddBytes(c.rawDetails)
 
-		// Skipping the curve and public key since those come across in a different part of the handshake
+		// Skipping the curve and  since those come across in a different part of the handshake
```
nbrownus, line 234: Dropped a couple of letters

lighthouse.go
JackDoanRivian: Unless of course, I have something backwards

cert/cert_v2.go
JackDoanRivian: dangit

cert/cert_v2.go
nbrownus: lgtm

overlay/tun_linux.go
nbrownus: Possibly, I hadn't dug in to see if it would work for linux yet

lighthouse.go
nbrownus: Ah gotcha. Yeah this and the hostmap don't understand follow on addresses yet. We need a final loop on the updates to point all subsequent addresses to the primary address. I have some of this code staged but its not ready just yet.

overlay/tun_linux.go
```diff
@@ -324,32 +357,35 @@ func (t *tun) Activate() error {
...
 	ifrf.Flags = ifrf.Flags | unix.IFF_UP
 	if err = ioctl(t.ioctlFd, unix.SIOCSIFFLAGS, uintptr(unsafe.Pointer(&ifrf))); err != nil {
 		return fmt.Errorf("failed to bring the tun device up: %s", err)
 	}
 
-	link, err := netlink.LinkByName(t.Device)
-	if err != nil {
-		return fmt.Errorf("failed to get tun device link: %s", err)
+	// Run the interface
+	ifrf.Flags = ifrf.Flags | unix.IFF_UP | unix.IFF_RUNNING
```
nbrownus, line 371: It's likely entirely academic but we had waited until everything else was done before `RUNNING` the device to avoid packet loss from not yet fully configured devices

nbrownus (approved):
Happy to merge this into the branch now if you are ready

JackDoanRivian:
Ship it!

PR #1233, opened 2024-10-03
Add TBSCertificate.SignWith to abstract out sources-of-signatures

JackDoanRivian:
This can probably also be rebased onto cert-interface, but I haven't tried yet

### DIFF
--- a/cert/sign.go
+++ b/cert/sign.go
@@ -42,22 +42,58 @@ type beingSignedCertificate interface {
 	setSignature([]byte) error
 }
 
+type SignerLambda func(certBytes []byte) ([]byte, error)
+
 // Sign will create a sealed certificate using details provided by the TBSCertificate as long as those
 // details do not violate constraints of the signing certificate.
 // If the TBSCertificate is a CA then signer must be nil.
-func (t *TBSCertificate) Sign(signer Certificate, curve Curve, key []byte) (Certificate, error) {
-	return t.sign(signer, curve, key, nil)
+func (t *TBSCertificate) Sign(signWith Certificate, curve Curve, key []byte) (Certificate, error) {
+	switch t.Curve {
+	case Curve_CURVE25519:
+		signer := ed25519.PrivateKey(key)
+		sp := func(certBytes []byte) ([]byte, error) {
+			sig := ed25519.Sign(signer, certBytes)
+			return sig, nil
+		}
+		return t.SignWith(signWith, curve, sp)
+	case Curve_P256:
+		signer := &ecdsa.PrivateKey{
+			PublicKey: ecdsa.PublicKey{
+				Curve: elliptic.P256(),
+			},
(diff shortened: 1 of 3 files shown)


cert/sign.go
```diff
@@ -42,22 +42,58 @@ type beingSignedCertificate interface {
...
+			hashed := sha256.Sum256(certBytes)
+			return ecdsa.SignASN1(rand.Reader, signer, hashed[:])
+		}
+		return t.SignWithPredicate(signWith, curve, sp)
+	default:
+		return nil, fmt.Errorf("invalid curve: %s", t.Curve)
+	}
 }
 
 func (t *TBSCertificate) SignPkcs11(signer Certificate, curve Curve, client *pkclient.PKClient) (Certificate, error) {
```
JackDoanRivian, line 81: I'd def be open to pulling this up a level and having it just be a predicate back where its used

cert/sign.go
```diff
@@ -42,22 +42,58 @@ type beingSignedCertificate interface {
...
 }
 
+type SignerLambda func(certBytes []byte) ([]byte, error)
+
 // Sign will create a sealed certificate using details provided by the TBSCertificate as long as those
 // details do not violate constraints of the signing certificate.
 // If the TBSCertificate is a CA then signer must be nil.
-func (t *TBSCertificate) Sign(signer Certificate, curve Curve, key []byte) (Certificate, error) {
-	return t.sign(signer, curve, key, nil)
+func (t *TBSCertificate) Sign(signWith Certificate, curve Curve, key []byte) (Certificate, error) {
```
nbrownus, line 50: I would choose whichever one reads better, `signWith` or `signer` and normalize all our signing entry points to that.

cert/sign.go
nbrownus: I think I am bought in on this idea as well. `Sign` and `SignWith` cover all use cases, ditching the special case pkcs11 means a bit more lift for implementations but also the possibility to use not P256 more easily if it was viable.

cert/sign.go
```diff
@@ -42,22 +40,45 @@ type beingSignedCertificate interface {
...
+		return t.SignWith(signer, curve, sp)
+	default:
+		return nil, fmt.Errorf("invalid curve: %s", t.Curve)
 	}
-
-	return t.sign(signer, curve, nil, client)
 }
 
-func (t *TBSCertificate) sign(signer Certificate, curve Curve, key []byte, client *pkclient.PKClient) (Certificate, error) {
+// SignWith will create a sealed certificate using details provided by the TBSCertificate as long as those
```
nbrownus, line 77: I think this comment is more appropriate on `Sign`.  I think a comment advising folks to use `Sign` unless its strictly necessary to use `SignWith` is really what belongs here since its a rather special case.

JackDoanRivian: good call, lmk if you like the new phrasing

nbrownus (approved):

PR #1242, opened 2024-10-10
Cert V2 firewall fixes

JackDoanRivian:
1. Handle all ICMPv6 traffic with identical rules to ICMP
2. Correct intepretation of port numbers on UDP traffic wrt incoming vs outgoing (I'd like to add a test for this as well, but there's no .Marshal() for IPv6 headers)


nbrownus (approved):

nbrownus:
blerg, wanna `go fmt` rq

PR #1243, opened 2024-10-10
fix nebula-cert json printing

JackDoanRivian:
Before it would print like this:
```
{cert}
{cert}
```

Now it prints like this, so tools that expect json can work with it. This technically breaks the stdout-contract, but so does printing more than one cert:

```
[
  {cert},
  {cert}
]


nbrownus (approved):
`print` always outputs json with the latest change and I think it's fine that we leave the non `-json` print as is, without the array.

PR #1251, opened 2024-10-11
cert-v2: more lighthousing

JackDoanRivian:


lighthouse.go
```diff
@@ -869,47 +868,76 @@ func (lh *LightHouse) SendUpdate() {
...
+	for lhVpnAddr := range lighthouses {
+		hi := lh.ifce.GetHostInfo(lhVpnAddr)
+		if hi != nil {
+			v = hi.ConnectionState.myCert.Version()
+		} else {
+			v = lh.ifce.GetCertState().defaultVersion
 		}
-		msg.Details.RelayVpnAddrs = relays
-		msg.Details.VpnAddr = netAddrToProtoAddr(lh.myVpnNetworks[0].Addr())
+		if v == cert.Version1 {
```
nbrownus, line 887: Since we are reusing the `msg` struct we need to zero out the v2 fields in v1 and v1 fields in v2.

nbrownus (approved):
I like this feels safer than my crazy version, I should go back and update the lh query stuff to match

PR #1252, opened 2024-10-15
cert-v2 relay stuff

JackDoanRivian:
Because hosts can have more than one IP now, we need to query the whole list of relay-fors to handle this flow:

1. Host A -> `CreateRelayRequest(HostA, (HostB's non-primary IP))` -> Relay
2. _things happen to set the relay up_
3. Host B -> replies to Host A's handshake via Relay
4. Relay has the following relay-tionships:
    - (Host A Primary) -> (Host B secondary)
    - (Host B Primary) -> (Host A Primary)
5. in order for this reply to make it back to Host A, we need to look at all the relays on Relay, not just the ones that exactly match the primary address
    - Said another way, if you only need to relay to node's primary IPs, this isn't a problem. 
    - But, if you only "know" a non-primary IP, you have no way to discover the primary IP without a handshake, which you may need a functional relay to get.

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -456,18 +456,19 @@ func (hm *HostMap) QueryVpnAddrsRelayFor(targetIps []netip.Addr, relayHostIp net
 	hm.RLock()
 	defer hm.RUnlock()
 
-	for _, targetIp := range targetIps {
-		h, ok := hm.Hosts[relayHostIp]
-		if !ok {
-			return nil, nil, errors.New("unable to find host")
-		}
-		for h != nil {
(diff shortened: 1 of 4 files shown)


brad-defined:
I think this PR looks great - it meets the original intent of the lookup with the new concept of multiple ip's per host. We could do some optimizations, but would need to profile it to see what real impacts it would have.

The IP lookup indirection in relaying was simple before rehandshaking introduced multiple hostinfos, and the multiple IP's further complicates the picture. I don't want to hold up this PR that at least makes relays work with multiple IP's for a general relay re-think though.

nbrownus (approved):

Issue #1124, opened 2024-04-19
🐛 BUG: tests fail after 2027-11-11


bmwiedemann:
### What version of `nebula` are you using? (`nebula -version`)

1.8.2

### What operating system are you using?

openSUSE Tumbleweed

### Describe the Bug

While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that
our nebula-1.8.2 package fails build-tests after 2027-11-11

To reproduce on Debian or openSUSE, use
osc checkout openSUSE:Factory/nebula && cd $_
osc build --no-service --vm-type=kvm --build-opt=--vm-custom-opt=-rtc\ base=2027-11-12T00:00:00 standard

or set the system time instead.

Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +16 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our package build.

### Logs from affected hosts

build log (from year 2040):
```
 === RUN   TestNewCAPoolFromBytes
     cert_test.go:596: 
                Error Trace:    /home/abuild/rpmbuild/BUILD/nebula-1.8.2/cert/cert_test.go:596
                Error:          Expected nil, but got: &errors.errorString{s:"certificate is expired"}
                Test:           TestNewCAPoolFromBytes
     cert_test.go:601: 
                Error Trace:    /home/abuild/rpmbuild/BUILD/nebula-1.8.2/cert/cert_test.go:601
                Error:          Expected nil, but got: &errors.errorString{s:"certificate is expired"}
                Test:           TestNewCAPoolFromBytes
     cert_test.go:619: 
                Error Trace:    /home/abuild/rpmbuild/BUILD/nebula-1.8.2/cert/cert_test.go:619
                Error:          Expected nil, but got: &errors.errorString{s:"certificate is expired"}
                Test:           TestNewCAPoolFromBytes
 --- FAIL: TestNewCAPoolFromBytes (0.00s) 
 === RUN   TestUnmrshalCertPEM
```

### Config files from affected hosts

```

```

johnmaguire:
We should be able to update the tests with new certs with longer expirations (say 100 or 1000 years.)

[...]

PR #1253, opened 2024-10-18
Fixup cert package tests

nbrownus:
This change fixes the old v1 tests and adds new tests for v2. 

Most notably it fixes the `TestMarshalingCertificateV1Consistency` test which needed to be modified. `netip.Prefix` can not represent netmasks like `255.0.255.0`. I went back to f22b4b5 and modified the test details to remove the non conforming netmasks and adjust the timestamps, the resulting expected byte arrays are represented in this test.

There are two spots where I am unsure if we want to panic, v1 and v2 `String()` calls `json.Marshal` which can fail. Currently they return a string but panic would be more obvious.


cert/test_helpers.go
```diff
@@ -1,28 +1,50 @@
-package e2e
+package cert
+
+//TODO: this should only be built if performing tests
```
nbrownus, line 3: I really dont want to require we use a tag to execute tests but it also feels bad to include this in the binary and I don't know of another way to deal with this without just copying these functions all over the place again.

JackDoanRivian: can't you define these in a `something_helpers_test.go` and use them in other `*_test.go` files?

nbrownus: I desperately want there to be a way but the only thing I can come up with is a build tag much like how `e2e` works

johnmaguire:
Could knock out https://github.com/slackhq/nebula/issues/1124 in this too.

PR #1262, opened 2024-11-01
[cert-v2] support compressed P256 public keys

JackDoanRivian:
Despite it being clunky in Go, supporting both compressed and uncompressed P256 public keys can help us shave 32 bytes from P256 certs. This PR adds _support_ for compressed keys, but does not yet expose it to users via nebula-cert.


wadey (approved):

JackDoanRivian:
There are problems here wrt handshakes and how we send the pubkey over "early"

PR #1266, opened 2024-11-05
enforce certificate correctness in TBSCertificate.SignWith

JackDoanRivian:


cert/sign.go
```diff
@@ -76,15 +76,89 @@ func (t *TBSCertificate) Sign(signer Certificate, curve Curve, key []byte) (Cert
 	}
 }
 
+// readyToSign checks all signing requirements that don't require us to cross-reference with a CA
+func (t *TBSCertificate) readyToSign() error {
+	if t.PublicKey == nil {
```
JackDoanRivian, line 81:
```suggestion
	if len(t.PublicKey) == 0 {
```

cert/sign.go
```diff
@@ -76,15 +76,89 @@ func (t *TBSCertificate) Sign(signer Certificate, curve Curve, key []byte) (Cert
...
+	}
+
+	for _, n := range t.UnsafeNetworks {
+		if !n.IsValid() || !n.Addr().IsValid() {
+			return fmt.Errorf("invalid unsafe network: %s", n)
+		}
+		if n.Addr().Zone() != "" {
+			return fmt.Errorf("unsafe_networks may not contain zones: %s", n)
+		}
+		//todo are unsafe networks that overlap networks allowed?
```
nbrownus, line 125: Yes, you might have `10.0.0.0/8` point to one host but `10.0.0.0/16` to another, it makes less sense to have overlapping in a CA certificate but that's also fine.

cert/sign.go
```diff
@@ -76,15 +76,89 @@ func (t *TBSCertificate) Sign(signer Certificate, curve Curve, key []byte) (Cert
 	}
 }
 
+// readyToSign checks all signing requirements that don't require us to cross-reference with a CA
+func (t *TBSCertificate) readyToSign() error {
```
nbrownus, line 80: I had started on this validation in `cert_v1.go` and `cert_v2.go` specifically that way we don't accidentally leak rules from one version to the other that shouldn't apply. We also need to do these things at unmarshal (or at a minimum at handshake and cert loading time) to ensure we are working with a valid certificate.

cert/sign.go
```diff
@@ -76,15 +76,89 @@ func (t *TBSCertificate) Sign(signer Certificate, curve Curve, key []byte) (Cert
...
+		if n.Addr().Zone() != "" {
+			return fmt.Errorf("unsafe_networks may not contain zones: %s", n)
+		}
+		//todo are unsafe networks that overlap networks allowed?
+
+		if n.Addr().Is6() {
+			if t.Version == Version1 {
+				return fmt.Errorf("certificate v1 may not contain IPv6 unsafe networks: %v", t.Networks)
+			}
+			if !hasV6Networks && !t.IsCA {
```
nbrownus, line 131: I don't disagree but I am curious if this will haunt us in the future

cert/sign.go
```diff
@@ -76,15 +76,89 @@ func (t *TBSCertificate) Sign(signer Certificate, curve Curve, key []byte) (Cert
...
+				return fmt.Errorf("IPv6 unsafe networks require an IPv6 address assignment")
+			}
+		} else if n.Addr().Is4() {
+			if !hasV4Networks && !t.IsCA {
```
nbrownus, line 135: I don't disagree but I am curious if this will haunt us in the future

cert/sign.go
JackDoanRivian: ah yeah nice catch, I'll move this

PR #1267, opened 2024-11-05
address cert-v2 lighthouse TODOs

JackDoanRivian:


lighthouse.go
```diff
@@ -1299,12 +1283,10 @@ func (lhh *LightHouseHandler) handleHostUpdateNotification(n *NebulaMeta, fromVp
 		}
 		vpnAddrB := fromVpnAddrs[0].As4()
 		n.Details.OldVpnAddr = binary.BigEndian.Uint32(vpnAddrB[:])
-
 	} else if useVersion == cert.Version2 {
 		n.Details.VpnAddr = netAddrToProtoAddr(fromVpnAddrs[0])
-
 	} else {
-		panic("unsupported version")
+		panic("unsupported version") //todo need a test that proves we can never get here so we can remove this
```
nbrownus, line 1289: We can just log an error and return like we do with the marshal failure below

lighthouse.go
```diff
@@ -1318,7 +1300,7 @@ func (lhh *LightHouseHandler) handleHostUpdateNotification(n *NebulaMeta, fromVp
 }
 
 func (lhh *LightHouseHandler) handleHostPunchNotification(n *NebulaMeta, fromVpnAddrs []netip.Addr, w EncWriter) {
-	//TODO: this is kinda stupid
+	//TODO: this is kinda stupid (WHY?)
```
nbrownus, line 1303: I was frustrated that its possible the lighthouse is communicating with us using a non primary vpn addr which meant we need to compare all `fromVpnAddrs` against all configured lighthouse vpn addrs.

We can remove it, mainly a performance reminder and it'll light up on a cpu profile to remind us.

nbrownus (approved):

PR #1268, opened 2024-11-05
[cert-v2] tun_linux.go: retry setting default MTU

JackDoanRivian:


overlay/tun_linux.go
```diff
@@ -420,7 +417,20 @@ func (t *tun) setDefaultRoute(cidr netip.Prefix) error {
 	}
 	err := netlink.RouteReplace(&nr)
 	if err != nil {
-		return fmt.Errorf("failed to set mtu %v on the default route %v; %v", t.DefaultMTU, dr, err)
+		t.l.WithError(err).WithField("cidr", cidr).Warn("Failed to set default route MTU, retrying")
```
nbrownus, line 420: The ordering of `Activate` changed a good deal, have we tried moving it back to the order it was in to see if that helps this situation?

https://github.com/slackhq/nebula/blob/e264a0ff888c7bf0568579306755a60fc42f6ecc/overlay/tun_linux.go#L328-L352

JackDoanRivian: I fiddled with the ordering and it didn't seem to make a difference.

I think the problem we're seeing here is somehow caused by having more than one IP -- especially an IPv6 IP. When I test with single-IP certs I'm completely unable to reproduce the problem.

JackDoanRivian: Oh also, when this fails it _always_ fails when adding the _second_ route. The first route always works.

nbrownus (approved):

PR #1270, opened 2024-11-11
Reestablish relays when last hop fails

brad-defined:
Relays can break and never recover if, with the following setup:
I<->R<->C
the R<->C connection breaks, and the only traffic in the network is I<->C, not C->I.

This change updates host deletion work to mark associated relays as 'Disestablished', because we can't quite tell if they are still working or not. When host `I` finds its relayed connection is in this state, it will now re-send CreateRelayRequest messages to ensure the end-to-end relayed connection.

Relay host behavior was also updated. Before, the Relay host would use the State status of relay peers to immediately return a CreateRelayResponse if the next hop is already in an Established state. This PR changes the Relay to always forward CreateRelayRequest messages along the path, and only terminal relay peers will generate the CreateRelayResponse message flow.

### DIFF
--- a/e2e/handshakes_test.go
+++ b/e2e/handshakes_test.go
@@ -471,7 +471,7 @@ func TestRelays(t *testing.T) {
 	//TODO: assert we actually used the relay even though it should be impossible for a tunnel to have occurred without it
 }
 
-func TestRelaysHH(t *testing.T) {
+func TestReestablishRelays(t *testing.T) {
 	ca, _, caKey, _ := cert_test.NewTestCaCert(cert.Version1, cert.Curve_CURVE25519, time.Now(), time.Now().Add(10*time.Minute), nil, nil, []string{})
 	myControl, myVpnIpNet, _, _ := newSimpleServer(cert.Version1, ca, caKey, "me     ", "10.128.0.1/24", m{"relay": m{"use_relays": true}})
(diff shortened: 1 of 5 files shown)


johnmaguire:
As an FYI for viewers, a log line like this at the Lighthouse is indicative of this issue:

```
Nov 11 16:10:12 linux.dsrgthyz.com nebula[1147]: time="2024-11-11T16:10:12Z" level=info msg="Failed to find target host info by ip" certName=android error="unable to find host with relay" localIndex=2672096877 relayTo=192.168.100.3 remoteIndex=442491203 vpnIp=192.168.100.6
Nov 11 16:10:13 linux.dsrgthyz.com nebula[1147]: time="2024-11-11T16:10:13Z" level=info msg="Failed to find target host info by ip" certName=android error="unable to find host with relay" localIndex=2672096877 relayTo=192.168.100.3 remoteIndex=442491203 vpnIp=192.168.100.6
Nov 11 16:10:14 linux.dsrgthyz.com nebula[1147]: time="2024-11-11T16:10:14Z" level=info msg="Failed to find target host info by ip" certName=android error="unable to find host with relay" localIndex=2672096877 relayTo=192.168.100.3 remoteIndex=442491203 vpnIp=192.168.100.6
```

nbrownus (approved):
Would be nice to backport to v1.9 as well

PR #1279, opened 2024-11-25
[v1.9.x] do not panic when loading a V2 CA certificate

JackDoanRivian:
but don't try to use it either


cert/ca.go
```diff
@@ -46,6 +50,8 @@ func NewCAPoolFromBytes(caPEMs []byte) (*NebulaCAPool, error) {
 
 	if expired {
 		return pool, ErrExpired
+	} else if caTooNew {
```
nbrownus, line 53: Kind of a wild situation but we could have an expired and a too new cert and we wouldn't know about the too new one until the expired one was removed.

pki.go
```diff
@@ -237,6 +237,8 @@ func loadCAPoolFromConfig(l *logrus.Logger, c *config.C) (*cert.NebulaCAPool, er
 			return nil, errors.New("no valid CA certificates present")
 		}
 
+	} else if errors.Is(err, cert.ErrInvalidPEMCertificateUnsupported) {
```
nbrownus, line 240: We should ensure there is at least 1 valid CA in the pool otherwise it won't work

JackDoanRivian: `ErrInvalidPEMCertificateUnsupported` never get added to the pool. Do we not already have semantics that catch size-zero pools? I'll check.

cert/ca.go
nbrownus: I was thinking maybe we hold back one of the errors and if the other appears then we wrap that. Then we can check if the error appears in the chain instead of direct equality. 

It would be a bummer to log from this here IMO

pki.go
nbrownus: We do a few lines up. `caPool, err := cert.NewCAPoolFromBytes(rawCA)` will continue parsing the provided bytes in the event of a `ErrInvalidPEMCertificateUnsupported` but if all are too new (or expired) and we hit this branch then we can swap in a `CAPool` that has no valid certificates.

JackDoanRivian: We did not. Added.

cert/ca.go
JackDoanRivian: @wadey suggested `(pool CAPool, warnings []error, err error)` so I gave that a try. It feels pretty nice.

JackDoanRivian:
Superseded by !1282

PR #1319, opened 2025-01-28
Make sure all vpnAddrs are hoisted to primary, resolve a few more TODOs

nbrownus:


control.go
```diff
@@ -234,7 +234,7 @@ func (c *Control) CloseAllTunnels(excludeLighthouses bool) (closed int) {
 		c.f.send(header.CloseTunnel, 0, h.ConnectionState, h, []byte{}, make([]byte, 12, 12), make([]byte, mtu))
 		c.f.closeTunnel(h)
 
-		c.l.WithField("vpnIp", h.vpnAddrs[0]).WithField("udpAddr", h.remote).
+		c.l.WithField("vpnAddrs", h.vpnAddrs).WithField("udpAddr", h.remote).
```
jasikpark, line 237: field should now be `udpAddrs`, no?

wadey (approved):

control.go
nbrownus: `udpAddr` is for a single address and `udpAddrs` would be for multiple. We don't want to mix the types (string vs array of string) since it breaks things like elasticsearch.

jasikpark: ope i wasn't tracking how field building works here, nvm 😅

thought udpAddr was the label for the vpnAddrs variable

PR #1321, opened 2025-01-29
Fix `testifylint` lint errors

jasikpark:
https://github.com/Antonboom/testifylint contains lints for validating that you avoid pitfalls of using `testify`. 

I fixed all of the rules except for [require-error](https://github.com/Antonboom/testifylint#require-error) which asks you to use `require.Error` instead of `assert.Error` because it calls `t.FailNow()` and makes it less likely that you accidentally swallow the resulting error.

Mostly it's syntactical style and better error messages, plus fixing cases where "expected" and "actual" were switched, making error messages more confusing.

There was one case where we were testing *whether* an error was an error, rather than whether it matched the error in question, caught by [error-is-as](https://github.com/Antonboom/testifylint?tab=readme-ov-file#error-is-as)


firewall_test.go
```diff
@@ -352,7 +352,7 @@ func TestFirewall_Drop2(t *testing.T) {
 	cp := cert.NewCAPool()
 
 	// h1/c1 lacks the proper groups
-	assert.Error(t, fw.Drop(p, true, &h1, cp, nil), ErrNoMatchingRule)
+	assert.ErrorIs(t, fw.Drop(p, true, &h1, cp, nil), ErrInvalidRemoteIP)
```
jasikpark, line 355: `fw.Drop` catches the fact that `h1` is initialized w/o a valid vpnIp before it actually checks the firewall rule later.

wadey:
Needs to be re-done now that we merged the `cert-v2` branch :cry:

jasikpark:
> Needs to be re-done now that we merged the `cert-v2` branch 😢

done!

nbrownus (approved):
`JSONEq` was the most surprising, didn't know about that one. Overall testify really makes me feel like I'm reliving the old days of PHP.

It would be great to wire this up as a general lint check next to `go vet`

jasikpark:
I'll see about adding that 💜

PR #1324, opened 2025-01-31
Alternative route table and 0.0.0.0 routing

dd0ke:
Added preliminary support for the creation of routes in alternative route tables on Linux, and using default routes via Nebula.
Nebula would crash anytime a default route was added or modified, regardless of the default route being added on to a separate table.

Requires new config param: 
  # Specify the alternative routing table for created routes.
  # If unset, defaults to main routing table.
  #alternative_routing_table: "my_alt_table"

### DIFF
--- a/overlay/tun_linux.go
+++ b/overlay/tun_linux.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"net/netip"
 	"os"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"unsafe"
@@ -36,6 +37,7 @@ type tun struct {
 	routeTree       atomic.Pointer[bart.Table[netip.Addr]]
 	routeChan       chan struct{}
 	useSystemRoutes bool
+	alternativeRoutingTable string  // New field for alternative routing table
 
 	l *logrus.Logger
 }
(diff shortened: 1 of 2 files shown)


overlay/tun_linux.go
```diff
@@ -147,16 +150,58 @@ func newTunGeneric(c *config.C, l *logrus.Logger, file *os.File, cidr netip.Pref
...
+		return unix.RT_TABLE_MAIN, nil
+	}
+
+	// Try parsing as a number first
+	if tableID, err := strconv.Atoi(t.alternativeRoutingTable); err == nil {
+		return tableID, nil
+	}
+
+	// If it's not a number, look up the table name in /etc/iproute2/rt_tables
+	content, err := os.ReadFile("/etc/iproute2/rt_tables")
```
JackDoan, line 164: The location of this file seems like it could be distro-specific. Should it be configurable? Can the information instead be obtained via netlink or /proc?

johnmaguire:
Hi @dd0ke - it's been a minute, so I wanted to touch base. We recently landed #1331 which may solve your needs. If not, could you share a bit about why you still prefer this approach? Additionally, I see there is some open review feedback. Are you still working on this PR?

dd0ke:
Hello hello there,

It’s been some time since I’ve tinkered with Nebula. The topic is somewhat digressing, but here is the scope objective:

Permit default route or the use of alternative routing table to apply default route for Nebula traffic. Packet marking is related to the issue indirectly. Since alternative tables were not usable, the workaround was to mark packets to respond via the appropriate gateway.
Packet marking is still necessary, and I will test your updates, as part of the end goal is to make Nebula behave like a stateless over-the-top VPN via anycast IPs for clients and gateways. 
I have this working with some of the modifications I suggested for a PR. Basically, my Nebula clients can get public IPs on an anycast subnet, and employ the most optimal gateway for the outbound and return path, also inducing some redundancy perks in the process.
This is an improvement over how Wireguard works, as Wireguard creates persistent routes for all its peers which go to the void if the endpoint is not connected. I didn’t want to setup Wireguard with a distributed table or coordination service like Tailscale/Headscale, or over complicated implementations in the likes of Netbird. 

Will update when I can test your changes!

nbrownus:
Closing as stale

Issue #1322, opened 2025-01-30
🐛 BUG: heavy CPU usage with use_system_route_table: yes


nikitos:
### What version of `nebula` are you using? (`nebula -version`)

1.9.5

### What operating system are you using?

Linux

### Describe the Bug

after massively (about 10k routes) adding routes via ospf to routing table nebula print several (30-40 times)
"Adding route" destination=<network>/24 via=10.8.0.3 and then consume 200% CPU, all routings seems works fine.

### Logs from affected hosts

```
"Adding route" destination=<network>/24 via=10.8.0.3
"Adding route" destination=<network>/24 via=10.8.0.3
"Adding route" destination=<network>/24 via=10.8.0.3
"Adding route" destination=<network>/24 via=10.8.0.3
"Adding route" destination=<network>/24 via=10.8.0.3
"Adding route" destination=<network>/24 via=10.8.0.3
"Adding route" destination=<network>/24 via=10.8.0.3
```

### Config files from affected hosts

```
static_host_map:
  "10.8.0.1": ["<IP>:4242"]

lighthouse:
  am_lighthouse: false
  interval: 60
  hosts:
    - "10.8.0.1"
routines: 2

cipher: aes
listen:
  host: <IP>
  port: 4242
tun:
  drop_local_broadcast: false
  drop_multicast: false
  mtu: 1400
  tx_queue: 5000
  use_system_route_table: yes
  unsafe_routes:
    - route: 10.0.0.0/8
      via: 10.8.0.1
      metric: 100
      install: false
    - route: 0.0.0.0/0
      via: 10.8.0.3
      metric: 100
      install: false
firewall:
  outbound_action: drop
  inbound_action: drop

  conntrack:
    tcp_timeout: 12m
    udp_timeout: 3m
    default_timeout: 10m

  outbound:
    # Allow all outbound traffic from this node
    - port: any
      proto: any
      host: any

  inbound:
    # Allow icmp between any nebula hosts
    - port: any
      proto: any
      host: any
```
all other host has the same configs

nikitos:
after enabling debug i mentioned that this message apperars
```
Ignoring route update, not a gateway route" route="{Ifindex: 0 Dst: <nil> Src: <nil> Gw: <nil> Flags: [] Table: 0 Realm: 0}"
Ignoring route update, not a gateway route" route="{Ifindex: 0 Dst: <nil> Src: <nil> Gw: <nil> Flags: [] Table: 0 Realm: 0}"
[...]

PR #1326, opened 2025-02-07
add netlink options 

nikitos:
should fix this issue https://github.com/slackhq/nebula/issues/1322
no regression for default config
added logs from netlink


nikitos:
You already signed the CLA on 2025-02-07

overlay/tun_linux.go
```diff
@@ -488,7 +490,13 @@ func (t *tun) watchRoutes() {
 	rch := make(chan netlink.RouteUpdate)
 	doneChan := make(chan struct{})
 
-	if err := netlink.RouteSubscribe(rch, doneChan); err != nil {
+	netlinkOptions := netlink.RouteSubscribeOptions{
+		ReceiveBufferSize: t.useSystemRoutesBuffer,
+		ReceiveBufferForceSize: true,
```
JackDoan, line 495:
```suggestion
		ReceiveBufferForceSize: t.useSystemRoutesBuffer != 0,
```

I poked through the function call stack that `netlink.RouteSubscribe` makes and compared it to `netlink.RouteSubscribeOptions` and this _should_ be fine. The only difference is this always passes `true` for `ReceiveBufferForceSize`, where the default is `false`, but `ReceiveBufferForceSize` is only considered if `ReceiveBufferSize !=0`.

If we want to be really really safe and consistent with the `netlink` calls, we could take this change, but I'm not convinced it's really needed.

overlay/tun_linux.go
```diff
@@ -36,6 +36,7 @@ type tun struct {
 	routeTree       atomic.Pointer[bart.Table[netip.Addr]]
 	routeChan       chan struct{}
 	useSystemRoutes bool
+	useSystemRoutesBuffer int
```
JackDoan, line 39:
```suggestion
	useSystemRoutesBufferSize int
```

sort of a nit, but I wouldn't want folks to think this was the actual buffer

JackDoan (changes requested):
Since this adds a new config option, could you please also add it to the example config.yml, commented out, with some explanation of what it does and why you might use it?

overlay/tun_linux.go
```diff
@@ -129,6 +130,7 @@ func newTunGeneric(c *config.C, l *logrus.Logger, file *os.File, cidr netip.Pref
 		cidr:            cidr,
 		TXQueueLen:      c.GetInt("tun.tx_queue", 500),
 		useSystemRoutes: c.GetBool("tun.use_system_route_table", false),
+		useSystemRoutesBuffer: c.GetInt("tun.use_system_route_table_buffer", 0),
```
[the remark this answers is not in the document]
nikitos: zero buffer make netlink socket use default system buffer (about 200k) it was enough on my system to receive 100+ routes updates at a time. Looks like default buffer size fits most use cases and increasing is needed only on massive updates.

overlay/tun_linux.go
nikitos: fixed

overlay/tun_linux.go
nikitos: fixed this just to pass the same parameters to netlink, but netlink use ReceiveBufferForceSize only when ReceiveBufferSize !=0 too.

johnmaguire (approved):

JackDoan (approved):
lgtm!

PR #1330, opened 2025-02-13
Change allow list to evaluate all vpnaddr tables when available

nbrownus:
This changes the allow list to operate on all known overlay addresses for a remote instead of just the currently used overlay address, when possible.

Since the default is to allow all, any deny is specific intent to not communicate with a remote host on that underlay range, so to reflect that intent we need to evaluate the underlay address against each of the overlay addresses allow list tables.

There is 1 tricky spot with stage 2 handshakes where we could operate on all known addresses but that would pollute our noise state since we have to consume the packet and ratchet the state forward to get a full set of overlay addresses in the remote certificate. Currently this is not being done but we can achieve it if deemed necessary, ideally the remote side has identical configuration that would have blocked this underlay route from being used.

There is also a similar issue with the lighthouse where it operates on the overlay address reported by the node instead of all overlay addresses in the hostinfo for the reporting node. If we leave it as it is now then it allows some surgical options in the lighthouse configuration but the effect is mismatched from non lighthouse node configuration.

### DIFF
--- a/outside.go
+++ b/outside.go
@@ -231,26 +231,26 @@ func (f *Interface) sendCloseTunnel(h *HostInfo) {
 	f.send(header.CloseTunnel, 0, h.ConnectionState, h, []byte{}, make([]byte, 12, 12), make([]byte, mtu))
 }
 
-func (f *Interface) handleHostRoaming(hostinfo *HostInfo, vpnAddr netip.AddrPort) {
-	if vpnAddr.IsValid() && hostinfo.remote != vpnAddr {
-		//TODO: CERT-V2 this is weird now that we can have multiple vpn addrs
-		if !f.lightHouse.GetRemoteAllowList().Allow(hostinfo.vpnAddrs[0], vpnAddr.Addr()) {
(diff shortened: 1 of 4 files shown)


outside.go
```diff
@@ -231,26 +231,26 @@ func (f *Interface) sendCloseTunnel(h *HostInfo) {
...
 			return
 		}
 
-		hostinfo.logger(f.l).WithField("udpAddr", hostinfo.remote).WithField("newAddr", vpnAddr).
+		hostinfo.logger(f.l).WithField("udpAddr", hostinfo.remote).WithField("newAddr", udpAddr).
 			Info("Host roamed to new udp ip/port.")
 		hostinfo.lastRoam = time.Now()
 		hostinfo.lastRoamRemote = hostinfo.remote
-		hostinfo.SetRemote(vpnAddr)
+		hostinfo.SetRemote(udpAddr)
```
wadey, line 253: good catch here!

wadey (approved):
LGTM

Issue #353, opened 2020-12-22
Add unsafe_route metric / hop count


Mrc527:
Hello guys,

In my scenario, I've multiple nodes which could serve the same unsafe route.
Let's say, I want to access a TV on my local network and I've nebula nodes on my NAS and Router. Both on the same local network.

So, while being on the road, I could connect to the TV through the NAS or the Router. But since the NAS is more powerful, I get a better bandwidth if I go through it rather than the Router.

I would like to have an option in the unsafe_routes to specify a priority like:
```
unsafe_routes:
    - route: 192.168.1.0/24
      via: 192.168.99.1
      priority: 1
    - route: 192.168.1.0/24
      via: 192.168.99.2
      priority: 2
```

So, my nebula client should first try to reach 192.168.1.0 network through 192.168.99.1 and, if 192.168.99.1 is not reachable, start looking for 192.168.99.2.

How do you see it?

Ciao!
M

hulucc:
Same requirement but I want to do unsafe_routes high availability.

ton31337:
Related: https://github.com/slackhq/nebula/pull/474

wadey:
#474 was merged, so closing this issue.

dhananjays:
Is this issue really addressed by #474.
It seems like that only supports the case where the two routes are via different interfaces (eg. one via openVpn while another via nebula)

When we are trying to define both routes via nebula (similar to the routes mentioned by OP https://github.com/slackhq/nebula/issues/353#issue-772812135), we are seeing that whichever route is later in the config gets used. We are using v1.6.1

CC: @hulucc @Mrc527 Is your use-case working with 1.6.1?

dhananjays:
@wadey I think this issue isn't addressed by #474. Can you please reopen this issue?

alexandru-filip-gv:
> Is this issue really addressed by #474 ? It seems like that only supports the case where the two routes are via different interfaces (eg. one via openVpn while another via nebula)
> 
> When we are trying to define both routes via nebula (similar to the routes mentioned by OP [#353 (comment)](https://github.com/slackhq/nebula/issues/353#issue-772812135)), we are seeing that whichever route is later in the config gets used (irrespective of value of `metric`). We are using v1.6.1
> 
> CC: @hulucc @Mrc527 Is your use-case working with 1.6.1?

[...]

PR #1332, opened 2025-02-19
Implement ECMP for unsafe_routes

dioss-Machiel:
This PR implements ECMP support in Nebula. The implementation uses hash-threshold mapping (like in the Linux kernel) which allows you to define weights per gateway.

This can be used for example to aggregate multiple links and provide redundancy to an external location where nebula cannot be installed.

The change is backwards compatible and should not impact normal operation.
ECMP routes can be defined in the config file (example below) and also via the use_system_route_table feature, which means that on Linux you can use a BGP daemon to add / remove multipath routes.

Example config:

```
tun:
  unsafe_routes:
	# Backwards compatibility
	- route: 192.168.86.0/24
	  via: 192.168.100.10
	
	# Single gateway, weight optional
	- route: 192.168.87.0/24
	  via: 
		- gateway: 10.0.0.1

	# Multipath, weight optional (defaults to 1), will balance equally
	- route: 192.168.87.0/24
	  via: 
		- gateway: 10.0.0.1
		- gateway: 10.0.0.2

	# Multipath, weight defined, will balance according to weights
	- route: 192.168.87.0/24
	  via: 
		- gateway: 10.0.0.1
		  weight: 10
		  
		- gateway: 10.0.0.2
		  weight: 5
```

Implementation note: if the gateway where the packet should be routed to is not reachable the first available gateway is chosen, so when one of the gateways is down the traffic is no longer properly being balanced, but there is still connectivity.

### DIFF
--- a/inside.go
+++ b/inside.go
@@ -152,14 +152,21 @@ func (f *Interface) getOrHandshakeConsiderRouting(fwPacket *firewall.Packet, cac
 	}
 
 	gateways := f.inside.RoutesFor(destinationAddr)
-	if len(gateways) == 0 {
+
+	switch len(gateways) {
+	case 0:
 		return nil, false
-	} else if len(gateways) == 1 {
+	case 1:
 		// Single gateway route
 		return f.handshakeManager.GetOrHandshake(gateways[0].Addr(), cacheCallback)
-	} else {
+	default:
 		// Multi gateway route, perform ECMP categorization
-		gatewayIp := routing.BalancePacket(fwPacket, gateways)
+		gatewayAddr, balancingOk := routing.BalancePacket(fwPacket, gateways)
+
+		if !balancingOk {
+			// This happens if the gateway buckets were not calculated, this _should_ never happen
+			f.l.Error("Gateway buckets not calculated, fallback from ECMP to random routing. Please report this bug.")
+		}
 
 		var handshakeInfoForChosenGateway *HandshakeHostInfo
 		var hhReceiver = func(hh *HandshakeHostInfo) {
@@ -169,7 +176,7 @@ func (f *Interface) getOrHandshakeConsiderRouting(fwPacket *firewall.Packet, cac
 		// Store the handshakeHostInfo for later.
 		// If this node is not reachable we will attempt other nodes, if none are reachable we will
 		// cache the packet for this gateway.
-		if hostinfo, ready := f.handshakeManager.GetOrHandshake(gatewayIp, hhReceiver); ready {
+		if hostinfo, ready := f.handshakeManager.GetOrHandshake(gatewayAddr, hhReceiver); ready {
 			return hostinfo, true
 		}
 
@@ -182,13 +189,13 @@ func (f *Interface) getOrHandshakeConsiderRouting(fwPacket *firewall.Packet, cac
 
 		if f.l.Level >= logrus.DebugLevel {
 			f.l.WithField("destination", destinationAddr).
-				WithField("originalGateway", gatewayIp).
+				WithField("originalGateway", gatewayAddr).
 				Debugln("Calculated gateway for ECMP not available, attempting other gateways")
 		}
 
 		for i := range gateways {
 			// Skip the gateway that failed previously
-			if gateways[i].Addr() == gatewayIp {
+			if gateways[i].Addr() == gatewayAddr {
 				continue
 			}
 
@@ -202,6 +209,7 @@ func (f *Interface) getOrHandshakeConsiderRouting(fwPacket *firewall.Packet, cac
 		cacheCallback(handshakeInfoForChosenGateway)
 		return hostinfo, false
 	}
+
 }
 
 func (f *Interface) sendMessageNow(t header.MessageType, st header.MessageSubType, hostinfo *HostInfo, p, nb, out []byte) {
--- a/inside.go
+++ b/inside.go
@@ -127,13 +127,17 @@ func (f *Interface) Handshake(vpnAddr netip.Addr) {
 
 // getOrHandshake returns nil if the vpnAddr is not routable.
 // If the 2nd return var is false then the hostinfo is not ready to be used in a tunnel
+
 func (f *Interface) getOrHandshake(vpnAddr netip.Addr, cacheCallback func(*HandshakeHostInfo)) (*HostInfo, bool) {
 	_, found := f.myVpnNetworksTable.Lookup(vpnAddr)
 	if !found {
-		vpnAddr = f.inside.RouteFor(vpnAddr)
-		if !vpnAddr.IsValid() {
+		possibleGateways := f.inside.RoutesFor(vpnAddr)
+		if len(possibleGateways) == 0 {
 			return nil, false
 		}
+
+		// Multipath routes can be defined, but always take the first one for now
+		vpnAddr = possibleGateways[0]
 	}
 
 	return f.handshakeManager.GetOrHandshake(vpnAddr, cacheCallback)
--- a/inside.go
+++ b/inside.go
@@ -9,7 +9,6 @@ import (
 	"github.com/slackhq/nebula/iputil"
 	"github.com/slackhq/nebula/noiseutil"
 	"github.com/slackhq/nebula/routing"
-	"github.com/zeebo/xxh3"
 )
 
 func (f *Interface) consumeInsidePacket(packet []byte, fwPacket *firewall.Packet, nb, out []byte, q int, localCache firewall.ConntrackCache) {
@@ -139,36 +138,6 @@ func (f *Interface) getOrHandshakeNoRouting(vpnAddr netip.Addr, cacheCallback fu
 	return nil, false
 }
 
-func hashPacket(p *firewall.Packet) int {
-	hasher := xxh3.Hasher{}
-
-	hasher.Write(p.LocalAddr.AsSlice())
-	hasher.Write(p.RemoteAddr.AsSlice())
-	hasher.Write([]byte{
-		byte(p.LocalPort & 0xFF),
-		byte((p.LocalPort >> 8) & 0xFF),
-		byte(p.RemotePort & 0xFF),
-		byte((p.RemotePort >> 8) & 0xFF),
-		byte(p.Protocol),
-	})
-
-	// Uses xxh3 as it is a fast hash with good distribution
-	return int(hasher.Sum64() & 0x7FFFFFFF)
-}
-
-func balancePacket(fwPacket *firewall.Packet, gateways []routing.Gateway) netip.Addr {
-	hash := hashPacket(fwPacket)
-
-	for i := range gateways {
-		if hash <= gateways[i].UpperBound() {
-			return gateways[i].Addr()
-		}
-	}
-
-	// This should never happen
-	panic("The packet hash value should always fall inside a gateway bucket")
-}
-
 // getOrHandshakeConsiderRouting will try to find the HostInfo to handle this packet, starting a handshake if necessary.
 // If the 2nd return var is false then the hostinfo is not ready to be used in a tunnel.
 func (f *Interface) getOrHandshakeConsiderRouting(fwPacket *firewall.Packet, cacheCallback func(*HandshakeHostInfo)) (*HostInfo, bool) {
@@ -190,7 +159,7 @@ func (f *Interface) getOrHandshakeConsiderRouting(fwPacket *firewall.Packet, cac
 		return f.handshakeManager.GetOrHandshake(gateways[0].Addr(), cacheCallback)
 	} else {
 		// Multi gateway route, perform ECMP categorization
-		gatewayIp := balancePacket(fwPacket, gateways)
+		gatewayIp := routing.BalancePacket(fwPacket, gateways)
 
 		// Do not pass a cacheCallback, if the node is not reachable we will attempt other nodes
 		// so we don't want to cache the packet to avoid sending duplicate packets if this gateway comes back up.
--- a/inside.go
+++ b/inside.go
@@ -123,7 +123,7 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
 	f.sendNoMetrics(header.Message, 0, ci, hostinfo, netip.AddrPort{}, out, nb, packet, q)
 }
 
-// This should only called for internal nebula traffic
+// Handshake will attempt to initiate a tunnel with the provided vpn address if it is within our vpn networks. This is a no-op if the tunnel is already established or being established
 func (f *Interface) Handshake(vpnAddr netip.Addr) {
 	f.getOrHandshakeNoRouting(vpnAddr, nil)
 }
@@ -169,9 +169,8 @@ func balancePacket(fwPacket *firewall.Packet, gateways []routing.Gateway) netip.
 	panic("The packet hash value should always fall inside a gateway bucket")
 }
 
-// This is called for external traffic
-// getOrHandshake returns nil if the vpnIp is not routable.
-// If the 2nd return var is false then the hostinfo is not ready to be used in a tunnel
+// getOrHandshakeConsiderRouting will try to find the HostInfo to handle this packet, starting a handshake if necessary.
+// If the 2nd return var is false then the hostinfo is not ready to be used in a tunnel.
 func (f *Interface) getOrHandshakeConsiderRouting(fwPacket *firewall.Packet, cacheCallback func(*HandshakeHostInfo)) (*HostInfo, bool) {
 
 	destinationAddr := fwPacket.RemoteAddr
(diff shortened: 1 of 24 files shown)


johnmaguire:
I think this PR achieves the goals of #353

johnmaguire:
Hi @dioss-Machiel, this looks very interesting. One note:

> This can be used for example to aggregate multiple links and provide redundancy to an external location where nebula cannot be installed.

It does look like this PR may offer effective load balancing between gateways, but I'm not sure if it provides redundancy. If a given gateway is down, will this code detect it, and route on the other gateway instead?

dioss-Machiel:
This implementation does indeed provides redundancy. 

How this technically works:
For each gateway that serves a route a bucket is calculated, according to the weight.
A packet that needs to be routed is hashed, and the gateway is determined by the bucket it falls in.
If that gateway is not reachable then the entire list of possible gateways for that route is tried until exhausted.
This means when one of the gateways is down the traffic is no longer properly being balanced, but there is still connectivity. 

For example: if you have 3 gateways with weight "1" then each of them handles 1/3 of the traffic.
If a gateway goes down then one will handle 2/3 and another will handle 1/3

A more advanced implementation could recalculate the buckets when nodes go up or down to keep balancing the traffic on the remaining gateways but that comes with more complexity and maybe a slight penalty to routing speed.

nbrownus:
@dioss-Machiel we merged a pretty large change to support ipv6 on the overlay, mind rebasing to clear up the conflicts?

dioss-Machiel:
@nbrownus Conflicts have been resolved

dioss-Machiel:
I fixed the testify errors

routing/gateway.go
```diff
@@ -0,0 +1,50 @@
+package routing
+
+import (
+	"fmt"
+	"net/netip"
+)
+
+type Gateway struct {
+	ip         netip.Addr
```
nbrownus, line 9: We've started moving from `ip` => `addr`, would be helpful to adjust the names in this file

inside.go
```diff
@@ -121,22 +123,118 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
 	f.sendNoMetrics(header.Message, 0, ci, hostinfo, netip.AddrPort{}, out, nb, packet, q)
 }
 
+// This should only called for internal nebula traffic
```
nbrownus, line 126: nit: 
```// Handshake will attempt to initiate a tunnel with the provided vpn address if it is within our vpn networks. This is a no-op if the tunnel is already established or being established```

I think there is a near term future where this function no longer exists but that is out of scope of this PR

File: inside.go [hunk 3]
```diff
@@ -121,22 +123,118 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
...
-		if !vpnAddr.IsValid() {
-			return nil, false
+	if found {
+		return f.handshakeManager.GetOrHandshake(vpnAddr, cacheCallback)
+	}
+
+	return nil, false
+}
+
+func hashPacket(p *firewall.Packet) int {
```
nbrownus, line 142: `hashPacket` and `balancePacket` should be moved to the `gateway` package in my opinion, `BalancePacket` would be the only thing we need to expose there

inside.go
```diff
@@ -121,22 +123,118 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
...
+
+	for i := range gateways {
+		if hash <= gateways[i].UpperBound() {
+			return gateways[i].Ip()
 		}
 	}
 
-	return f.handshakeManager.GetOrHandshake(vpnAddr, cacheCallback)
+	// This should never happen
+	panic("The packet hash value should always fall inside a gateway bucket")
```
nbrownus, line 169: Instead of a panic we should add a 2nd return `ok` or advise callers to check that the returned `netip.Addr` is valid before continuing with `IsValid()`

Basically, don't force a crash or recovery, let the caller decide to log in a lighter weight way

inside.go
```diff
@@ -121,22 +123,118 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
...
+}
+
+// This is called for external traffic
```
nbrownus, line 172: nit:
```
// getOrHandshakeConsiderRouting will try to find the HostInfo to handle this packet, starting a handshake if necessary.
// If the 2nd return var is false then the hostinfo is not ready to be used in a tunnel.
```

inside.go
```diff
@@ -121,22 +123,118 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
...
+// getOrHandshake returns nil if the vpnIp is not routable.
+// If the 2nd return var is false then the hostinfo is not ready to be used in a tunnel
+func (f *Interface) getOrHandshakeConsiderRouting(fwPacket *firewall.Packet, cacheCallback func(*HandshakeHostInfo)) (*HostInfo, bool) {
+
+	destinationIp := fwPacket.RemoteAddr
```
nbrownus, line 177: We are trying to move from `ip` -> `addr`, it would be helpful to rename to `destinationAddr`

inside.go
```diff
@@ -121,22 +123,118 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
...
+		// use of the use_system_route_table option
+
+		if f.l.Level >= logrus.DebugLevel {
+			f.l.WithField("destination", destinationIp).
+				WithField("originalGateway", gatewayIp).
+				Debugln("Calculated gateway for ECMP not available, attempting other gateways")
+		}
+
+		var hostInfo *HostInfo
+		var ready bool
```
nbrownus, line 216: nit: shadowing an already declared `ready`

inside.go
```diff
@@ -121,22 +123,118 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
...
+}
+
+// This is called for external traffic
+// getOrHandshake returns nil if the vpnIp is not routable.
+// If the 2nd return var is false then the hostinfo is not ready to be used in a tunnel
+func (f *Interface) getOrHandshakeConsiderRouting(fwPacket *firewall.Packet, cacheCallback func(*HandshakeHostInfo)) (*HostInfo, bool) {
+
+	destinationIp := fwPacket.RemoteAddr
+
+	hostinfo, ready := f.getOrHandshakeNoRouting(destinationIp, cacheCallback)
```
nbrownus, line 179: nit: renaming to `hostInfo` or changing other references of `hostInfo` to `hostinfo` would simplify the variable declarations

inside.go
```diff
@@ -121,22 +123,118 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
...
+		// Do not pass a cacheCallback, if the node is not reachable we will attempt other nodes
+		// so we don't want to cache the packet to avoid sending duplicate packets if this gateway comes back up.
+		if hostInfo, ready := f.handshakeManager.GetOrHandshake(gatewayIp, nil); ready {
+			return hostInfo, true
+		}
+
+		// It appears the selected gateway cannot be reached, find another gateway to fallback on.
+		// The current implementation breaks ECMP but that seems better than no connectivity.
+		// If ECMP is also required when a gateway is down then connectivity status
+		// for each gateway nees to be kept and the weights recalculated when they go up or down.
```
nbrownus, line 205: nit: `nees` -> `needs`

inside.go
```diff
@@ -121,22 +123,118 @@ func (f *Interface) rejectOutside(packet []byte, ci *ConnectionState, hostinfo *
...
+			}
+
+			// Store the HandshakeHostInfo for the cache callback
+			if hostInfo, ready = f.handshakeManager.GetOrHandshake(gateways[i].Ip(), hhReceiver); ready {
+				return hostInfo, true
+			}
+		}
+
+		// No gateways reachable, cache packet in last gateway attempted
+		cacheCallback(handshakeHostInfo)
```
nbrownus, line 235: I am a bit uncomfortable about this naive selection of the last configured unsafe router to cache the packet.

In the most degenerate scenario there is an amplified DDoS opportunity. Where each node in the mesh is establishing tunnels with all unsafe routers for this destination address and the unsafe route is highly subscribed. This goes against the intent of the feature to spread the packet load across 2 or more unsafe routers.

I think a weighted random selection would be a better approach, 2nd being just a plain random selection.

routing/gateway.go
dioss-Machiel: Done!

inside.go
dioss-Machiel: Done

inside.go
dioss-Machiel: I agree, they have been moved to "balance.go"

inside.go
dioss-Machiel: This was a bit of hard one for me to figure out, as it really should be impossible to hit this panic unless there is a bug in the code. However I understand you may prefer Nebula to keep working, even if there is a bug.

So I have reworked this function with a second return parameter that indicates if balancing succeeded. If balancing fails then a random strategy will be chosen instead so connectivity is kept.

inside.go
dioss-Machiel: Done

inside.go
dioss-Machiel: Done

inside.go
dioss-Machiel: I must have run out of coffee at this moment, thank you for spotting this, has been fixed.

inside.go
dioss-Machiel: Done

inside.go
dioss-Machiel: Done

inside.go
dioss-Machiel: Valid point, I have changed the implementation so it caches the packet in the gateway that was selected by BalancePacket. This should result in the caching being balanced by weight when all gateways are down.

inside.go
nbrownus: Thanks! We can make this `hostinfo, ready = f.getOrHandshakeNoRouting` now as well.

dioss-Machiel: Good spot, missed that one, should be resolved now

nbrownus:
Did a quick benchmark, `xxh3` has a really nice distribution but it's quite slow compared to some other options.

```
BenchmarkHashTest/cespare/xxhash/v2-10         	74028885	        15.99 ns/op	       0 B/op	       0 allocs/op
BenchmarkHashTest/maphash-10                   	65325824	        18.25 ns/op	       0 B/op	       0 allocs/op
BenchmarkHashTest/zeebo/xxh3-10                	27690496	        43.31 ns/op	       0 B/op	       0 allocs/op
BenchmarkHashTest/OneOfOne/xxhash-10           	79155019	        15.29 ns/op	       0 B/op	       0 allocs/op
```

This benchmark includes 2 other xxhash implementations but they have slightly worse distribution while being much faster than xxh3.

I included `maphash` since its baked into golang, it has inherent randomness though so the result of the balance tests can change between runs.

As a reference point, the firewall takes ~40ns/op on my machine to pass a packet on a group or name match. Worst case performance in the firewall is ~95ns/op to evaluate and fail a full rule evaluation. 

If we went with `OneOfOne/xxhash` or `cespare/xxhash/v2` the current tests would fail:
```
--- FAIL: TestPacketsAreBalancedEqually (0.02s)
    balance_test.go:60: 
        	Error Trace:	routing/balance_test.go:60
        	Error:      	Max difference between 21845 and 21651 allowed is 100, but difference was 194
        	Test:       	TestPacketsAreBalancedEqually
        	Messages:   	Expected 21845 +/- 100, but got 21651
    balance_test.go:61: 
        	Error Trace:	routing/balance_test.go:61
        	Error:      	Max difference between 21845 and 22042 allowed is 100, but difference was -197
        	Test:       	TestPacketsAreBalancedEqually
        	Messages:   	Expected 21845 +/- 100, but got 21651
```

Which doesn't seem too terrible to me

overlay/route.go
```diff
@@ -201,14 +204,63 @@ func parseUnsafeRoutes(c *config.C, networks []netip.Prefix) ([]Route, error) {
...
+				gatewayWeight, ok := rGatewayWeight.(int)
+				if !ok {
+					_, err = strconv.ParseInt(rGatewayWeight.(string), 10, 32)
+					if err != nil {
+						return nil, fmt.Errorf("entry .weight in tun.unsafe_routes[%v].via[%v] is not an integer", i+1, ig+1)
+					}
+				}
+
+				if gatewayWeight < 1 || gatewayWeight > math.MaxInt32 {
+					return nil, fmt.Errorf("entry .weight in tun.unsafe_routes[%v].via[%v] is not in range (1-%d) : %v", i+1, ig+1, math.MaxInt32, metric)
```
nbrownus, line 255: `metric` should be `gatewayWeight`

overlay/tun_linux.go
```diff
@@ -550,30 +551,76 @@ func (t *tun) watchRoutes() {
...
+				t.l.WithField("route", r).Debug("Ignoring multipath route update, invalid gateway address")
+			} else {
+				gwAddr = gwAddr.Unmap()
+
+				if !t.isGatewayInVpnNetworks(gwAddr) {
+					// Gateway isn't in our overlay network, ignore
+					t.l.WithField("route", r).Debug("Ignoring route update, not in our network")
+				} else {
+					// p.Hops+1 = weight of the route
+					gateways = append(gateways, routing.NewGateway(gwAddr, p.Hops+1))
```
nbrownus, line 607: When doing `sudo ip route add 3.3.3.4/32 nexthop via 192.168.5.13 weight 5 nexthop via 192.168.5.14 weight 10` I do see `p.Weight` set correctly on my machine, is there a reason to use `r.Hops+1` instead?

overlay/tun_linux.go
```diff
@@ -589,12 +636,12 @@ func (t *tun) updateRoutes(r netlink.RouteUpdate) {
 	newTree := t.routeTree.Load().Clone()
 
 	if r.Type == unix.RTM_NEWROUTE {
-		t.l.WithField("destination", r.Dst).WithField("via", r.Gw).Info("Adding route")
-		newTree.Insert(dst, gwAddr)
+		t.l.WithField("destination", dst).WithField("via", gateways).Info("Adding route")
```
nbrownus, line 639: We need to transform `gateways` to a human readable string here and below.

Currently this is logging:
```
via="[{{{0 281473913980173} {0xc00014a150}} 5 715827882} {{{0 281473913980174} {0xc00014a150}} 10 2147483647}]"
```

I haven't audited all log sites for `[]routing.Gateway` yet but it would probably be best to deal with a concrete type `Gateways` than the slice to avoid future foot guns when trying to log this structure.

Doing the easy thing and casting for logging is a bit easier but the end implementation looks mostly the same
```
type viaGateways []routing.Gateway
func (g viaGateways) String() string {
	str := ""
	for i, gw := range g {
		str += gw.String()
		if i < len(g)-1 {
			str += ", "
		}
	}
	return str
}

...

    t.l.WithField("destination", dst).WithField("via", viaGateways(gateways)).Info("Adding route")
```

This gives a slightly better log line:
```
via="192.168.5.13:5/715827882, 192.168.5.14:10/2147483647"
```

To which I think we should adjust the `Gateway.String()` to look something along the lines of
```
func (g *Gateway) String() string {
	return fmt.Sprintf("%s weight %d", g.addr, g.weight)
}
```

I would like to avoid changing `via` from a string to an `[]` when logging in json to avoid the pitfalls of changing the field type for systems like elasticsearch.

overlay/tun_linux.go
dioss-Machiel: I might be overlooking something but I don't see a Weight field on NexthopInfo in netlink v1.3.0, there is only a "Weight" in the String() method for NexthopInfo 


https://github.com/vishvananda/netlink/blob/6f5713947556a0288c5cb71f036f9e91924ebcaa/route.go#L169

```
type NexthopInfo struct {
	LinkIndex int
	Hops      int
	Gw        net.IP
	Flags     int
	NewDst    Destination
	Encap     Encap
	Via       Destination
}

func (n *NexthopInfo) String() string {
	elems := []string{}
	elems = append(elems, fmt.Sprintf("Ifindex: %d", n.LinkIndex))
	if n.NewDst != nil {
		elems = append(elems, fmt.Sprintf("NewDst: %s", n.NewDst))
	}
	if n.Encap != nil {
		elems = append(elems, fmt.Sprintf("Encap: %s", n.Encap))
	}
	if n.Via != nil {
		elems = append(elems, fmt.Sprintf("Via: %s", n.Via))
	}
	elems = append(elems, fmt.Sprintf("Weight: %d", n.Hops+1))
	elems = append(elems, fmt.Sprintf("Gw: %s", n.Gw))
	elems = append(elems, fmt.Sprintf("Flags: %s", n.ListFlags()))
	return fmt.Sprintf("{%s}", strings.Join(elems, " "))
}
```

overlay/tun_linux.go
dioss-Machiel: Love the idea, does this look like a reasonable log line to you, or does it look to much like json?

`level=info msg="Adding route" destination=10.152.5.22/32 via="{addr: 10.159.0.1, weight: 1}, {addr: 10.159.0.2, weight: 1}"`

overlay/route.go
dioss-Machiel: Indeed!, Fixed now

overlay/tun_linux.go
nbrownus: Seems fine to me, @wadey do you have any issues with this format?

overlay/tun_linux.go
nbrownus: Yep, I got confused by the string output.

PR #1337, opened 2025-03-06
Group based query protection

HenryGrahamRivian:
This MR introduces an interface for filtering incoming host queries based on groups in the querying host's cert and groups in the queried host's cert.  

This will improve the privacy of the network for its users by preventing users from querying hosts that they are not supposed to have access to as defined by the lighthouse configuration. 

<details>
<summary> Required config file additions </summary>

```yaml
lighthouse:
  query_protection:
    test-protection:
      - allowed-group
```
</details>

<details>
<summary> Example positive interaction </summary>

mermaid-diagram-2025-02-14-140947

</details>

<details>
<summary> Example negative interaction </summary>

mermaid-diagram-2025-02-14-140746

</details>

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -72,9 +72,10 @@ type LightHouse struct {
 
 	calculatedRemotes atomic.Pointer[bart.Table[[]*calculatedRemote]] // Maps VpnAddr to []*calculatedRemote
 
-	metrics           *MessageMetrics
-	metricHolepunchTx metrics.Counter
-	l                 *logrus.Logger
+	metrics              *MessageMetrics
+	metricHolepunchTx    metrics.Counter
+	l                    *logrus.Logger
+	queryProtectionTable lightHouseQueryProtectionTable
 }
 
 // NewLightHouseFromConfig will build a Lighthouse struct from the values provided in the config object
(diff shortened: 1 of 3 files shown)


nbrownus:
Thanks for the PR!

To clarify (mostly for myself), this is default allow any unless a protected group is specified, once a group is declared as protected then only hosts with any 1 of the allowed groups can query for it. All in common lighthouses in the mesh between the host doing the querying and the host being queried must have the same configuration or else the information this is trying to protect will leak.

I immediately look to the firewall config and consider the precedent it sets.

Naming things is hard so please try to look past the names.

```yaml
lighthouse:
  query_protection:
    - target_groups: // Any host with this subset of groups 
        - "region: us-east-1"
        - "role: database"
      querying_groups: // Can only be queried by a host with this subset of groups
        - "region: us-east-1"
        - "role: webserver"
        
    // Additional identical targets_groups can be defined to allow other querying groups to query
    // This allows for easy rule appending from disparate sources, its easier for config management systems
    // to append individual rules than to roll them up into a single config stanza
    - target_groups: // Any host with this subset of groups 
        - "region: us-east-1"
        - "role: database"
      querying_groups: // Can only be queried by a host with this subset of groups
        - "region: us-east-1"
        - "role: job-worker"
        
    - target_groups: // Any host with this subset of groups 
        - "region: us-east-1"
        - "role: database"
      querying_groups: // Can only be queried by a host with this subset of groups
        - "region: us-east-1"
        - "role: database"
```

In the above example imagine you have a fleet of database instances in multiple aws regions. A database instance can only be discovered by a webserver, job worker, or another database within the same region.

I like the functionality but the performance of that could be quite poor.

2 other ways to accomplish this would be:

1. Denormalizing the group to something a single value can target `pg:us-east-1:job-worker`.

The downsides would be:
- Could lead to certificate bloat through having normalized and denormalized groups, this could eventually cause handshake packets to be fragmented and make it near impossible to stand up a tunnel.
- All hosts need to have the denormalized groups before you can install the rule

2. Introduce a non caching lighthouse mode where a host can query about another host but the lighthouse does not cache information about the host doing the query. You could then partition caching lighthouses for the groups you want to protect and allow lighthouses to cache information about them and expose all lighthouses to all hosts in a non caching way

The downsides would be:
- Requires real infrastructure
- Does not scale well in terms of query performance
- Does not scale well in terms of new protected groups

## TLDR

If we keep it simple, like it is now, I would suggest twisting the config to look like (forgive the naming)

```yaml
lighthouse:
  query_protection:
    - target_group: "role: database"
      querying_group: "role: webserver"
        
    - target_group: "role: database"
      querying_group: "role: job-worker"
        
    - target_group: "role: database"
      querying_group: "role: database"
```

I am curious what @JackDoan and @wadey think

PR #1338, opened 2025-03-06
build with go1.24

wadey:
This doesn't change our go.mod, which still only requires go1.22 as a minimum. It only changes our builds to use go1.24 so we have the latest improvements.

### DIFF
--- a/.github/workflows/gofmt.yml
+++ b/.github/workflows/gofmt.yml
@@ -18,7 +18,7 @@ jobs:
 
     - uses: actions/setup-go@v5
       with:
-        go-version: '1.23'
+        go-version: '1.24'
         check-latest: true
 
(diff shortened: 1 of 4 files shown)


nbrownus (approved):

.github/workflows/gofmt.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
     - uses: actions/setup-go@v5
       with:
-        go-version: '1.23'
```
jasikpark, line 21: Might want to pin to 1.24.1 to avoid some bugs: https://github.com/golang/go/issues?q=sort%3Aupdated-desc+milestone%3AGo1.24.1+label%3ACherryPickApproved

PR #1345, opened 2025-03-07
Improve logging when handshaking with an invalid cert

nbrownus:
Fixes a bug where the cert was not printed in debug log level when an invalid cert is encountered.

Adds some information about an invalid cert when in info log level is set. This is to help locate the offending host/certificate.

Logs the cert in some cases after we have determined it is trusted but still invalid for some reason.

### DIFF
--- a/cert/cert.go
+++ b/cert/cert.go
@@ -113,10 +113,10 @@ func (cc *CachedCertificate) String() string {
 	return cc.Certificate.String()
 }
 
-// RecombineAndValidate will attempt to unmarshal a certificate received in a handshake.
+// Recombine will attempt to unmarshal a certificate received in a handshake.
 // Handshakes save space by placing the peers public key in a different part of the packet, we have to
 // reassemble the actual certificate structure with that in mind.
-func RecombineAndValidate(v Version, rawCertBytes, publicKey []byte, curve Curve, caPool *CAPool) (*CachedCertificate, error) {
+func Recombine(v Version, rawCertBytes, publicKey []byte, curve Curve) (Certificate, error) {
 	if publicKey == nil {
 		return nil, ErrNoPeerStaticKey
 	}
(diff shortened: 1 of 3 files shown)


wadey (dismissed):

handshake_ix.go
```diff
@@ -487,30 +498,42 @@ func ixHandshakeStage2(f *Interface, addr netip.AddrPort, via *ViaSender, hh *Ha
...
-		if f.l.Level > logrus.DebugLevel {
-			e = e.WithField("cert", remoteCert)
+		if f.l.Level >= logrus.DebugLevel {
+			e = e.WithField("cert", rc)
+		} else {
+			fp, err := rc.Fingerprint()
+			if err != nil {
+				fp = "<error generating certificate fingerprint>"
+			}
+			e = e.WithField("certFingerprint", fp).
```
JackDoan, line 523: IMO it'd be nice to have these fields even if debug logging is enabled

handshake_ix.go
```diff
@@ -132,13 +132,28 @@ func ixHandshakeStage1(f *Interface, addr netip.AddrPort, via *ViaSender, packet
...
```
JackDoan, line 155: same as below

cert/cert.go
```diff
@@ -127,15 +127,9 @@ func RecombineAndValidate(v Version, rawCertBytes, publicKey []byte, curve Curve
 
 	c, err := unmarshalCertificateFromHandshake(v, rawCertBytes, publicKey, curve)
```
JackDoan, line 128: now that this function is just error checks, we could inline unmarshalCertificateFromHandshake?

handshake_ix.go
nbrownus: I had considered it but it duplicates values that would be logged in the cert field when debug is on. 

Not opposed if anyone has a strong opinion here.

cert/cert.go
nbrownus: I have a slight preference to keep it here to reduce the complexity of the handshake functions.

JackDoan (approved):

Issue #1352, opened 2025-03-14
🐛 BUG: Incorrect conversion between integer types


odaysec:
### What version of `nebula` are you using? (`nebula -version`)

v1.9.5

### What operating system are you using?

Linux

### Describe the Bug

https://github.com/slackhq/nebula/blob/3de36c99b6c7e304a463128ae9319d96bfd822e9/config/config.go#L249-L249

If a string is parsed into an int using `strconv.Atoi`, and subsequently that int is converted into another integer type of a smaller size, the result can produce unexpected values. This also applies to the results of `strconv.ParseInt` and `strconv.ParseUint` when the specified size is larger than the size of the type that number is converted to.

## POC
In the first assume that an input string is passed to `parseAllocateBad1` function, parsed by `strconv.Atoi`, and then converted into an `int32` type:
```go
package main

import (
	"strconv"
)

func parseAllocateBad1(wanted string) int32 {
	parsed, err := strconv.Atoi(wanted)
	if err != nil {
		panic(err)
	}
	return int32(parsed)
}
func parseAllocateBad2(wanted string) int32 {
	parsed, err := strconv.ParseInt(wanted, 10, 64)
	if err != nil {
		panic(err)
	}
	return int32(parsed)
}
```
The bounds are not checked, so this means that if the provided number is greater than the maximum value of type `int32`, the resulting value from the conversion will be different from the actual provided value.

To avoid unexpected values, you should either use the other functions provided by the `strconv` package to parse the specific types and bit sizes as shown in the `parseAllocateGood2` function; or check bounds as in the `parseAllocateGood1` function.

```go
package main

import (
	"math"
	"strconv"
)

func main() {

}

const DefaultAllocate int32 = 256

func parseAllocateGood1(desired string) int32 {
	parsed, err := strconv.Atoi(desired)
	if err != nil {
		return DefaultAllocate
	}
	// GOOD: check for lower and upper bounds
	if parsed > 0 && parsed <= math.MaxInt32 {
		return int32(parsed)
	}
	return DefaultAllocate
}
func parseAllocateGood2(desired string) int32 {
	// GOOD: parse specifying the bit size
	parsed, err := strconv.ParseInt(desired, 10, 32)
	if err != nil {
		return DefaultAllocate
	}
	return int32(parsed)
}

func parseAllocateGood3(wanted string) int32 {
	parsed, err := strconv.ParseInt(wanted, 10, 32)
	if err != nil {
		panic(err)
	}
[...]

PR #1353, opened 2025-03-14
Fix: Incorrect conversion between integer types

odaysec:
## Ticket 🎟️ #1352
---
To fix the problem, we need to ensure that the value being converted to `uint32` is within the valid range for `uint32`, which is from 0 to `math.MaxUint32`. This involves adding a lower bound check to ensure the value is not negative. 

The best way to fix this is to modify the `GetUint32` function to include a check for negative values before performing the conversion. This can be done by adding a condition to check if the value is less than 0 and returning the default value `d` if it is.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -243,7 +243,7 @@ func (c *C) GetInt(k string, d int) int {
 // GetUint32 will get the uint32 for k or return the default d if not found or invalid
 func (c *C) GetUint32(k string, d uint32) uint32 {
 	r := c.GetInt(k, int(d))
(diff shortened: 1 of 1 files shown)


johnmaguire (approved):

johnmaguire:
@odaysec Are you able to sign the CLA? Thanks in advance.

odaysec:
/cla `You already signed the CLA on 2025-03-14`

johnmaguire:
Sorry about that and thanks for the contribution @odaysec

PR #1374, opened 2025-04-01
Rename pki.default_version to pki.initiating_version

johnmaguire:
Default may imply that we will fallback to v1 in the event that v2 does not work. This is not true - this config value specifies the exact certificate version that should be used when initiating a tunnel.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -13,11 +13,11 @@ pki:
   # disconnect_invalid is a toggle to force a client to be disconnected if the certificate is expired or invalid.
   #disconnect_invalid: true
 
-  # default_version controls which certificate version is used in handshakes.
+  # outgoing_version controls which certificate version is used when initiating handshakes.
   # This setting only applies if both a v1 and a v2 certificate are configured, in which case it will default to `1`.
   # Once all hosts in the mesh are configured with both a v1 and v2 certificate then this should be changed to `2`.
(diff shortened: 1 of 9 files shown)


examples/config.yml
```diff
@@ -13,11 +13,11 @@ pki:
   # disconnect_invalid is a toggle to force a client to be disconnected if the certificate is expired or invalid.
   #disconnect_invalid: true
 
-  # default_version controls which certificate version is used in handshakes.
+  # outgoing_version controls which certificate version is used when initiating handshakes.
```
nbrownus, line 16: Seeing it here now makes me think `initiating_version` might make the most sense?

wadey: I agree, I think I like initiating version best.

johnmaguire: I'm fine with initiating. I considered it first, since I know we commonly use that terminology in the code, but I wasn't sure if it was as clear.

The argument for outgoing is that from "my" perspective, I or the other end may be "initiating" the connection, making this config option a little ambiguous. But "outgoing" tunnels are clearly tunnels that originate from my end.

The argument for initiating is that it's language we use everywhere else when talking about standing up tunnel. And "outgoing" has the potential to confuse with outgoing connections vs. outgoing handshakes.

johnmaguire: Updated the PR to `initiating_version`

PR #1386, opened 2025-04-08
populate default Build version if missing

wadey:
Use the Go module information built into the binary if the Build var wasn't set during the build.

This means if you install via a specific tag, you get:

    go install github.com/slackhq/nebula/cmd/nebula@v1.9.5

    $ nebula -version
    Version: 1.9.5

And if you install master, you get:

    go install github.com/slackhq/nebula/cmd/nebula@master

    $ nebula -version
    Version: 1.9.5-0.20250408154034-18279ed17b10

### DIFF
--- a/cmd/nebula-cert/main.go
+++ b/cmd/nebula-cert/main.go
@@ -5,9 +5,27 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime/debug"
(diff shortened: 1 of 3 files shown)


wadey:
I guess we should also do this higher up in the `nebula` library to also work when folks are using Nebula as a library

PR #1399, opened 2025-05-01
Support IPv6 tunneling in FreeBSD

sl274:
Recent merge of cert-v2 support introduced the ability to tunnel IPv6. However, FreeBSD's IPv6 tunneling does not work for 2 reasons:
* The ifconfig commands did not work for IPv6 addresses
* The tunnel device was not configured for link-layer mode, so it only supported IPv4

This PR improves FreeBSD tunneling support in 3 ways:
* Use ioctl instead of exec'ing ifconfig to configure the interface, with additional logic to support IPv6
* Configure the tunnel in link-layer mode, allowing IPv6 traffic
* Use readv() and writev() to communicate with the tunnel device, to avoid the need to copy the packet buffer


sl274:
CLA is signed

dch:
@sl274 awesome! as somebody who's new to nebula, how would I test this?

nbrownus:
@dch If you want to test prior to merging then pull this branch and run `make`, this will produce the `nebula` binary that you can run and verify the change on your system.

nbrownus (approved):
Thanks, change looks good! We need to update route table management as well and I've got that coded up and will create a new PR for it.

PR #1403, opened 2025-05-08
Fix relay migration panic by covering every possible relay state

brad-defined:
With the move to ipnet.Addr types, there will be a panic if an uninitialized netip.Addr type has `As4()` called on it.

We received a stack trace from a user showing this was happening on line 274 of connection_manager. Based on code inspection, this PR ensures that every State case is covered to avoid an uninitialized Addr panicking on line 274.

I tried for a bit, but couldn't come up with a nice automated test for this case.

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"fmt"
 	"net/netip"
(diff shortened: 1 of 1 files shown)


jasikpark (approved):
lgtm, nice refactor

nbrownus (approved):

PR #1405, opened 2025-05-12
WIP: Remove unused and stale tunnels. punch less.

nbrownus:

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -45,7 +45,7 @@ type connectionManager struct {
 	relayUsedLock *sync.RWMutex
 
 	// Track last communication with hosts
-	lastCommMap       map[uint32]*LastCommunication
+	lastCommMap       map[uint32]time.Time
 	lastCommLock      *sync.RWMutex
 	inactivityTimer   *LockingTimerWheel[uint32]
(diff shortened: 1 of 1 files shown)


connection_manager.go
```diff
@@ -379,7 +577,7 @@ func (n *connectionManager) makeTrafficDecision(localIndex uint32, now time.Time
 			// This is similar to the old punchy behavior with a slight optimization.
 			// We aren't receiving traffic but we are sending it, punch on all known
 			// ips in case we need to re-prime NAT state
-			n.sendPunch(hostinfo)
+			//n.sendPunch(hostinfo)
```
JackDoan, line 580: imo we should change this punch behavior separately

nbrownus: Agree, the default UDP timeout in many NATs is going to be 180 seconds. If we remove punching then the maximum we can somewhat safely assume an idle tunnel can idle for is 3 minutes.

rawdigits: We should modify this entire function to not punch to lighthouses, which means that we trend toward idle/no punches when there are no tunnels besides lighthouses.

PR #1413, opened 2025-05-23
Drop inactive tunnels

nbrownus:
This PR enables nebula to close tunnels after a period of inactivity, it is not enabled by default but we should enable it in a future major/minor release. It also disables punchy from punching to lighthouses, which comes with some very narrow caveats which are detailed in the comments.

The original PoC is present in #1405 

2 new config items are added, both are reload-able.
- `tunnels.drop_inactive` - Enables or disables the feature, default is disabled (false) currently.
- `tunnels.inactivity_timeout` - Determines the duration of inactivity before `tunnels.drop_inactive` is armed. Default is currently 10 minutes.

This is a rather large refactor because each state we wanted to track became map of all tunnel indexes with their own mutexes, those mutexes have caused problems in the past, and there isn't a compelling reason to have all that state stored in a single container. Instead, this stores a majority of the individual states on the `HostInfo` with less hot-path-blocking atomics.

These configs are not documented but setting `tunnels.inactivity_timeout` below `timers.connection_alive_interval` (5s) or `timers.pending_deletion_interval` (10s) would lead to possibly unexpected outcomes.

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -245,7 +245,11 @@ type HostInfo struct {
 
 	//TODO: in, out, and others might benefit from being an atomic.Int32. We could collapse connectionManager pendingDeletion, relayUsed, and in/out into this 1 thing
 	in, out, pendingDeletion atomic.Bool
-	lastComms                atomic.Pointer[time.Time]
+
+	// lastUsed tracks the last time ConnectionManager checked the tunnel and it was in use.
+	// This value will be behind against actual tunnel utilization in the hot path.
+	// This should only be used by the ConnectionManagers ticker routine.
+	lastUsed time.Time
 }
 
 type ViaSender struct {
--- a/hostmap.go
+++ b/hostmap.go
@@ -242,6 +242,10 @@ type HostInfo struct {
 	// Used to track other hostinfos for this vpn ip since only 1 can be primary
 	// Synchronised via hostmap lock and not the hostinfo lock.
 	next, prev *HostInfo
+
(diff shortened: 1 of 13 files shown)


wadey:
It looks like this needs https://github.com/slackhq/nebula/pull/1311 backported to fix the smoke tests

hostmap.go
```diff
@@ -242,6 +242,10 @@ type HostInfo struct {
 	// Used to track other hostinfos for this vpn ip since only 1 can be primary
 	// Synchronised via hostmap lock and not the hostinfo lock.
 	next, prev *HostInfo
+
+	//TODO: in, out, and others might benefit from being an atomic.Int32. We could collapse connectionManager pendingDeletion, relayUsed, and in/out into this 1 thing
+	in, out, pendingDeletion atomic.Bool
+	lastComms                atomic.Pointer[time.Time]
```
brad-defined: Does this result in a lot of allocations? I think the time.Time pointer must be allocated and is modified per packet in/out?

I wonder if we can use something cheaper here, like an atomic int. I bet we could use something from the timer wheel to track the advancement of time without being so exact as a time.Time struct.

nbrownus: Good call, we can use a counter that lives in `ConnectionManager` and is incremented on each tick. That gives us 500ms resolution. If we are ok losing a bit more resolution we can have `ConnectionManager` set that in a given `HostInfo` and avoid an additional atomic op per packet.

PR #1416, opened 2025-05-28
Add ability to generate a certificate from an existing key.

SohamG:
Hello,

I made this quick patch to add the ability to make a CA certificate from an already existing private key. This could be useful for when you accidentally delete your ca.crt file.

Note that my original intention was to be able to have a v1 CA and a v2 CA signed by the same key and hence trusted/verify-able. My use case is that I have a v2 network for all of my Linux hosts, but the iPhone app only supports v1 CA crts and I could not find a way to reconcile this. This problem is still unsolved as I realized in testing this patch that the fingerprint of the Certs changes between v1 and v2 even if all other parameters including the private/public key remain constant. Nebula seems to be using the fingerprint to find the CA before it does any assymetric key verification.

(I could hack it further to make identical fingerprint v1 and v2 certs but thats too dirty...)

Please let me know if I need to make any formatting/code style changes. Licensed same as project, will sign CLA.

### DIFF
--- a/cmd/nebula-cert/ca.go
+++ b/cmd/nebula-cert/ca.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	ed25519_std "crypto/ed25519"
 	"github.com/skip2/go-qrcode"
 	"github.com/slackhq/nebula/cert"
 	"github.com/slackhq/nebula/pkclient"
@@ -24,6 +25,7 @@ type caFlags struct {
 	name             *string
 	duration         *time.Duration
 	outKeyPath       *string
+	inKeyPath        *string
 	outCertPath      *string
 	outQRPath        *string
 	groups           *string
@@ -50,6 +52,7 @@ func newCaFlags() *caFlags {
 	cf.version = cf.set.Uint("version", uint(cert.Version2), "Optional: version of the certificate format to use")
 	cf.duration = cf.set.Duration("duration", time.Duration(time.Hour*8760), "Optional: amount of time the certificate should be valid for. Valid time units are seconds: \"s\", minutes: \"m\", hours: \"h\"")
 	cf.outKeyPath = cf.set.String("out-key", "ca.key", "Optional: path to write the private key to")
+	cf.inKeyPath = cf.set.String("in-key", "", "Optional: path to read private key for the new certificate")
 	cf.outCertPath = cf.set.String("out-crt", "ca.crt", "Optional: path to write the certificate to")
 	cf.outQRPath = cf.set.String("out-qr", "", "Optional: output a qr code image (png) of the certificate")
 	cf.groups = cf.set.String("groups", "", "Optional: comma separated list of groups. This will limit which groups subordinate certs can use")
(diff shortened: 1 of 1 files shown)


cmd/nebula-cert/ca.go
```diff
@@ -220,9 +226,25 @@ func ca(args []string, out io.Writer, errOut io.Writer, pr PasswordReader) error
 		switch *cf.curve {
 		case "25519", "X25519", "Curve25519", "CURVE25519":
 			curve = cert.Curve_CURVE25519
-			pub, rawPriv, err = ed25519.GenerateKey(rand.Reader)
-			if err != nil {
-				return fmt.Errorf("error while generating ed25519 keys: %s", err)
+			if *cf.inKeyPath == "" {
```
JackDoan, line 229: I'd really like to have this feature, but I think if we're going to add it, we need to support P256 as well.

cmd/nebula-cert/sign.go
```diff
@@ -404,6 +404,14 @@ func newKeypair(curve cert.Curve) ([]byte, []byte) {
 	}
 }
 
+func x25519KeyFromPriv(priv []byte) ([]byte, []byte) {
+	pubkey, err := curve25519.X25519(priv, curve25519.Basepoint)
+	if err != nil {
+		panic(err)
```
JackDoan, line 410: In most cases, I prefer to bubble the error back up and let the caller decide to panic

cmd/nebula-cert/keygen.go
```diff
@@ -59,8 +63,22 @@ func keygen(args []string, out io.Writer, errOut io.Writer) error {
 	} else {
 		switch *cf.curve {
 		case "25519", "X25519", "Curve25519", "CURVE25519":
-			pub, rawPriv = x25519Keypair()
-			curve = cert.Curve_CURVE25519
+			if *cf.inKeyPath == "" {
```
JackDoan, line 66: Do you think this ought to be it's own "command"? Making this a bonus feature of `keygen` is potentially confusing.

Also, many users encrypt their private keys, we may need to support that. I think there's a helper function for this.... somewhere?

JackDoan (changes requested):
Thanks for the contribution! I really want us to have this feature! (writing out the pubkey PEM from a certificate is also a thing I'd like to add btw). 

We just need a little bit more polish to get this mergeable (P256 support, passwords, maybe a `to-public-key` subcommand?)

SohamG:
Hey! Thanks for the reply. As I said, I will look into adding P256 now that I know there is interest in the PR. I'll add a new subcommand to `nebula-cert` and perhaps we can finalize the name of said command later.

cmd/nebula-cert/ca.go
SohamG: On it. My particular use case was Curve25519 but I don't see too much trouble adding this for P256 too.

cmd/nebula-cert/sign.go
SohamG: Well the original function I modeled this one from (similar name in the same file) didn't bubble err, so I didn't do it either. That being said its a relatively minor fix, which I will implement.

cmd/nebula-cert/keygen.go
SohamG: Yes, I did spot the helper, I threw this together as a Proof-Of-Concept and without knowing if this feature was desired. I'll add encrypted keys and a new subcommand.

PR #1417, opened 2025-05-29
DNS server fixes and improvements

andreasbell:
This PR adds a number of fixes and improvements to the DNS server, including:

- Adding configurable DNS suffix. This can be used to add an optional DNS suffix for example `.nebula` to all DNS records.  This is useful for adding a common namespace for all nebula hosts, and ensuring that all host have a fully qualified domain name.
- Adding PTR records for all IP addresses. This makes it possible to do reverse DNS lookup and finding the host name from an IP address.
- Fixing broken TXT records. There seems to be a bug in the current implementation stopping the DNS server form sending out TXT records. This PR fixes this issue, and reduces the information in the TXT record to only include Name, Networks/Ips, Groups, and UnsafeNetworks
- Adding support for multiple IP addresses by providing one A/AAAA record for each IPv4/IPv6 address of the host.
- Adds DNS record for the host runng the DNS server.

### DIFF
--- a/dns_server.go
+++ b/dns_server.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/miekg/dns"
 	"github.com/sirupsen/logrus"
+	"github.com/slackhq/nebula/cert"
 	"github.com/slackhq/nebula/config"
 )
 
@@ -76,9 +77,22 @@ func (d *dnsRecords) AddPTR(name string, addr netip.Addr) {
 	}
 }
 
+func (d *dnsRecords) AddTXT(name string, crt cert.Certificate) {
+	q := dns.Question{Name: name, Qclass: dns.ClassINET, Qtype: dns.TypeTXT}
+	qType := dns.TypeToString[q.Qtype]
+	rr, err := dns.NewRR(fmt.Sprintf("%s %s \"Name: %v\" \"Networks: %v\" \"Groups: %v\" \"UnsafeNetworks: %v\"", name, qType, crt.Name(), crt.Networks(), crt.Groups(), crt.UnsafeNetworks()))
+	if err == nil {
+		d.Lock()
+		defer d.Unlock()
+		d.dnsMap[q] = rr
+		d.l.Debugf("DNS record added %s", rr.String())
+	}
+}
+
 // Add adds the first IPv4 and IPv6 address that appears in `addresses` as the record for `host`
-func (d *dnsRecords) Add(name string, addresses []netip.Addr) {
-	host := dns.Fqdn(strings.ToLower(name + dnsSuffix))
+func (d *dnsRecords) Add(crt cert.Certificate, addresses []netip.Addr) {
+	host := dns.Fqdn(strings.ToLower(crt.Name() + dnsSuffix))
+	d.AddTXT(host, crt)
 	haveV4 := false
 	haveV6 := false
 	for _, addr := range addresses {
@@ -100,7 +114,7 @@ func (d *dnsRecords) Add(name string, addresses []netip.Addr) {
 func (d *dnsRecords) parseQuery(m *dns.Msg) {
 	for _, q := range m.Question {
 		switch q.Qtype {
-		case dns.TypeA, dns.TypeAAAA, dns.TypePTR:
+		case dns.TypeA, dns.TypeAAAA, dns.TypePTR, dns.TypeTXT:
 			d.RLock()
 			if rr, ok := d.dnsMap[q]; ok {
 				m.Answer = append(m.Answer, rr)
@@ -132,7 +146,7 @@ func dnsMain(l *logrus.Logger, cs *CertState, c *config.C) func() {
 	dnsSuffix = getDnsSuffix(c)
 
 	// Add self to dns records
-	dnsR.Add(cs.GetDefaultCertificate().Name(), cs.myVpnAddrs)
+	dnsR.Add(cs.GetDefaultCertificate(), cs.myVpnAddrs)
 
 	// attach request handler func
 	dns.HandleFunc(".", dnsR.handleDnsRequest)
--- a/dns_server.go
+++ b/dns_server.go
@@ -3,7 +3,6 @@ package nebula
 import (
 	"fmt"
 	"net"
-	"net/netip"
 	"strconv"
 	"strings"
 	"sync"
@@ -34,14 +33,14 @@ func newDnsRecords(l *logrus.Logger) *dnsRecords {
 	}
 }
 
-func (d *dnsRecords) addA(name string, addresses []netip.Addr) {
+func (d *dnsRecords) addA(name string, crt cert.Certificate) {
 	q := dns.Question{Name: name, Qclass: dns.ClassINET, Qtype: dns.TypeA}
 	d.dnsMap[q] = nil
 
-	for _, addr := range addresses {
-		if addr.Is4() {
+	for _, n := range crt.Networks() {
+		if n.Addr().Is4() {
 			qType := dns.TypeToString[q.Qtype]
-			rr, err := dns.NewRR(fmt.Sprintf("%s %s %s", name, qType, addr.String()))
+			rr, err := dns.NewRR(fmt.Sprintf("%s %s %s", name, qType, n.Addr().String()))
 			if err == nil {
 				d.dnsMap[q] = append(d.dnsMap[q], rr)
 				d.l.Debugf("DNS record added %s", rr.String())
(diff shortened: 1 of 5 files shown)


andreasbell:
@nbrownus I thought I would check in and see if any of the functionality in this PR  is something you would like to have as part of the Nebula DNS server. If so, I would be happy to make any necessary changes. Otherwise I will close the PR.

JackDoan:
Hi @andreasbell! Sorry to keep you waiting, we've been trying to ship 1.10 for about a year now, and we finally did it! Now we can look at the rest of the backlog, including this.

dns_server.go
```diff
@@ -19,129 +19,106 @@ import (
...
+			}
+		}
 	}
-	return string(b)
 }
 
-// Add adds the first IPv4 and IPv6 address that appears in `addresses` as the record for `host`
-func (d *dnsRecords) Add(host string, addresses []netip.Addr) {
-	host = strings.ToLower(host)
+func (d *dnsRecords) AddPTR(name string, addresses []netip.Addr) {
```
JackDoan: I can see why this functionality (PTR records) might be useful to you, but I hesitate to add it. We have always avoided "trusting the lighthouse". I do not want to encourage users to rely on the lighthouse to tell them "who" a host is. The only secure way Nebula provides learn the identity of a host at an overlay IP address is to handshake with that host.

If you can tell me more about your use-case for this feature, we might be able to work together to find another path. One of the things we've been talking about for ages but have yet to solidify, is a local API for applications on the same host as the `nebula` daemon to query the daemon for information about a host.

dns_server.go
```diff
@@ -19,129 +19,106 @@ import (
...
+	rr, err := dns.NewRR(fmt.Sprintf("%s %s \"Name: %v\" \"Networks: %v\" \"Groups: %v\" \"UnsafeNetworks: %v\"", name, qType, crt.Name(), crt.Networks(), crt.Groups(), crt.UnsafeNetworks()))
+	if err == nil {
+		d.dnsMap[q] = []dns.RR{rr}
+		d.l.Debugf("DNS record added %s", rr.String())
 	}
+}
 
-	//if we found it in this table, it's good
-	return d.myVpnAddrsTable.Contains(b)
+func (d *dnsRecords) Add(crt cert.Certificate, addresses []netip.Addr) {
```
JackDoan: if you're passing the whole certificate in, I would suggest dropping the `addresses` argument, and obtain the addresses from the certificate instead. Historically it was separate because we filtered the addresses that make it into the `hostinfo` object, but, we don't anymore :D

dns_server.go
```diff
@@ -19,129 +19,106 @@ import (
...
-	if hostinfo == nil {
-		return ""
-	}
-
-	q := hostinfo.GetCert()
-	if q == nil {
-		return ""
-	}
+func (d *dnsRecords) AddAAAA(name string, addresses []netip.Addr) {
+	d.Lock()
```
JackDoan: this is minor, but it would be nice to make these `Add*` functions private, and then obtain the lock only once, in the `Add()` function

dns_server.go
```diff
@@ -187,6 +168,11 @@ func getDnsServerAddr(c *config.C) string {
 	return net.JoinHostPort(dnsHost, strconv.Itoa(c.GetInt("lighthouse.dns.port", 53)))
 }
 
+func getDnsSuffix(c *config.C) string {
```
JackDoan, line 165: Can you tell me more about the use-case for this? Why not put the desired suffix into the `Name` on the host's certificate?

dns_server.go
```diff
@@ -157,14 +134,18 @@ func (d *dnsRecords) handleDnsRequest(w dns.ResponseWriter, r *dns.Msg) {
...
 }
 
-func dnsMain(l *logrus.Logger, cs *CertState, hostMap *HostMap, c *config.C) func() {
-	dnsR = newDnsRecords(l, cs, hostMap)
+func dnsMain(l *logrus.Logger, cs *CertState, c *config.C) func() {
+	dnsR = newDnsRecords(l)
+	dnsSuffix = getDnsSuffix(c)
+
+	// Add self to dns records
+	dnsR.Add(cs.GetDefaultCertificate(), cs.myVpnAddrs)
```
JackDoan: Thank you! This is overdue, and it's such an easy add.

hostmap.go
```diff
@@ -565,7 +565,7 @@ func (hm *HostMap) queryVpnAddr(vpnIp netip.Addr, promoteIfce *Interface) *HostI
 func (hm *HostMap) unlockedAddHostInfo(hostinfo *HostInfo, f *Interface) {
 	if f.serveDns {
 		remoteCert := hostinfo.ConnectionState.peerCert
-		dnsR.Add(remoteCert.Certificate.Name()+".", hostinfo.vpnAddrs)
```
JackDoan, line 568: Moving the `+"."` inside the `Add()` function is a good idea

dns_server.go
```diff
@@ -19,129 +19,106 @@ import (
...
-			ip := d.Query(q.Qtype, q.Name)
-			if ip.IsValid() {
-				rr, err := dns.NewRR(fmt.Sprintf("%s %s %s", q.Name, qType, ip))
-				if err == nil {
-					m.Answer = append(m.Answer, rr)
-				}
-			}
-		case dns.TypeTXT:
-			// We only answer these queries from nebula nodes or localhost
-			if !d.isSelfNebulaOrLocalhost(w.RemoteAddr().String()) {
```
JackDoan, line 134: I'll admit I don't know the full history on why we had this check, but I hesitate to remove it. I don't want to encourage people to obtain information about a Nebula host (like group membership) _outside_ of a Nebula tunnel.

JackDoan (changes requested):

dns_server.go
andreasbell: I agree that it is a good idea to avoid "trusting the lighthouse", however with DNS in general we are putting trust in the DNS server to respond with correct records, and i don't think the PTR record is more risky than a regular A or AAAA record, as it essentially is just a reverse lookup of a hostname from an IP address. In fact it essentially provides similar information that the current nebula DNS server TXT record provides.

My main usecase is one of convenience, being able to look up the hostname of a nebula host when only knowing an ip address, and providing a more complete set of functionality in the nebula DNS server.

However another good reason for keeping the PTR record is to replace the "none standard" TXT record lookup. 
```bash
# Current TXT record lookup
dig @100.100.0.1 +short "100.100.0.2" TXT
```
Currently nebula TXT records are fetched by querying the IP address in quotation marks, as part of this PR i have also changed TXT records to be queried by using the host name
```bash
# New TXT record lookup
dig @100.100.0.1 +short alice-laptop TXT
```
In this case the PTR records give a way of still getting the host name in the case that only the IP address is known
```bash
# New PTR record lookup
dig @100.100.0.1 +short -x 100.100.0.2
```

dns_server.go
andreasbell: I remember seeing some references to `addresses` containing only IP addresses "valid" in the subnet defined in the CA, while the certificate potentially containing address outside of the "valid" subnet. I assume that is the filtering you are referencing?

But if it is no longer used i will remove it as it makes things a bit cleaner :+1:

dns_server.go
andreasbell: In my case i have a Nebula network where I have used short names for my hosts. For example `alice-laptop` and `bob-pc` however, these names are not fully qualified domain names (FQDNs) as they do not contain a top level domain (suffix). Because of this DNS resolvers will treat them as local hostnames and may try to resolve these with for example the `.lan`, `.local` or `.my.company.domain.com` depending on the rules set up for the dns resolver. The addition of this suffix gives me the option to ensure that all nebula hosts have a FQDN. For example if i use the `.nebula` suffix, the domain name for my hosts would be `alice-laptop.nebula` and `bob-pc.nebula`

As you point out, it is possible to add the suffix into the `Name` when creating a host certificate, however nebula does not enforce that `Name` is an FQDN, but with the addition of the common suffix is a way of guaranteeing that all hosts have a FQDN.

I think this is a useful addition, however if you don't think it is necessary i will remove it.

dns_server.go
andreasbell: I agree. In general I would say that the host running the lighthouse and DNS server should have the appropriate firewall rules in place so as not to allow DNS traffic to clients outside the nebula network, or the `dns.host` parameter in the config file should be set to only allow nebula traffic.

From testing I also see that the `d.isSelfNebulaOrLocalhost()` does not seem to work for other nebula nodes but will only respond to the DNS requests from localhost. Which is an additional reason that I removed it.

But I can add this back in if you think that would be better?

andreasbell:
Thanks for the feedback @JackDoan I will try to make the proposed changes in the next few weeks. Let me know if you have any further comments.

dns_server.go
andreasbell: Resolved with [6262ef6](https://github.com/slackhq/nebula/pull/1417/commits/6262ef6e353b762c6cd3edc81db150a7d5fb73d8)

dns_server.go
andreasbell: Resolved with [925f802](https://github.com/slackhq/nebula/pull/1417/commits/925f802b74884d5e2ae163640861efb6572e0f3e)

PR #1421, opened 2025-06-16
nebula-cert: support reading CA passphrase from env

halmartin:
This patch extends the `nebula-cert` command to support reading the CA passphrase from the environment variable `CA_PASSPHRASE`.

Currently `nebula-cert` depends in an interactive session to obtain the CA passphrase. This presents a challenge for automation tools like ansible. With this change, ansible can store the CA passphrase in a vault and supply it to `nebula-cert` via the `CA_PASSPHRASE` environment variable for non-interactive signing.

This allows the user to preserve the security of the CA certificate by encrypting it at rest (slackhq/nebula#386) while also enabling the use of automation tools that can securely store and provide the CA password to automate certificate signing operations.

Let me know if there are any technical or stylistic issues with this PR and I'll try my best to fix them :)

### DIFF
--- a/cmd/nebula-cert/ca.go
+++ b/cmd/nebula-cert/ca.go
@@ -173,23 +173,26 @@ func ca(args []string, out io.Writer, errOut io.Writer, pr PasswordReader) error
 
 	var passphrase []byte
 	if !isP11 && *cf.encryption {
-		for i := 0; i < 5; i++ {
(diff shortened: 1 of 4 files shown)


halmartin:
I'm not sure why the status of this hasn't updated, but I have already signed the CLA:
> You already signed the CLA on 2025-06-16

PR #1423, opened 2025-06-25
Use a list for relay IPs instead of a map

brad-defined:
port #1422 to master from release-1.9

Use an ordered list of relay IPs for consistent relay use


jasikpark (approved):
lgtm

nbrownus (approved):

PR #1426, opened 2025-07-01
Darwin udp fix

nbrownus:
Golang stdlib has a situation in darwin where a UDP write can return `EAGAIN` but never sees a writeable event bubble up from `kevent` which causes the write to deadlock, [bug report](https://github.com/golang/go/issues/73919). 

This PR attempts to avoid the deadlock by making our own `sendto` syscall and returning an error when `EAGAIN` is encountered.

The reason I am lifting `sendto` out of the `x/sys` package is because Apple doesn't like folks calling the kernel directly and instead wants them to go through `libSystem` and I would rather not re-implement the trampoline through assembly in our codebase. Doing a direct syscall might also block future releases of mobile nebula.
- https://github.com/golang/go/issues/17490
- https://github.com/golang/go/issues/28984

Side note: Rebind support was added for ipv4 only sockets as well as a matter of completion. 

Co-authored-by: @brad-defined

### DIFF
--- a/udp/udp_darwin.go
+++ b/udp/udp_darwin.go
@@ -7,14 +7,15 @@ package udp
 
 import (
 	"context"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"net"
 	"net/netip"
-	"os"
 	"sync/atomic"
 	"syscall"
 	"time"
+	"unsafe"
 
 	"github.com/sirupsen/logrus"
 	"github.com/slackhq/nebula/config"
@@ -25,11 +26,10 @@ import (
 
 type StdConn struct {
 	*net.UDPConn
-	rc          syscall.RawConn
 	isV4        bool
+	sysFd       uintptr
 	writeErrors atomic.Int32
 	l           *logrus.Logger
-	fd          *os.File
 }
 
 var _ Conn = &StdConn{}
@@ -42,11 +42,20 @@ func NewListener(l *logrus.Logger, ip netip.Addr, port int, multi bool, batch in
 	}
 
 	if uc, ok := pc.(*net.UDPConn); ok {
+		c := &StdConn{UDPConn: uc, l: l}
+
 		rc, err := uc.SyscallConn()
 		if err != nil {
 			return nil, fmt.Errorf("failed to open udp socket: %w", err)
 		}
-		c := &StdConn{UDPConn: uc, rc: rc, l: l}
+
+		err = rc.Control(func(fd uintptr) {
+			c.sysFd = fd
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to get udp fd: %w", err)
+		}
+
 		la, err := c.LocalAddr()
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


udp/udp_darwin.go
```diff
@@ -43,16 +90,138 @@ func NewListenConfig(multi bool) net.ListenConfig {
...
+//go:noescape
+func sendto(s int, buf []byte, flags int, to unsafe.Pointer, addrlen int32) (err error)
+
+func (u *StdConn) WriteTo(b []byte, ap netip.AddrPort) error {
+	var sa unsafe.Pointer
+	var addrLen int32
+
+	if u.isV4 {
+		if ap.Addr().Is6() {
+			return fmt.Errorf("listener is IPv4, but writing to IPv6 remote")
```
brad-defined: This may happen on a per packet basis, maybe better to create a sentinel error and return that here? especially as the error string is static.

udp/udp_darwin.go
```diff
@@ -43,16 +90,138 @@ func NewListenConfig(multi bool) net.ListenConfig {
...
+		}
+
+		if errors.Is(err, syscall.EINTR) {
+			// Write was interrupted, retry
+			continue
+		}
+
+		if errors.Is(err, syscall.EAGAIN) {
+			// Try to suss out whether this is a transient error or something more permanent
+			writeErrors++
```
brad-defined: I hacked up the writeErrors local counter business when I didn't understand what was going on with the UDP failures on Windows. I think in this case, it's OK to have a tight retry loop on EINTR, and for all other errors to immediately return the error to the caller.

I prefer this simpler suggested routine to the more complicated error counter and sleep loop.

It would be nice for an error handler to increment an error counter associated with the socket, but it looks like we don't have any interface counters on non-Linux systems?

udp/udp_darwin.go
```diff
@@ -43,16 +90,138 @@ func NewListenConfig(multi bool) net.ListenConfig {
...
+	h := &header.H{}
+	fwPacket := &firewall.Packet{}
+	nb := make([]byte, 12, 12)
+
+	for {
+		// Just read one packet at a time
+		n, rua, err := u.ReadFromUDPAddrPort(buffer)
 		if err != nil {
-			u.l.WithError(err).Error("Failed to rebind udp socket")
+			u.l.WithError(err).Debug("udp socket is closed, exiting read loop")
```
brad-defined: At least on Windows, we see errors from the receive that are not socket closed errors. We should probably check here, too, whether the error is a socket closed error or not before exiting the routine.
https://github.com/slackhq/nebula/blob/release-1.9/udp/udp_rio_windows.go#L151
```
			if errors.Is(err, net.ErrClosed) {
				return
			}
```

udp/udp_darwin.go
nbrownus: It's unlikely to occur but I agree. This came over from linux where we handle the case that ipv6 has been forcibly disabled on the system.

udp/udp_darwin.go
nbrownus: Works for me as long as the `EAGAIN` issue is not persistent. Without any rate limiting or exit condition we can produce a whole lot of logs.

brad-defined (approved):

brad-defined:
I locally reproduced the mac lockup by installing little snitch and pinging a remote host.

Then I updated to this branch and re-ran the test scenario. There was some chaos in establishing a tunnel with a bunch of write failures "resource temporarily unavailable" (EAGAIN), but it did not lock up, and eventually packets started flowing again. It settled down and established the tunnel.

PR #1428, opened 2025-07-01
Darwin udp fix

nbrownus:
Port of #1426 to the main branch

### DIFF
--- a/udp/udp_darwin.go
+++ b/udp/udp_darwin.go
@@ -17,8 +17,6 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/slackhq/nebula/config"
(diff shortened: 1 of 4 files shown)


udp/udp_darwin.go
```diff
@@ -6,17 +6,61 @@ package udp
 // Darwin support is primarily implemented in udp_generic, besides NewListenConfig
```
jrwren, line 6: I'm late, but I think this comment is no longer true?

brad-defined (approved):

udp/udp_darwin.go
nbrownus: It is, nice catch. Will remove in #1427

PR #1437, opened 2025-07-24
netlink: ignore route updates with no destination

wadey:
Currently we assume each route update must have a destination, but we should check that it is set before we try to use it.

See: #1436


jrwren (approved):

brad-defined (approved):
It'd be great to have some tests around this, but that looks like a larger lift.

PR #1441, opened 2025-08-04
add more tests around bits counters

wadey:
Trying to see if the logic around "lost" packets is correct. This test should fail as it misses packet 8

### DIFF
--- a/bits_test.go
+++ b/bits_test.go
@@ -212,6 +212,39 @@ func TestBitsLostCounter(t *testing.T) {
 	assert.Equal(t, int64(36), b.lostCounter.Count())
 	assert.Equal(t, int64(0), b.dupeCounter.Count())
 	assert.Equal(t, int64(0), b.outOfWindowCounter.Count())
+
+	b = NewBits(10)
+	b.lostCounter.Clear()
+	b.dupeCounter.Clear()
+	b.outOfWindowCounter.Clear()
+
+	assert.True(t, b.Update(l, 0))
+	assert.True(t, b.Update(l, 4))
+	assert.True(t, b.Update(l, 1))
+	assert.True(t, b.Update(l, 9))
+	assert.True(t, b.Update(l, 2))
+	assert.True(t, b.Update(l, 3))
+	assert.True(t, b.Update(l, 5))
+	assert.True(t, b.Update(l, 6))
+	assert.True(t, b.Update(l, 7))
+	// assert.True(t, b.Update(l, 8))
+	assert.True(t, b.Update(l, 10))
+	assert.True(t, b.Update(l, 11))
+
+	assert.True(t, b.Update(l, 14))
+	assert.True(t, b.Update(l, 19))
+	assert.True(t, b.Update(l, 12))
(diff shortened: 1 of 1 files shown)


nbrownus:
I fixed the bug and updated the code to simplify and try to explain more betterer whats going on.

bits.go
```diff
@@ -59,61 +60,26 @@ func (b *Bits) Update(l *logrus.Logger, i uint64) bool {
 		return true
 	}
 
-	// If i packet is greater than current but less than the maximum length of our bitmap,
-	// flip everything in between to false and move ahead.
-	if i > b.current && i < b.current+b.length {
-		// In between current and i need to be zero'd to allow those packets to come in later
-		for n := b.current + 1; n < i; n++ {
-			b.bits[n%b.length] = false
```
nbrownus, line 67: The original issue was not tracking lost for these zeroed out counter values

bits.go
```diff
@@ -9,111 +9,84 @@ type Bits struct {
...
-		return true
-	}
-
-	// If i is within the window of current minus length (the total pat window size),
-	// allow it and flip to true but to NOT change current. We also have to account for the first window
-	if ((b.current >= b.length && i > b.current-b.length) || (b.current < b.length && i < b.length)) && i <= b.current {
-		if b.current == i {
+	// If i is within the current window but below the current counter,
+	// Check to see if it's a duplicate
+	if i > b.current-b.length || i < b.length && b.current < b.length {
```
JackDoan, line 88: this statement frightens the C programmer

nbrownus: Can break it back out again

JackDoan: eh I think it's find, the most I'd do is add some `()` around things to strongly communicate order

but golang already has sane evaluation order so we probably don't even need to do that

bits.go
```diff
@@ -9,111 +9,84 @@ type Bits struct {
...
 }
 
 func (b *Bits) Update(l *logrus.Logger, i uint64) bool {
 	// If i is the next number, return true and update current.
 	if i == b.current+1 {
-		// Report missed packets, we can only understand what was missed after the first window has been gone through
-		if i > b.length && b.bits[i%b.length] == false {
+		// Check if the oldest bit was lost since we are shifting the window by 1 and occupying it with this counter
+		// The very first window can only be tracked as lost once we are on the 2nd window or greater
+		if b.bits[i%b.length] == false && i > b.length {
```
jrwren, line 56: any particular reason the left and ride side of && are swapped from what they were before? I don't know how good the Go compiler is at optimizing this, but at first glance it seems like i > b.length would be the less costly operation and would be good to do first?

JackDoan (approved):

bits.go
jasikpark: I'm surprised go fmt doesn't add the parentheses for you, I appreciate that prettier adds them to make evaluation order explicitly obvious.

bits.go
nbrownus: Silly micro-op since `i > b.length` is more than likely always true. Our window size is 1024 currently, after which every packet would pass that part of the if statement.

PR #1453, opened 2025-08-19
Quietly log error on UDP_NETRESET ioctl on Windows.

brad-defined:
Windows is returning errors on UDP_NETRESET ioctls now. This code was inspired by golang stdlib code, and included as a best effort to minimize the number of errors returned by Windows udp receive operations.

When this fails, Nebula falls back to the generic UDP code.

This PR updates the ioctls to quietly log failures and continue processing.

I manually tested this and I do not see errors being returned by the udp receive operation when the Windows host receives icmp unreachable messages.

### DIFF
--- a/udp/udp_rio_windows.go
+++ b/udp/udp_rio_windows.go
@@ -69,7 +69,7 @@ func NewRIOListener(l *logrus.Logger, addr netip.Addr, port int) (*RIOConn, erro
 
 	u := &RIOConn{l: l}
 
-	err := u.bind(&windows.SockaddrInet6{Addr: addr.As16(), Port: port})
+	err := u.bind(l, &windows.SockaddrInet6{Addr: addr.As16(), Port: port})
 	if err != nil {
 		return nil, fmt.Errorf("bind: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


udp/udp_rio_windows.go
```diff
@@ -152,7 +157,7 @@ func (u *RIOConn) ListenOut(r EncReader, lhf LightHouseHandlerFunc, cache *firew
 				u.l.WithError(err).Debug("udp socket is closed, exiting read loop")
 				return
 			}
-			u.l.WithError(err).Error("unexpected udp socket receive error")
+			u.l.WithError(err).Warn("unexpected udp socket receive error")
```
nbrownus: I suspect this can get really noisy if we failed to disable the icmp error reporting on a busy host. 

Are we able to do any error equality checking and suppress logging when we encounter those icmp errors?

brad-defined: oh sorry, I got deep into trying to repro scenarios and forgot I promised to dampen this. Let me see what I can do here.

brad-defined:
@nbrownus I failed to generate ICMP TTL errors and get them to actually be received by the host. In lieu of that, I hacked up some dampening code so that the unexpected warnings will not be output any more often than once per minute.

nbrownus (approved):

Issue #1398, opened 2025-04-28
Feature: Expose Nebula API as a socket on the local machine or Nebula interface


johnmaguire:
We have an existing [SSH debug server](https://nebula.defined.net/docs/config/sshd/) that can be used to inspect and tamper with the running Nebula process (e.g. checking or changing which remote is used for a given node's tunnel or printing internal process information.) This tool can be dangerous which is why we protect it with SSH authentication. However, we often hear and experience that this is a bit painful to setup: you have to generate a server SSH key and add client SSH keys for each user who wants to connect.

We could simplify this by exposing the API as a local socket protected by Unix permissions and/or a remote socket protected via the Nebula firewall (i.e. using groups to restrict access.)

We are leaning towards an implementation that uses HTTP as a transport layer and JSON as a serialization format. We believe that this would make consumption trivial, since most machines have cURL or similar tooling installed. cURL can make HTTP requests to Unix sockets since 7.40 (released Feb 2015.) It also would allow users to use off-the-shelf tools like nginx to add their own authentication layer and broker access to the API if they desire.

Some other ideas that have been discussed:

- Additional event-based debugging (i.e. "watching for changes")
- Additional webhooks APIs for event-based reactivity (e.g. fail2ban-like functionality? Or logging if an unexpected host tries to access a privileged port.)

We're seeking input from the community on this feature.

johnmaguire:
Hey @oix7 I'm not sure I see how these issues are related. Can you share a little more? Thanks!

Issue #920, opened 2023-07-20
Prometheus service discovery


benbro:
Prometheus supports [dynamic service discovery](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
Is there a way Prometheus could automatically discover all registered nodes from a lighthouse? The certificate groups could be returned as labels to help filter and [relabel config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config).
Something like [consul_sd_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#consul_sd_config) will be great.

benbro:
I'll try it. thanks.

benbro:
Using sshd will work but it seems to be the wrong tool for this task. [Docs](https://nebula.defined.net/docs/config/sshd/) says "sshd enables nebula's built-in debugging console". It exposes functions I don't want to expose like change-remote and profiling and it requires managing a new set of ssh keys in addition to the nebula certificates.
Isn't it possible to expose an http server that on the nebula overlay that will output metadata or at least a command line interface that won't require ssh.

nbrownus:
I agree the ssh interface exposes too much for you goals here. We have talked about adding an allow list config option to limit what a given user can accomplish via the ssh interface, would that suffice?

benbro:
The ssh interface is also a little more cumbersome to use because it requires additional ssh keys. Can't it use the existing nebula overlay? I don't think I'll use an ssh interface to get static metadata about the network so feel free to ignore my feature request. Thank you for taking the time to consider it.

nbrownus:
The main trouble is helping folks avoid accidental information disclosure. Exposing a hostmap list over http with the current capabilities of nebula would at best allow any user on the computer where its enabled to access the hostmap list. It could give a would-be attacker a treasure map for lateral movement within the environment. The ssh system bulks this up by requiring authentication prior to use, an attacker has to gain root access to the system or a private key for an already authorized ssh key.

That's not to say there isn't value in the idea and we won't do it. We also have #913 to consider which brings about similar challenges and if we determine to merge it then I would imagine we would desire to add more capability to the http server. The trick will be striking the right balance between usability and security.

benbro:
Can't I already discover all IPs when lighthouse dns is enabled?
```
nmap -sL --dns-server=100.64.64.9 100.64.64.0/24
dig @100.64.64.9 +short "100.64.64.10" TXT
```

PR #1457, opened 2025-08-26
Feature: Expose Nebula API as a socket

HenryGrahamRivian:
This MR introduces an info server socket to serve information about the Hostmap. This is in reference to #1398. 

<details>
<summary> Required config file additions </summary>

```yaml
info:
  listen: 127.0.0.2:5555
```
</details>

<details>
<summary> Example curls</summary

```
curl http://127.0.0.2:5555/host/100.100.100.100
{"curve":"P256","details":{"groups":["group1","group2"],"isCa":false,"issuer":"6a20d559ef58a72a2c8e599284fc11bfb95a61b21cd55ec51c0bdd59b4582537","name":"demo","networks":["100.100.100.100/11"],"notAfter":"2026-11-29T11:22:32-06:00","notBefore":"2025-1-03T12:00:11-05:00","unsafeNetworks":["0.0.0.0/0"]},"fingerprint":"c175687fec1ace05821b77b39f4b7520d1839e948fb7140570c0c4ffb99ac67f","publicKey":"77de6cf8a3541e527b6517532d36c0b3e01cdfcf54fe5270b7538881d5886e1885bbaa05b919a17b1d2b7dd49b5bfb103725ef1cf25c6a6212e00943caf9d8191e","signature":"9ec4e33140a71ee6e3c5ccb305f8920c3445d868b24b58f5217f84928e4dca2b8a408c32d108c5b99af0bf1f395a4468b4d21690d1fad1591a216e48d0da5a8c7726f2b34f54","version":2}% 

curl http://127.0.0.2:5553/hostmap
{"demo":{"vpnAddrs":["100.100.100.100"],"lastHandshakeTime":"2025-08-26T09:07:13.609073708-05:00","groups":["demo1","demo2"]},{"demo2":{"vpnAddrs":["100.100.100.101"],"lastHandshakeTime":"2025-08-26T09:07:13.609073708-05:00","groups":["demo3","demo4"]}}
```
</details>

### DIFF
--- a/control.go
+++ b/control.go
@@ -35,6 +35,7 @@ type Control struct {
 	dnsStart               func()
 	lighthouseStart        func()
 	connectionManagerStart func(context.Context)
+	infoStart       func()
 }
 
 type ControlHostInfo struct {
@@ -69,6 +70,9 @@ func (c *Control) Start() {
 	}
 	if c.lighthouseStart != nil {
 		c.lighthouseStart()
+	}
+	if c.infoStart != nil {
+		go c.infoStart()
 	}
 
 	// Start reading packets.
--- a/info.go
+++ b/info.go
@@ -1,0 +1,109 @@
+package nebula
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/netip"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/slackhq/nebula/config"
+)
+
+func handleHostmapList(l *logrus.Logger, hm *HostMap, w http.ResponseWriter, r *http.Request) {
+	type HostListItem struct {
+		VpnAddrs []netip.Addr `json:"vpnAddrs"`
+		//Remote            netip.AddrPort `json:"remote"`
+		Relayed           bool      `json:"relayed,omitempty"`
+		LastHandshakeTime time.Time `json:"lastHandshakeTime"`
+		Groups            []string  `json:"groups"`
+	}
+
+	out := map[string]HostListItem{}
+	hm.ForEachVpnAddr(func(hi *HostInfo) {
+		cert := hi.GetCert().Certificate
+		out[cert.Name()] = HostListItem{
+			VpnAddrs: hi.vpnAddrs,
+			//Remote:            hi.remote,
+			Relayed:           !hi.remote.IsValid(),
+			LastHandshakeTime: time.Unix(0, int64(hi.lastHandshakeTime)),
+			Groups:            cert.Groups(),
+		}
+	})
+
+	w.Header().Set("Content-Type", "application/json")
+	js := json.NewEncoder(w)
+	err := js.Encode(out)
+	if err != nil {
+		http.Error(w, "json error: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
+
+func handleHostCertLookup(l *logrus.Logger, hm *HostMap, w http.ResponseWriter, r *http.Request) {
+	ipStr := r.PathValue("ipStr")
+	if ipStr == "" {
+		http.Error(w, "you must provide an IP address", http.StatusNotFound)
+		return
+	}
+
+	addr, err := netip.ParseAddr(ipStr)
+	if err != nil {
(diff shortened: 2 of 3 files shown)


johnmaguire:
Thanks for the contribution. The SSH debug server is definitely a bit painful to setup and I would love to see an easier-to-use debug interface merged. Some background:

- The SSH server requires authentication which is somewhat painful to configure, but does offer some security value. In #1398 we discussed potentially protecting the HTTP API via Nebula firewall rules. If that is the plan, we might want to avoid a listen host and instead only listen on the Nebula IP.
- Another potential option for securing this API would be HTTP Basic Auth, with either a username/password combo (perhaps ideal for logs) or simple Bearer tokens.
- Another idea floated in #1398 was to only expose this as a Unix socket protected by file system level permissions - cURL supports querying Unix domain sockets since 2015, but to expose them remotely you'd need to create a separate forwarding binary.

That said, on another ticket related to Prometheus service discovery, a user made the point that the DNS server already allows exposing some information publicly without authentication: https://github.com/slackhq/nebula/issues/920#issuecomment-1656245370

How were you thinking about security around this API?

JackDoan:
> How were you thinking about security around this API?

https://github.com/slackhq/nebula/pull/1457/files#diff-f17225d549b302ca95e2dc524a83c9ad067923eb49ef688e8093c6d45a3e7384R86 there's a todo to refuse (or maybe just warn?) if you bind to something _silly_. Refusing to bind to (non-localhost || non-nebula-ip) feels right to me.

IMO, this HTTP API would be informational, a-la DNS. The driving use-case is to make "nebula aware" applications that can use stuff like group information to make authorization decisions. (such as, you need the `api-user` group to get past the firewall and talk to the app, but `api-superuser` lets you call other endpoints, without requiring the app to bind on two ports)

^ this is what makes binding to localhost valuable imo. If we only bound to the nebula IP, applications on the same host would need to discover the nebula IP to use it, which isn't _hard_, but it's less-pleasant.

HenryGrahamRivian:
I'd be happy to implement the TODO for interface binding. Personally, I think there is value in having the option to bind to non-localhost interfaces (for example, a proxy terminating nebula before the service in Jacks example) so I'll be starting out with the warning implementation. I'd love to hear other opinions though.

nbrownus:
We discussed this PR a little today and here are the takeaways
1. The built-in `ssh` server is neither easy to configure or friendly to use as an api.
2. There continues to be a desire to have _some other_ method for probing `nebula`.
3. There is a general discomfort in having the method be unauthenticated.
4. There is a general desire in having the method require as few required libraries as possible to make functional as an api.

One option we keep coming back to is a simple unix domain socket, likely in `SOCK_STREAM` mode, communicating entirely via plain JSON. This provides an api that requires very few libraries and is about as straightforward as you can get to interact with. Security relies entirely on the local filesystem permissions which has decent properties assuming the socket can't be created by a less privileged user than the one running `nebula`. `SO_PEERCRED` or `SCM_CREDENTIALS` can be used to further raise the level of confidence the client has that it is speaking with `nebula` directly and could be used by `nebula` as well to restrict which programs it would willingly interact with, this is not 100% portable though. 

Another option would be to do plain JSON over HTTP, which is a bit more ergonomic but raises the clients required library complexity and includes attack surface for the client if it were to be speaking with an adversarial `nebula` or man-in-the middle.

There are many other possibilities as well.

The argument to having some authentication method mainly comes down to a defense-in-depth strategy. By requiring _something_ here we should be able to avoid low effort drive-by attacks but obviously anyone with `root` on the machine could nullify the entire point.

In summary: It remains to be unclear what specifically we are trying to accomplish with the api and what amount of security we should try to sprinkle on it, how ergonomic it should or shouldn't be, or how much the person configuring it should need to understand it to avoid an information disclosure foot gun.

HenryGrahamRivian:
This was stuck in my head so I put together a rough picture of what this approach might look like. If this is more the direction y'all were thinking I'll go ahead and clean this up.


<details>
<summary> Example usage </summary>

```
echo '{"command":"hostmap"}' | sudo nc -U /tmp/nebula.sock
```
</details>

PR #1459, opened 2025-08-27
Fix `recv_error` receipt limit allowance for v1.9.x

nbrownus:
Since #955 (v1.8.0) we have changed from dropping a tunnel on the 4th `recv_error` receipt to dropping a tunnel on the 1st `recv_error` receipt. This change reverts to the pre-v1.8.0 behavior.

Ultimately, I think it would be best to remove `recv_error` messages entirely and rely on `tunnels.drop_inactive` (which landed in v1.9.6 and `test` packets to sort out tunnel issues). I am submitting this change to v1.9.x in case we do not want to take on that deprecation at this point but would argue that, at worst, we should in v1.10.


brad-defined (approved):

brad-defined:
oh, tests failed. I thought I saw all green checks when I approved...sorry.

brad-defined:
> oh, tests failed. I thought I saw all green checks when I approved...sorry.

I re-ran the tests and they worked. yay.

brad-defined (approved):

PR #1463, opened 2025-09-02
don't send recv errors for packets outside the connection window anymore

JackDoan:
Sending a recv_error on window mismatches is mostly unhelpful

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -17,12 +17,10 @@ import (
 	"github.com/slackhq/nebula/header"
 )
 
-// const ProbeLen = 100
 const defaultPromoteEvery = 1000       // Count of packets sent before we try moving a tunnel to a preferred underlay ip address
 const defaultReQueryEvery = 5000       // Count of packets sent before re-querying a hostinfo to the lighthouse
 const defaultReQueryWait = time.Minute // Minimum amount of seconds to wait before re-querying a hostinfo the lighthouse. Evaluated every ReQueryEvery
 const MaxRemotes = 10
-const maxRecvError = 1 // TODO this was 4, but changed to 1 to match 1.9.x behavior. Config option?
 
 // MaxHostInfosPerVpnIp is the max number of hostinfos we will track for a given vpn ip
 // 5 allows for an initial handshake and each host pair re-handshaking twice
@@ -225,8 +223,7 @@ type HostInfo struct {
 	// vpnAddrs is a list of vpn addresses assigned to this host that are within our own vpn networks
 	// The host may have other vpn addresses that are outside our
 	// vpn networks but were removed because they are not usable
-	vpnAddrs  []netip.Addr
-	recvError atomic.Uint32
+	vpnAddrs []netip.Addr
 
 	// networks are both all vpn and unsafe networks assigned to this host
 	networks   *bart.Lite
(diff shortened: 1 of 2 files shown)


outside.go
```diff
@@ -255,18 +255,17 @@ func (f *Interface) handleHostRoaming(hostinfo *HostInfo, udpAddr netip.AddrPort
...
-	if ci == nil || !ci.window.Check(f.l, h.MessageCounter) {
-		if addr.IsValid() {
-			f.maybeSendRecvError(addr, h.RemoteIndex)
-			return false
-		} else {
-			return false
-		}
+	// If connectionstate exists, process packet
+	// Else, send recv error to allow fast reconnection.
+	if ci != nil {
```
nbrownus: We still want to avoid handling the encrypted packet if the window check fails, just not bother sending the recv_error. 

There is also the accompanying bug in hostmap.go that #1459 speaks to

JackDoan: > We still want to avoid handling the encrypted packet if the window check fails, just not bother sending the recv_error.

oh crap, good catch, thanks

> There is also the accompanying bug in hostmap.go that https://github.com/slackhq/nebula/pull/1459 speaks to

do we still actually want this? Wanting 4 packets hurts the fast-reconnect story

nbrownus: Well the current behavior increments a counter and ignores it so I'd just remove the function entirely. Honestly would take a big hammer to a lot of this.

hostmap.go
```diff
@@ -22,7 +22,7 @@ const defaultPromoteEvery = 1000       // Count of packets sent before we try mo
 const defaultReQueryEvery = 5000       // Count of packets sent before re-querying a hostinfo to the lighthouse
 const defaultReQueryWait = time.Minute // Minimum amount of seconds to wait before re-querying a hostinfo the lighthouse. Evaluated every ReQueryEvery
 const MaxRemotes = 10
-const maxRecvError = 4
```
JackDoan, line 25: @nbrownus Even though #1459 was a bug, I think correcting the logic but keeping the behavior is the way to go, thoughts?

hostmap.go
```diff
@@ -734,10 +734,7 @@ func (i *HostInfo) SetRemoteIfPreferred(hm *HostMap, newRemote netip.AddrPort) b
 }
 
 func (i *HostInfo) RecvErrorExceeded() bool {
-	if i.recvError.Add(1) >= maxRecvError {
-		return true
-	}
-	return true
+	return i.recvError.Add(1) >= maxRecvError
```
JackDoan: from #1459

outside.go
JackDoan: I also had that temptation but didn't want to do it if folks really wanted the old behavior back (but I think we've been functionally on `maxRecvError = 1` for at least two+ years)

nbrownus (approved):
LGTM

The remaining missing things are
1. Config option to disable handling the receipt of a `recv_error`
2. Functionality to support `recv_error` for a relayed index
3. Since this is the desired direction currently, we should apply this change to `release-1.9` instead of #1459 

The first one looks like an easy fast follow on this PR. The second one is a bit trickier since we need to modify the `recv_error` packet to indicate its a relay index.

PR #1464, opened 2025-09-03
Don't delete static host mappings for non-primary IPs

### LINKED ISSUE #1447: 🐛 BUG: Nebula may lose a static host mapping

master branch

Linux

### Describe the Bug

@JackDoan reported a host losing its static host mapping.

The config of Nebula included a static host mapping for one IP of the host. The host's certificate includes multiple IPs. Spelunking through the code, the function DeleteVpnAddrs is updated in master to support a host with multiple IPs. However, that function will only skip address deletion if the first IP in the certificate is recorded as having a static host mapping entry. We believe that in the user's reported case, the first IP in the host cert was not the IP listed in the config's static host mapping.

Reading through the code to investigate this issue, I noticed that unlockedGetRemoteList may also overwrite a static mapping if it receives a list of IP's and if the first IP in the slice is not in the addrMap but any subsequent IP is in the map. In that case, the subsequent IP entries will be overwritten by the newly created entry.

To close this issue, investigate the lighthouse addrMap to cover the scenario in which a given host's secondary IP appears in the static host map.

### Logs from affected hosts

```

```

```

```

JackDoan:
Fixes #1447 

- [x] The config of Nebula included a static host mapping for one IP of the host. The host's certificate includes multiple IPs. Spelunking through the code, the function DeleteVpnAddrs is updated in master to support a host with multiple IPs. However, that function will only skip address deletion if the first IP in the certificate is recorded as having a static host mapping entry. We believe that in the user's reported case, the first IP in the host cert was not the IP listed in the config's static host mapping.

- [x] Reading through the code to investigate this issue, I noticed that unlockedGetRemoteList may also overwrite a static mapping if it receives a list of IP's and if the first IP in the slice is not in the addrMap but any subsequent IP is in the map. In that case, the subsequent IP entries will be overwritten by the newly created entry.
 
- [x] To close this issue, investigate the lighthouse addrMap to cover the scenario in which a given host's secondary IP appears in the static host map.

- [x] Add tests to prove this is fixed!

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -519,11 +519,15 @@ func (lh *LightHouse) queryAndPrepMessage(vpnAddr netip.Addr, f func(*cache) (in
 }
 
 func (lh *LightHouse) DeleteVpnAddrs(allVpnAddrs []netip.Addr) {
-	// First we check the static mapping
-	// and do nothing if it is there
-	if _, ok := lh.GetStaticHostList()[allVpnAddrs[0]]; ok {
-		return
+	// First we check the static host map. If any of the VpnAddrs to be deleted are present, do nothing.
+	staticList := lh.GetStaticHostList()
+	for _, addr := range allVpnAddrs {
+		if _, ok := staticList[addr]; ok {
+			return
+		}
 	}
+
+	// None of the VpnAddrs were present. Now we can do the deletes.
 	lh.Lock()
 	rm, ok := lh.addrMap[allVpnAddrs[0]]
 	if ok {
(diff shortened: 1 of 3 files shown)


lighthouse.go
```diff
@@ -630,13 +634,18 @@ func (lh *LightHouse) addCalculatedRemotes(vpnAddr netip.Addr) bool {
...
 // assumes you have the lh lock
 func (lh *LightHouse) unlockedGetRemoteList(allAddrs []netip.Addr) *RemoteList {
-	am, ok := lh.addrMap[allAddrs[0]]
-	if !ok {
-		am = NewRemoteList(allAddrs, func(a netip.Addr) bool { return lh.shouldAdd(allAddrs[0], a) })
-		for _, addr := range allAddrs {
-			lh.addrMap[addr] = am
+	// before we go and make a new remotelist, we need to make sure we don't have one for any of this set of vpnaddrs yet
+	for _, addr := range allAddrs {
+		am, ok := lh.addrMap[addr]
```
JackDoan, line 638: in the case where there is a remotelist for `allAddrs[1]` and not for `allAddrs[0]`, this will return the remotelist for `allAddrs[1]` and _not_ create an index in `addrMap` for `allAddrs[0]`. As I'm poking around, I can't tell where we'd ever fill it in if it didn't exist yet (which is a condition `static_host_map` could cause).

nbrownus (approved):

PR #1466, opened 2025-09-04
cert-v2 chores

JackDoan:


lighthouse.go
```diff
@@ -1263,7 +1267,6 @@ func (lhh *LightHouseHandler) handleHostUpdateNotification(n *NebulaMeta, fromVp
 		return
 	}
 
-	//TODO: CERT-V2 hosts with only v2 certs cannot provide their ipv6 addr when contacting the lighthouse via v4?
```
JackDoan, line 1266: this was fixed near line 770

overlay/tun_linux.go
```diff
@@ -293,7 +293,6 @@ func (t *tun) addIPs(link netlink.Link) error {
 
 	//add all new addresses
 	for i := range newAddrs {
-		//TODO: CERT-V2 do we want to stack errors and try as many ops as possible?
```
JackDoan, line 296: I have decided no. If we can't add our addresses, we should abort asap.

pki.go
```diff
@@ -173,7 +173,6 @@ func (p *PKI) reloadCerts(c *config.C, initial bool) *util.ContextualError {
 
 	p.cs.Store(newState)
 
-	//TODO: CERT-V2 newState needs a stringer that does json
```
JackDoan, line 176: CertState has one!

lighthouse.go
```diff
@@ -710,14 +710,10 @@ func (lh *LightHouse) unlockedShouldAddV6(vpnAddr netip.Addr, to *V6AddrPort) bo
...
 func (lh *LightHouse) IsLighthouseAddr(vpnAddr netip.Addr) bool {
-	if _, ok := lh.GetLighthouses()[vpnAddr]; ok {
-		return true
-	}
-	return false
+	_, ok := lh.GetLighthouses()[vpnAddr]
+	return ok
 }
 
-// TODO: CERT-V2 IsLighthouseAddr should be sufficient, we just need to update the vpnAddrs for lighthouses after a handshake
```
JackDoan, line 719: I think it is more burdensome to do that? This feels fine to me, but I'm open to making the change if you feel differently

nbrownus: The goal of this TODO was to remove the loop and made a stab at how we might do it. 

#1464 gave us a more reliable `lh.addrMap` for non primary vpn addrs in the static host map and a TODO in that PR was to update the `RemoteList.vpnAddrs` once we received and authenticated the certificate from the remote (to fix the `remote_allow_ranges`). With that, we would have a way to kill this loop.

lighthouse.go
```diff
@@ -1199,8 +1188,12 @@ func (lhh *LightHouseHandler) coalesceAnswers(v cert.Version, c *cache, n *Nebul
 			}
 
 		} else {
-			//TODO: CERT-V2 don't panic
-			panic("unsupported version")
+			if lhh.l.Level >= logrus.DebugLevel {
+				lhh.l.WithField("version", v).Debug("unsupported protocol version")
+			}
+			for _, r := range c.relay.relay { //same handling as V2, for optimistic compatibility
```
nbrownus: I think we can get by with just the debug log, which should be impossible to hit since this version the code wouldn't be able to handshake with a nebula cert version that isn't 1 or 2.

lighthouse.go
JackDoan: Ahhh I like that.

Do you feel this is a 1.10 blocker?

nbrownus: Not specifically, it is a performance regression in some ways but I think it has low impact.

JackDoan: oh actually this is easier than I thought, I think. Pushed something that I think takes care of it

JackDoan: scooting this to new MR so we can merge everything else

PR #1467, opened 2025-09-05
update RemoteList.vpnAddrs when we complete a handshake

JackDoan:


nbrownus (changes requested):
I think we should also change lighthouse.go:648 to reference the remote lists vpnAddrs instead of holding a copy of the zeroth addr

lighthouse.go
```diff
@@ -645,18 +645,17 @@ func (lh *LightHouse) unlockedGetRemoteList(allAddrs []netip.Addr) *RemoteList {
 		}
 	}
 
-	//TODO lighthouse.remote_allow_ranges is almost certainly broken in a multiple-address-per-cert scenario
-	am := NewRemoteList(allAddrs, func(a netip.Addr) bool { return lh.shouldAdd(allAddrs[0], a) })
+	am := NewRemoteList(allAddrs, func(a netip.Addr) bool { return lh.shouldAdd(allAddrs, a) })
```
nbrownus: The `shouldAdd` argument in `NewRemoteList` should change to `func([]netip.Addr, netip.Addr) bool` where the first argument is `r.vpnAddrs` and the 2nd argument remains to be the underlay address we are testing.

This line should probably just look like `am := NewRemoteList(allAddrs, lh.shouldAdd)` unless you see a more clever way to inject it.

JackDoan: OH I get it now. Thanks for the pointer.

remote_list.go
```diff
@@ -368,6 +370,14 @@ func (r *RemoteList) CopyBlockedRemotes() []netip.AddrPort {
 	return c
 }
 
+// RefreshFromHandshake locks and updates the RemoteList to account for data learned upon a completed handshake
+func (r *RemoteList) RefreshFromHandshake(vpnAddrs []netip.Addr) {
+	r.Lock()
+	r.badRemotes = nil
+	r.vpnAddrs = vpnAddrs //should this be remade and copied, a-la NewRemoteList?
```
nbrownus: Good call, the copies came about a while back after a memory reuse bug lead to corrupted lighthouse address maps. I don't think we get tricky with the vpnAddrs but it would be good to copy and avoid the situation entirely.

nbrownus (approved):

nbrownus:
I think there is a follow up item here which is to update the documentation to cover both ipv4 and ipv6 overlay addresses with the allow lists. You can get in a situation where you think you've only allowed a certain ipv4 range but if you don't disallow the ipv6 range on the underlay then you might be in for a surprise.

PR #1468, opened 2025-09-08
try to make certificate addition/removal reloadable in some cases

JackDoan:


connection_manager.go
```diff
@@ -488,18 +488,38 @@ func (cm *connectionManager) isInvalidCertificate(now time.Time, hostinfo *HostI
...
+		out = true
+	}
 
-	return true
+	if out {
+		hostinfo.logger(cm.l).WithField("fingerprint", remoteCert.Fingerprint).
+			WithField("version", remoteVersion).
+			Info("We no longer have a certificate in common with remote, tearing down the tunnel")
+	}
+	return out
```
JackDoan: Is this the right place to do this? Would it be better to pull this out into its own function in connectionmanager?

nbrownus: We have 2 conditions to cover here:
1. A certificate at a given version was removed from the process (either through reload or expiry). It does feel correct to tear down these tunnels but we may also have a certificate with a different version that we could quickly re-handshake to the remote with instead of disrupting traffic.

2. The current version to use changes (from 1 to 2 or 2 to 1). This is largely covered by `cm.tryRehandshake` and its goal was to facilitate re-handshaking after a host cert renewal, but it's not entirely sufficient for moving between cert versions.

For graceful re-handshaking, I think it makes sense if we want to limit it to from v1 -> v2 since the set of overlay addresses can only increase there. Going from v2 -> v1 is a bit tricky since the addresses that were primarily in use might not be available in the v1 certificate and re-handshaking may break the tunnel as a result.

Basically, we could make this a little bit more nuanced, allow for graceful tunnel upgrades from v1 -> v2 certificates just as we would if we refreshed the current current certificate and forcibly close tunnels that don't have a matching certificate version.

0. Start with v1/2, load a new v1/2 cert: handled today by `cm.tryRehandshake`
1. Start with v1, load a v2 cert: Try to upgrade with `cm.tryRehandshake`
2. Start with v2, load a v1 cert: Do nothing
3. Start with v1 and v2, remove the v1 cert: Try to upgrade v1 tunnels with `cm.tryRehandshake`? Eventually they do need to close.
4. Start with v1 and v2, remove the v2 cert: Kill v2 tunnels

Of course we'd need to follow the configuration for `pki.initiating_version` in this table.

And finally, if you want to keep some of that functionality here you'll need to remove the
```
if err == nil {
		return false
```
to allow for a signed, trusted, and non-expired host cert to be checked.

handshake_ix.go
```diff
@@ -167,12 +167,11 @@ func ixHandshakeStage1(f *Interface, addr netip.AddrPort, via *ViaSender, packet
 		if rc == nil {
 			f.l.WithError(err).WithField("udpAddr", addr).
 				WithField("handshake", m{"stage": 1, "style": "ix_psk0"}).WithField("cert", remoteCert).
-				Info("Unable to handshake with host due to missing certificate version")
-			return
```
JackDoan, line 171: This change makes this flow actually work

`them(v1+v2) ----handshake(v1)---> me (v2 only)`

a potential drawback is if `them` is actually an old client with only v1 certs, we may make it upset?

with this current flow, we'll also create a connectionstate with mismatched cert versions. I think this is actually fine? But maybe we should try to trigger a rehandshake when the connectionmanager sees that, so everyone can know everyone's VPN IPs?

JackDoan: I think yes. I added that.

JackDoan: hmm maybe we should only bother rehandshaking if there's a meaningful difference though? (like different number of VPN-IPs)

nbrownus: > a potential drawback is if them is actually an old client with only v1 certs, we may make it upset?

I think this is all of the problem, how does the remote behave when encountered with an unknown certificate version. I think anything pre 1.9.3ish is going to have a bad time.

> hmm maybe we should only bother rehandshaking if there's a meaningful difference though? (like different number of VPN-IPs)

The only thing we can safely make a determination about is our own local state. Is this tunnel using a certificate at the current configured version level? Is the certificate the latest one available? I would be very wary of trying to reason about the remote state in a re-handshake flow.

JackDoan: > The only thing we can safely make a determination about is our own local state. Is this tunnel using a certificate at the current configured version level? Is the certificate the latest one available? I would be very wary of trying to reason about the remote state in a re-handshake flow.

Exactly, I agree. What I was referring to here was:

1. I define two parties, one named Host, with v1+v2 certs, and Peer, with only a v2 cert
2. Host handshakes to Peer. Host uses a V1 cert, as this is the default, most-compatible behavior
3. Peer doesn't have a V1 cert. Peer replies to Host with a V2 cert.
4. Host accepts Peer's V2 cert
5. Host now has a decision to make:
    a.  if Host's `certv1.Networks() == certv2.Networks()`, (imo), there's no reason to do anything. Just use this tunnel.
    b. else, (imo) it's worth having Host trigger a rehandshake with its V2 cert, so Peer will be informed of all of Host's VPN IPs. This is important on mixed-dual-stack networks for things like relays to behave reliably (and dual stack configurations that lack a dual-stack lighthouse!)
    c. **currently** this PR always chooses path (b)


All the other rehandshake-flow bits should remain in place as they always have. I think my phrasing was poor: I meant "bother rehandshaking when scenario 5.b applies"

JackDoan: > I think this is all of the problem, how does the remote behave when encountered with an unknown certificate version. I think anything pre 1.9.3ish is going to have a bad time.

We should specifically test this to be sure, but shouldn't it just drop the packet? We'll error unmarshalling the not-proto in stage 1 and delete the handshake-hostinfo.

This class of problem _should_ also only crop up when an un-updated, v1-only host, attempts to contact a v2-only host, I think?

connection_manager.go
```diff
@@ -488,18 +491,38 @@ func (cm *connectionManager) isInvalidCertificate(now time.Time, hostinfo *HostI
...
+		hostinfo.logger(cm.l).WithError(err).
+			WithField("fingerprint", remoteCert.Fingerprint).
+			Info("Remote certificate is no longer valid, tearing down the tunnel")
+		return true
 	}
 
-	hostinfo.logger(cm.l).WithError(err).
-		WithField("fingerprint", remoteCert.Fingerprint).
-		Info("Remote certificate is no longer valid, tearing down the tunnel")
+	//check that we still have a cert version in common with this connection. If we do not, disconnect.
```
nbrownus: We need to remove the check on line 492 - 494
```
	if err == nil {
		return false
	}
```

Otherwise we can only enter this section if the remoteCert is invalid and nebula is not configured to disconnect invalid certificates

connection_manager.go
```diff
@@ -488,18 +491,38 @@ func (cm *connectionManager) isInvalidCertificate(now time.Time, hostinfo *HostI
...
+	remoteVersion := remoteCert.Certificate.Version()
+	cs := cm.intf.pki.getCertState()
```
nbrownus: I think what we want to do here is look at `hostinfo.ConnectionState.myCert` first to determine how we initially formed this tunnel since we can technically complete a handshake with a mismatched certificate version. If the version we used is no longer in memory then we could tear it down but I think this might be more of a candidate for `cm.tryRehandshake` than anything

e2e/helpers_test.go
```diff
@@ -129,6 +129,109 @@ func newSimpleServer(v cert.Version, caCrt cert.Certificate, caKey []byte, name
 	return control, vpnNetworks, udpAddr, c
 }
 
+// newSimpleServer creates a nebula instance with fewer assumptions
```
nbrownus: nit: `newServer`

connection_manager.go
```diff
@@ -530,8 +553,32 @@ func (cm *connectionManager) sendPunch(hostinfo *HostInfo) {
...
+				WithField("peerVersion", peerCrt.Certificate.Version()).
+				WithField("reason", "local certificate version lower than peer, attempting to correct").
+				Info("Re-handshaking with remote")
+			cm.intf.handshakeManager.StartHandshake(hostinfo.vpnAddrs[0], func(hh *HandshakeHostInfo) {
+				hh.initiatingVersionOverride = peerCrt.Certificate.Version()
+			})
+			return
+		}
+	}
+	if curCrtVersion >= cs.initiatingVersion && bytes.Equal(curCrt.Signature(), myCrt.Signature()) == true {
```
nbrownus: This feels out of place now with the other tests being things to opt into re-handshaking and this being an opt out. Seems it would be less of a foot gun to invert the if and initiate the re-handshake from within

handshake_ix.go
```diff
@@ -167,12 +170,11 @@ func ixHandshakeStage1(f *Interface, addr netip.AddrPort, via *ViaSender, packet
...
+				Info("Might be unable to handshake with host due to missing certificate version")
```
nbrownus: This seems like it could get quite spammy and its not directly obvious if there is any action the operator can take to fix it or if they even need to. Think we can take this to debug level and let the initiator deal with logging the error on stage 2

handshake_ix.go
nbrownus: Yeah I think it's "initiator only understands v1 and responder only has v2". 

The other side of this is "initiator only has v1 but understands v2 and is trying to handshake with an ipv6 overlay address" which isn't allowed at the moment through https://github.com/slackhq/nebula/pull/1468/files#diff-d27b1b01b189fd93eae405de59c38e5bd23ff1c9868a0dabac92703842666b78R31

handshake_manager.go
```diff
@@ -74,7 +74,8 @@ type HandshakeHostInfo struct {
 	lastRemotes []netip.AddrPort // Remotes that we sent to during the previous attempt
 	packetStore []*cachedPacket  // A set of packets to be transmitted once the handshake completes
 
-	hostinfo *HostInfo
+	hostinfo                  *HostInfo
+	initiatingVersionOverride cert.Version
```
nbrownus: It's unfortunate to have this here and I'm sure we could wash it out with a little bit of pain refactoring stage 0. Maybe not a now thing but the additional memory for transitional feature makes me feel bad.

pki.go
```diff
@@ -100,55 +100,62 @@ func (p *PKI) reloadCerts(c *config.C, initial bool) *util.ContextualError {
 		currentState := p.cs.Load()
 		if newState.v1Cert != nil {
 			if currentState.v1Cert == nil {
-				return util.NewContextualError("v1 certificate was added, restart required", nil, err)
+				//adding certs is fine, actually. Networks-in-common confirmed in newCertState().
+			} else {
+				// did IP in cert change? if so, don't set
+				if !slices.Equal(currentState.v1Cert.Networks(), newState.v1Cert.Networks()) {
+					return util.NewContextualError(
```
nbrownus, line 107: Realizing now it would be incredibly helpful to indicate the certificate version we are failing on in these errors

handshake_ix.go
JackDoan: I did confirm the packet gets dropped with an ugly log line, no bad behavior occurs otherwise

handshake_manager.go
JackDoan: what if I told you that since we already have a bool in the struct (the rest are pointers and int64), adding a u8 will probably land in there as a u32, and make this structure 8-byte-aligned instead of 4-byte-aligned :D

or, if it was already self-aligning to 8-wide, this is free?

handshake_ix.go
JackDoan: good call, I had it like this for testing but this will def be annoying in The Real World

connection_manager.go
JackDoan: Yeah I agree, I think my other bug at 493 was avoiding a doom-loop of tunnel teardowns because of this section. Because we allow mismatch-handshakes now, I want to remove this logic entirely.

connection_manager.go
JackDoan: same problem as https://github.com/slackhq/nebula/pull/1468#discussion_r2399264175

connection_manager.go
```diff
@@ -475,31 +478,34 @@ func (cm *connectionManager) swapPrimary(current, primary *HostInfo) {
...
+		hostinfo.logger(cm.l).WithError(err).
+			WithField("fingerprint", remoteCert.Fingerprint).
+			Info("Remote certificate is blocked, tearing down the tunnel")
+		return true
+	} else if cm.intf.disconnectInvalid.Load() {
+		hostinfo.logger(cm.l).WithError(err).
+			WithField("fingerprint", remoteCert.Fingerprint).
+			Info("Remote certificate is no longer valid, tearing down the tunnel")
+		return true
+	} else {
```
nbrownus, line 505: nit, I'd leave out the else branch and `return false` at the end of the function to try and avoid future shenanigans

handshake_manager.go
nbrownus: On my machine, `master` is 104 bytes, in this branch it's currently `112` but if we move `initiatingVersionOverride cert.Version` next to `ready bool` then we can "get this one for free"

JackDoan: wow I didn't know structs were ordered in go. Will move!

connection_manager.go
JackDoan: I kinda see this the other way. If all the return statements are inside a conditional, anything outside the conditions is unreachable. It avoids future shenanigans in the form of "someone adds another condition that doesn't have a `return`".

JackDoan: Strong opinion loosely held though, I can swap it if you wanna

handshake_ix.go
```diff
@@ -23,13 +23,17 @@ func ixHandshakeStage0(f *Interface, hh *HandshakeHostInfo) bool {
...
 	cs := f.pki.getCertState()
 	v := cs.initiatingVersion
-	for _, a := range hh.hostinfo.vpnAddrs {
-		if a.Is6() {
-			v = cert.Version2
-			break
+	if hh.initiatingVersionOverride != cert.VersionPre1 {
+		v = hh.initiatingVersionOverride
+	} else if v < cert.Version2 {
+		// If we're connecting to a v6 address we must use a v2 cert
```
nbrownus: I think the `must` here is really a `should try super hard but maybe its ok`

connection_manager.go
```diff
@@ -475,31 +478,34 @@ func (cm *connectionManager) swapPrimary(current, primary *HostInfo) {
 	cm.hostMap.Unlock()
 }
 
-// isInvalidCertificate will check if we should destroy a tunnel if pki.disconnect_invalid is true and
```
JackDoan, line 478: before merge, check that we handle multi-IP certv1 in a sane-ish way

e2e/tunnels_test.go
```diff
@@ -55,3 +59,262 @@ func TestDropInactiveTunnels(t *testing.T) {
...
+			version := c.Cert.Version()
+			theirVersion := c2.Cert.Version()
+			r.Logf("version %d,%d", version, theirVersion)
+			if version == theirVersion {
+				break
+			}
+		}
+		since := time.Since(waitStart)
+		if since > time.Second*5 {
+			r.Log("wtf")
```
wadey, line 307: Maybe some better log here or remove it?

handshake_ix.go
wadey: agreed, the "must" isn't enforced since it looks like we just log a debug and continue.

nbrownus: I know it's just a comment but in the spirit of #1512, it would be nice to reflect that intent here ("try to use a v2 cert for this" type of thing)

JackDoan: "If we're connecting to a v6 address we should encourage use of a V2 cert" :D

e2e/tunnels_test.go
JackDoan: ooh thanks nice catch

PR #1472, opened 2025-09-08
don't require a detailsVpnAddr in a HostUpdateNotification

JackDoan:
If we don't use this information, don't put it on the wire, so we then do not have to check it. If this does indeed have a use, we can just remove the comment and close this

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -937,7 +937,6 @@ func (lh *LightHouse) SendUpdate() {
 						V4AddrPorts:   v4,
 						V6AddrPorts:   v6,
 						RelayVpnAddrs: relays,
-						VpnAddr:       netAddrToProtoAddr(lh.myVpnNetworks[0].Addr()),
 					},
 				}
 
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -1060,7 +1060,15 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, fromVpnAddrs []neti
 	queryVpnAddr, useVersion, err := n.Details.GetVpnAddrAndVersion()
 	if err != nil {
 		if lhh.l.Level >= logrus.DebugLevel {
-			lhh.l.WithField("from", fromVpnAddrs).WithField("details", n.Details).Debugln("Dropping malformed HostQuery")
+			lhh.l.WithField("from", fromVpnAddrs).WithField("details", n.Details).
+				Debugln("Dropping malformed HostQuery")
+		}
+		return
+	} else if useVersion == cert.Version1 && queryVpnAddr.Is6() {
+		// this case really shouldn't be possible to represent, but reject it anyway.
+		if lhh.l.Level >= logrus.DebugLevel {
+			lhh.l.WithField("vpnAddrs", fromVpnAddrs).WithField("queryVpnAddr", queryVpnAddr).
+				Debugln("invalid vpn addr for v1 handleHostQuery")
 		}
 		return
 	}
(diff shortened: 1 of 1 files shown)


nbrownus:
Places where we set `Details.VpnAddr`
- [x] `handleHostQuery` where we prep a lighthouse response to a non lighthouse host.
- [x] `handleHostUpdateNotification` this currently results in a `NebulaMeta_HostUpdateNotificationAck` back to the host that sent the update which is a noop.
- [x] `innerQueryServer` where we send `NebulaMeta_HostQuery` to lighthouses.
- [x] `sendHostPunchNotification` where we send `NebulaMeta_HostPunchNotification` to responder hosts after an initiator sent a `NebulaMeta_HostQuery` about them to a lighthouse. Line 1351 could be an interesting follow on to this PR.
- [x] `SendUpdate` where we send `NebulaMeta_HostUpdateNotification` to the lighthouses with our current known underlay addresses in it.
- [x] `resetMeta` which just zeroes the shared memory between lighthouse requests

Looks like all is well!

Side note, the `unlockedSetV4` and `unlockedSetV6` need to be updated to deal with the known `vpnAddrs` instead of just the current one being passed around to completely cover the allow lists cases.

Site side note, that cat is out of the bag and I only bring it up because of #1464, since it looks like the primary addr choice isn't entirely sufficient, we could change the proto definition for `NebulaMetaDetails.VpnAddr` to be a repeated field and have the lighthouse always inform the remote of all known vpn addrs for a host at that point in time. No doubt it will create other problems but it might also short circuit some of the issues around choosing a primary vpn addr.

JackDoan:
> Side note, the `unlockedSetV4` and `unlockedSetV6` need to be updated to deal with the known `vpnAddrs` instead of just the current one being passed around to completely cover the allow lists cases.

> Site side note, that cat is out of the bag and I only bring it up because of #1464, since it looks like the primary addr choice isn't entirely sufficient, we could change the proto definition for `NebulaMetaDetails.VpnAddr` to be a repeated field and have the lighthouse always inform the remote of all known vpn addrs for a host at that point in time. No doubt it will create other problems but it might also short circuit some of the issues around choosing a primary vpn addr.

I kind of like this idea -- we could also send a whole hostQueryReply per IP on a query, but that sounds much worse I think?

nbrownus:
Yeah sending individual replies per vpnAddr will confuse the situation more than anything I think, the client host has no way to correlate the answers into a single RemoteList.

lighthouse.go
```diff
@@ -1060,7 +1059,15 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, fromVpnAddrs []neti
 	queryVpnAddr, useVersion, err := n.Details.GetVpnAddrAndVersion()
 	if err != nil {
 		if lhh.l.Level >= logrus.DebugLevel {
-			lhh.l.WithField("from", fromVpnAddrs).WithField("details", n.Details).Debugln("Dropping malformed HostQuery")
+			lhh.l.WithField("from", fromVpnAddrs).WithField("details", n.Details).
+				Debugln("Dropping malformed HostQuery")
+		}
+		return
+	} else if useVersion == cert.Version1 && queryVpnAddr.Is6() {
```
wadey:
```suggestion
		return
	}
	if useVersion == cert.Version1 && queryVpnAddr.Is6() {
```

We don't need the `else` since the if block has a `return`, right?

wadey (approved):
approved, although anyone already using V2 certs will need to be careful to deploy this change to their lighthouses before updating their clients.

PR #1476, opened 2025-09-09
prevent linux from assigning ipv6 link-local addresses

JackDoan:
works on my machine! :D


nbrownus (approved):
Could see this logging as a warning instead, not really anything actionable when it fires

nbrownus (approved):

PR #1477, opened 2025-09-09
use netlink events to emit lighthouse updates instead of a timer

JackDoan:
This needs to be 

- [ ] confined to Linux (afaik Android will kill you if you try to even touch netlink)
- [ ] made configurable
- [ ] less log spammy
- [ ] do we still need to poke the lh every so often to make punchy work?
- [ ] debounced

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -21,6 +21,7 @@ import (
 	"github.com/slackhq/nebula/header"
 	"github.com/slackhq/nebula/udp"
 	"github.com/slackhq/nebula/util"
+	"github.com/vishvananda/netlink"
 )
 
 var ErrHostNotKnown = errors.New("host not known")
@@ -229,7 +230,7 @@ func (lh *LightHouse) reload(c *config.C, initial bool) error {
 				lh.updateCancel()
 			}
 
-			lh.StartUpdateWorker()
+			lh.StartUpdateWorkerNetlink()
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


nbrownus:
The main reason to stay in communication with a lighthouse is to learn about roaming real underlay addresses which might be opaque to the client.

Real underlay address changes are learned by simply receiving a properly crafted/encrypted packet though, so as long as the client sends encrypted packets on some regular cadence, we can squash the host update traffic.

Generally I would rely on test packets for this behavior but we made a change recently to reduce the amount of paperwork a nebula mesh has to deal with. If we are receiving traffic from the lighthouse then the tunnel is declared alive and we will never initiate a test packet to it. The only time we would send a test packet to a lighthouse is if we haven't sent or received a message from them within a 5 second period (by default) so we likely would want to adjust this to send test packets regardless of rx traffic.

JackDoan:
> The main reason to stay in communication with a lighthouse is to learn about roaming real underlay addresses which might be opaque to the client.

oof I forgot about NAT :cry: 

> Real underlay address changes are learned by simply receiving a properly crafted/encrypted packet though, so as long as the client sends encrypted packets on some regular cadence, we can squash the host update traffic.
> 
> Generally I would rely on test packets for this behavior but we made a change recently to reduce the amount of paperwork a nebula mesh has to deal with. If we are receiving traffic from the lighthouse then the tunnel is declared alive and we will never initiate a test packet to it. The only time we would send a test packet to a lighthouse is if we haven't sent or received a message from them within a 5 second period (by default) so we likely would want to adjust this to send test packets regardless of rx traffic.

yeah I like this approach

PR #1481, opened 2025-09-25
Pem stuff

JackDoan:


nbrownus (approved):

cert/cert.go
```diff
@@ -58,6 +58,9 @@ type Certificate interface {
 	// PublicKey is the raw bytes to be used in asymmetric cryptographic operations.
 	PublicKey() []byte
 
+	// PublicKeyPem is the value of PublicKey marshalled to PEM
+	PublicKeyPem() []byte
```
nbrownus: My only nit would be this, `MarshalPublicKeyPEM`?

JackDoan: I considered it but I think I prefer the symmetry with PublicKey()?

but hmmmmm consistency with elsewhere

cert/cert.go
```diff
@@ -58,6 +58,9 @@ type Certificate interface {
 	// PublicKey is the raw bytes to be used in asymmetric cryptographic operations.
 	PublicKey() []byte
 
+	// PublicKeyPem is the value of PublicKey marshalled to PEM
```
nbrownus: nit `// MarshalPublicKeyPEM`

JackDoan: oof

nbrownus (approved):

PR #1483, opened 2025-09-25
update to go 1.25, use the cool new ECDSA key marshalling functions

JackDoan:


jasikpark (commented):
lgtm

wadey:
We need to update `go.mod` as well

PR #1487, opened 2025-10-02
linux: opt out of naming your tun device yourself

JackDoan:
This allows you to end your `tun.dev` name setting with `%d` to tell Nebula "please find the lowest-numbered interface that matches this pattern and use it"

I'm honestly not sure what would be needed to bring this to other platforms. I think Mac OS mostly doesn't care. Windows seems like it will need different handling as well, but won't break with this change.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -242,6 +242,10 @@ tun:
   # When tun is disabled, a lighthouse can be started without a local tun interface (and therefore without root)
   disabled: false
   # Name of the device. If not set, a default will be chosen by the OS.
+  # For Linux: a single `%d` anywhere in the name is treated as a template and replaced with the
+  #   lowest number that yields an unused device name (e.g. `nebula%d` becomes `nebula0`, then `nebula1`, and so on, `neb%dprod` becomes `neb0prod`).
+  #   Only on Linux: `nebula%d` is the default if tun.dev is unset.
+  #   The resulting name must be shorter than the kernel limit of 16 characters.
   # For macOS: if set, must be in the form `utun[0-9]+`.
   # For NetBSD: Required to be set, must be in the form `tun[0-9]+`
   dev: nebula1
--- a/overlay/tun_linux.go
+++ b/overlay/tun_linux.go
@@ -5,6 +5,7 @@ package overlay
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"log/slog"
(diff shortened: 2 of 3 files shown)


jasikpark (commented):
looks reasonable to me

overlay/tun_linux.go
```diff
@@ -123,6 +130,38 @@ func newTun(c *config.C, l *logrus.Logger, vpnNetworks []netip.Prefix, multiqueu
...
+		good := true
+		for _, link := range links {
+			if candidateName == link.Attrs().Name {
+				good = false
+				break
+			}
+		}
+		if good {
+			if len(candidateName) > 16 {
+				return "", errors.New("you have too many nebula networks")
```
johnmaguire: I think it would be nice to be a little bit more clear about exactly what's going on here in order to make this error actionable. It's actually OK to have this many nebula networks, but we can't fit the tun name.

nbrownus: Could also be a pre-flight check and should be included the example config comments/docs. Can't compute a name shorter than `unix.IFNAMSIZ` if the `tunNameTemplate` is longer than it to begin with.

examples/config.yml
```diff
@@ -242,6 +242,10 @@ tun:
   # When tun is disabled, a lighthouse can be started without a local tun interface (and therefore without root)
   disabled: false
   # Name of the device. If not set, a default will be chosen by the OS.
+  # For Linux: a trailing `%d` is treated as a template and replaced with the
+  #   lowest number that yields an unused device name (e.g. `nebula%d` becomes `nebula0`, then `nebula1`, and so on).
+  #   Only on Linux: `nebula%d` is the default if tun.dev is unset.
```
johnmaguire, line 247: nit: you're already in `For Linux:`

overlay/tun_linux.go
```diff
@@ -298,6 +305,64 @@ func newTun(c *config.C, l *slog.Logger, vpnNetworks []netip.Prefix, multiqueue
...
+	return nextTunName(tunName, used)
+}
+
+// nextTunName substitutes the lowest unused integer into a "%d" template,
+// skipping any name present in used. tunName is assumed to have already passed
+// validateTunName. It errors only if every candidate that is shorter than
+// IFNAMSIZ is already taken.
+func nextTunName(tunName string, used map[string]struct{}) (string, error) {
+	tunNameTemplate := tunName[:len(tunName)-len("%d")]
+	for i := 0; ; i++ {
```
johnmaguire, line 359: this loop looks scary with a device name like `a%d` but the number of links registered should not be anywhere close to the theoretical limit (10^14)

Image

overlay/tun_linux.go
```diff
@@ -298,6 +305,64 @@ func newTun(c *config.C, l *slog.Logger, vpnNetworks []netip.Prefix, multiqueue
...
+		used[link.Attrs().Name] = struct{}{}
+	}
+	return nextTunName(tunName, used)
+}
+
+// nextTunName substitutes the lowest unused integer into a "%d" template,
+// skipping any name present in used. tunName is assumed to have already passed
+// validateTunName. It errors only if every candidate that is shorter than
+// IFNAMSIZ is already taken.
+func nextTunName(tunName string, used map[string]struct{}) (string, error) {
```
johnmaguire, line 357: This scan method is susceptible to TOCTOU issues, but I don't expect them to occur in practice.

johnmaguire (dismissed):
LGTM

PR #1488, opened 2025-10-03
Use x/net/route to manage routes directly

nbrownus:
This is a continuation of #1399 to finish off the last of the `exec.Command`s in freebsd which is needed for ipv6 unsafe route support

### DIFF
--- a/overlay/tun.go
+++ b/overlay/tun.go
@@ -1,6 +1,7 @@
 package overlay
 
 import (
+	"net"
 	"net/netip"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


overlay/tun.go
```diff
@@ -70,3 +71,13 @@ func findRemovedRoutes(newRoutes, oldRoutes []Route) []Route {
 
 	return removed
 }
+
+func prefixToMask(prefix netip.Prefix) netip.Addr {
```
JackDoan, line 75: nit: this is the same as `getNetmask` in tun_freebsd

nbrownus: Nice catch, ty

JackDoan (dismissed):

JackDoan (approved):

Issue #1480, opened 2025-09-25
Feature request: relay traffic to unsafe routes to another peer (for running Kubernetes over Nebula)


nbdd0121:
Sorry for the long text; this includes a background story on why the feature is requested; if the maintainers think there is a better way to address this, I am more than happy to listen and adjust my setup.

## Feature request

Have an opt-in feature, which I tentatively name "unsafe_route_peer_relay", that if enabled, allows:
* If both peer A and B have a certificate that allows a specific CIDR as unsafe route, say `10.100.0.0/15`, then A is able to send a packet with destination IP address within `10.100.0.0/15` to B regardless what is the source IP address.

This feature is intended for Nebula nodes that are routing nodes for unsafe networks and want to hand traffic to each other. It's an opt-in feature that only needs to be enabled on those routing nodes, and does not need to apply to any other nodes.

## Background

I am looking into use Nebula to provide the overlay for Kubernetes and replace VXLAN or Geneve. I've had some mixed success, so I am sharing the story and a feature that would make it work smoothly:

This is what I want to achieve:
1. Run a Kubernetes cluster, not necessarily on a same LAN, and have them communicate over Nebula.
2. Pods be able to reach Nebula peers (so it can connects to private services not on the cluster), and vice versa.
3. Able to expose services from the cluster to Nebula peers.

This is what I have done:
* For each cluster node, it was given a Nebula certificate that allows it to route to pod and service CIDR range (say, 10.100.0.0/16 for pods, and 10.101.0.0/16 for services). Cluster nodes have IP allocated within  `10.0.0.0/16`.
* I've selected Cilium as it's popular choice, but I think what I describe here would apply to Calico too.
* Cilium is configured to have Nebula IP as node IPs, native routing mode is used (instead of tunneling mode).
*  `use_system_route_table` is enabled on Nebula and `auto-direct-node-routes` is enabled on Cilium.
* Cilium is configured to expose service IPs (`lbExternalClusterIP`), as we do not have a way to automatically deploy load balancers with Nebula.
* Other Nebula clients have an unsafe route that point the `10.100.0.0/15` CIDR to one of the cluster node.

This is what happened:
* With native routing mode and `auto-direct-node-routes`, Cilium automatically inserts routes such as  `10.100.5.0/16 via 10.0.0.5 dev nebula` 
into the routing table. Nebula is able to pick it up using `use_system_route_table`.
[...]

PR #1493, opened 2025-10-07
Fix incorrect CIDR construction in hostmap

nbdd0121:
I tried to submit this through the channel specified in SECURITY.md however it was closed due to "not of critical severity" and was asked to fill a PR instead.

## Report

An improper CIDR construction in hostmap.go causes inbound firewall to allow a Nebula host where the certificate allows subnets (unsafe routes) or multiple IPs to send using any arbitrary IPs within the Nebula network rather than just its allocated address.
The subnet/multiple IP is needed as otherwise it triggers the fast path in the firewall, which is not vulnerable.

Reproduction:
*  Create a Nebula network, where one node would be rogue node.
*  This rogue node needs a certificate with multiple IP addresses, or a subnet routed through it.
*  On the rogue node, compile a version of Nebula with the outbound firewall disabled completely (outbound firewall construct CIDR correctly).
* The rogue node is now capable of sending packets with source address being any IP in the Nebula network, and it would not be rejected by the peer's firewall as from invalid remote IP.

The bug was introduced in https://github.com/slackhq/nebula/commit/e264a0ff888c7bf0568579306755a60fc42f6ecc where previously only the IP specified in the certificate is added, but the changed code adds the entire network. The correct code should be using addr's `BitLen` instead of `Mask.Size()`.

Nebula 1.9.4+ and current master branch are all affected.

## Impact

If an attacker has control, or has compromised a node with certificate that matches the requirement, it might be able to use it to send traffic impersonating another node (IP spoofing).
This enables attacks that IP spoofing allows. For example, it could use this to send arbitrary UDP packets to services or disrupt traffic by sending TCP RST packets.

## Backport

If the maintainer would wish to backport this to 1.9.x, here's the diff:

```diff
commit 570d173fd5d78b21803eb661af0a121fb54e9b3c
Author: Gary Guo <gary@garyguo.net>
Date:   Tue Oct 7 15:09:48 2025 +0100

    Fix incorrect CIDR construction in hostmap

diff --git a/hostmap.go b/hostmap.go
index 0b34de4..ea35f9c 100644
--- a/hostmap.go
+++ b/hostmap.go
@@ -723,7 +723,7 @@ func (i *HostInfo) CreateRemoteCIDR(c *cert.NebulaCertificate) {
 		//TODO: IPV6-WORK what to do when ip is invalid?
 		nip, _ := netip.AddrFromSlice(ip.IP)
 		nip = nip.Unmap()
-		bits, _ := ip.Mask.Size()
+		bits := nip.BitLen()
 		remoteCidr.Insert(netip.PrefixFrom(nip, bits), struct{}{})
 	}
```

### DIFF
--- a/firewall_test.go
+++ b/firewall_test.go
@@ -692,6 +692,50 @@ func TestFirewall_DropConntrackReload(t *testing.T) {
 	assert.Equal(t, fw.Drop(p, false, &h, cp, nil), ErrNoMatchingRule)
 }
 
+func TestFirewall_DropIPSpoofing(t *testing.T) {
+	l := test.NewLogger()
+	ob := &bytes.Buffer{}
+	l.SetOutput(ob)
+
+	c := cert.CachedCertificate{
+		Certificate: &dummyCert{
+			name:     "host-owner",
+			networks: []netip.Prefix{netip.MustParsePrefix("192.0.2.1/24")},
+		},
+	}
+
+	c1 := cert.CachedCertificate{
+		Certificate: &dummyCert{
+			name:           "host",
+			networks:       []netip.Prefix{netip.MustParsePrefix("192.0.2.2/24")},
+			unsafeNetworks: []netip.Prefix{netip.MustParsePrefix("198.51.100.0/24")},
+		},
+	}
+	h1 := HostInfo{
+		ConnectionState: &ConnectionState{
+			peerCert: &c1,
+		},
+		vpnAddrs: []netip.Addr{c1.Certificate.Networks()[0].Addr()},
+	}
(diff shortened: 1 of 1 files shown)


nbrownus:
Thanks for the report!

I do agree this is a problem and have created #1494 to fix it within the v1.9.x release family. 

Do you mind signing the CLA so that we can get this merged for `master`?

nbdd0121:
I found this by accident when working on #1480. When adding test for the behaviour with the feature gate enabled, I thought I might as well add test to show that it's currently permitted. To my surprise of the the firewall check that I expected to fail without feature enabled passed unexpectedly, and then the issue is found by root-causing that.

PR #1495, opened 2025-10-08
Support for multi proto tun device on OpenBSD

nbrownus:
Similar to #1492 but this time for OpenBSD

Tested on
```
# uname -a
OpenBSD openbsd.localdomain 7.7 GENERIC#619 amd64
```


JackDoan (dismissed):
fun!

JackDoan (approved):

PR #1503, opened 2025-10-16
Allow defining a route table id where routes will be installed

dioss-Machiel:
This feature allows defining the route table id where Nebula will install routes to unlock the full potential of policy routing.

Additionally this also makes it so that the 'default routes' are always installed with the correct MTU, previously there was a tiny window where the MTU was not set.

This uses the IFA_F_NOPREFIXROUTE flag when setting the IP of the interface to avoid adding a route by default.
This flag was added in Linux kernel 4.4, if support for older kernel versions is required then more changes are going to be required.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -280,6 +280,11 @@ tun:
   # SO_RCVBUFFORCE is used to avoid having to raise the system wide max
   #use_system_route_table_buffer_size: 0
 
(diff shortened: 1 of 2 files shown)


nbrownus:
@dioss-Machiel Thanks for the PR! I like that this resolves some silly race conditions at startup. We will need to support pre v4.4 linux kernels though, sadly.

PR #1509, opened 2025-10-21
Firewall types and cross-stack subnet stuff

JackDoan:
firewall can distinguish if the host connecting has an overlapping network, is a VPN peer without an overlapping network, or is a unsafe network

has elements of #1498, but not the behavioral changes

- [x] needs many more tests!

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -433,7 +433,7 @@ func (lh *LightHouse) loadStaticMap(c *config.C, staticList map[netip.Addr]struc
 
 		if !lh.myVpnNetworksTable.Contains(vpnAddr) {
 			lh.l.WithFields(m{"vpnAddr": vpnAddr, "networks": lh.myVpnNetworks, "entry": i + 1}).
-				Warn("static_host_map key is not in our network, this might be a mistake")
+				Warn("static_host_map key is not within our networks, layer 3 network traffic to this host will not work")
 		}
 
 		vals, ok := v.([]any)


lighthouse.go
```diff
@@ -360,7 +360,8 @@ func (lh *LightHouse) parseLighthouses(c *config.C) ([]netip.Addr, error) {
 		}
 
 		if !lh.myVpnNetworksTable.Contains(addr) {
-			return nil, util.NewContextualError("lighthouse host is not in our networks, invalid", m{"vpnAddr": addr, "networks": lh.myVpnNetworks}, nil)
+			lh.l.WithFields(m{"vpnAddr": addr, "networks": lh.myVpnNetworks}).
+				Warn("lighthouse host is not in our networks, this might be a mistake")
```
nbrownus: I still think we should be more descriptive of the possible mistake (and the one below) since this will appear to be a working lighthouse but the moment someone tries to ping it or do anything else it's gonna appear to be a strangely broken lighthouse.

hostmap.go
```diff
@@ -730,20 +742,26 @@ func (i *HostInfo) SetRemoteIfPreferred(hm *HostMap, newRemote netip.AddrPort) b
...
 
-func (i *HostInfo) buildNetworks(networks, unsafeNetworks []netip.Prefix) {
-	if len(networks) == 1 && len(unsafeNetworks) == 0 {
-		// Simple case, no CIDRTree needed
-		return
+// buildNetworks fills in the networks field of HostInfo. It accepts a cert.Certificate so you never ever mix the network types up.
+func (i *HostInfo) buildNetworks(myVpnNetworksTable *bart.Lite, c cert.Certificate) {
+	if len(c.Networks()) == 1 && len(c.UnsafeNetworks()) == 0 {
+		if myVpnNetworksTable.Contains(c.Networks()[0].Addr()) {
+			return // Simple case, no CIDRTree needed
```
nbrownus: nit: could update the comment to refer to bart instead of the defunct CIDRTree

firewall.go
```diff
@@ -417,6 +417,8 @@ func AddFirewallRulesFromConfig(l *logrus.Logger, inbound bool, c *config.C, fw
 	return nil
 }
 
+var ErrUnknownNetworkType = errors.New("unknown network type")
+var ErrPeerRejected = errors.New("remote IP is not within a subnet that we handle")
```
nbrownus: As part of the earlier cert v2 stuff, we began changing from `ip` and `subnet` to `addr` and `network`. Would be nice to change these error names and text to match. Of course there are still other places that need updating.

lighthouse.go
JackDoan: do you have phrasing to suggest?

firewall.go
JackDoan: well, I got these :D

lighthouse.go
nbrownus:
```
lighthouse host is not within our networks, lighthouse functionality will work but layer 3 network traffic will not
```
?

lighthouse.go
```diff
@@ -431,7 +432,8 @@ func (lh *LightHouse) loadStaticMap(c *config.C, staticList map[netip.Addr]struc
 		}
 
 		if !lh.myVpnNetworksTable.Contains(vpnAddr) {
-			return util.NewContextualError("static_host_map key is not in our network, invalid", m{"vpnAddr": vpnAddr, "networks": lh.myVpnNetworks, "entry": i + 1}, nil)
+			lh.l.WithFields(m{"vpnAddr": vpnAddr, "networks": lh.myVpnNetworks, "entry": i + 1}).
+				Warn("static_host_map key is not in our network, this might be a mistake")
```
nbrownus: This one is slightly different
```
static_host_map key is not within our networks, layer 3 network traffic to this host will not work
```
?

nbrownus (approved):

PR #1512, opened 2025-10-21
Cross stack subnet stuff

JackDoan:
replaces #1506

- [x] We want to allow handshakes (and other L2 traffic like relays+lighthouseing) if your certs are valid and trusted (IP network overlap not required)
    - [x] handshakes allowed
    - [x] checks removed from other L2 traffic areas, esp config file
    - [x] tests!
- [x] We do not want to pass L3 traffic without a vpn-network in common between the two hosts
- [x] We will store and remember all IPs in a certificate in the hostinfo
- [ ] When we log vpnaddr(s), print as many IPs as we know

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -232,7 +232,9 @@ type HostInfo struct {
 	remoteIndexId   uint32
 	localIndexId    uint32
 
-	// vpnAddrs is a list of vpn addresses assigned to this host
+	// vpnAddrs is a list of vpn addresses assigned to this host that are within our own vpn networks
+	// The host may have other vpn addresses that are outside our
+	// vpn networks but were removed because they are not usable
 	vpnAddrs []netip.Addr
 
 	// networks is a combination of specific vpn addresses (not prefixes!) and full unsafe networks assigned to this host.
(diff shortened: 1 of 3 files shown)


hostmap.go
```diff
@@ -730,20 +740,27 @@ func (i *HostInfo) SetRemoteIfPreferred(hm *HostMap, newRemote netip.AddrPort) b
 	return false
 }
 
-func (i *HostInfo) buildNetworks(networks, unsafeNetworks []netip.Prefix) {
+func (i *HostInfo) buildNetworks(myVpnNetworksTable *bart.Lite, networks, unsafeNetworks []netip.Prefix) {
```
nbrownus: I keep thinking this would be better as `buildNetworks(myVpnNetworksTable *bart.Lite, crt cert.Certificate)`

lighthouse.go
```diff
@@ -360,7 +360,8 @@ func (lh *LightHouse) parseLighthouses(c *config.C) ([]netip.Addr, error) {
 		}
 
 		if !lh.myVpnNetworksTable.Contains(addr) {
-			return nil, util.NewContextualError("lighthouse host is not in our networks, invalid", m{"vpnAddr": addr, "networks": lh.myVpnNetworks}, nil)
+			lh.l.WithFields(m{"vpnAddr": addr, "networks": lh.myVpnNetworks}).
+				Warn("lighthouse host is not in our networks, this might be a mistake")
```
nbrownus, line 364: I think it would be beneficial to explain a little more about the mistake, maybe `lighthouse host is not in our networks, only lighthouse control traffic will work, you will not be able to ping this lighthouse.` or something. Same with the static host map.

JackDoan: Ah yeah good idea

hostmap.go
JackDoan: Why? I don't like passing whole objects around when I only need a member or two

nbrownus: Less to type and screw up, they are the literal `Networks()` and `UnsafeNetworks()` from the certificate, anything else would be an error.

JackDoan: ohhhh ordering bc they're the same type. Smart.

There's a school of thought that would tell us to define a `VPNNetwork` and `UnsafeNetwork` type, but, I don't wanna.

Will change.

JackDoan: (this change actually is part of #1509, I forgot to target this PR at that one to reflect that, updated)

JackDoan: ah crap #1509 doesn't include the removal of filteredNetworks

PR #1518, opened 2025-10-31
extra stinky slop that works kinda. DO NOT EVER MERGE.

rawdigits:


JackDoan:
if no merge, is not PR

PR #1521, opened 2025-11-09
feat: support via gateway for v6 multihop for v4 routes

6ixfalls:
When using use_system_route_table, IPv4 routes with an IPv6 nexthop are not added to the internal routing map. As a result, connectivity does not work. This fixes that by adding the via addresses in the route as a gateway, since it appears that Gw is not present on these kinds of routes.

I'm not too well-versed with the Linux kernel, so I'm not sure what the distinction between Gw and Via is. I tried to preserve existing behavior as much as possible (use Gw first, before checking Via). There may need to be more testing done around potential routes that could exist.

### DIFF
--- a/overlay/tun_linux.go
+++ b/overlay/tun_linux.go
@@ -608,7 +608,7 @@ func (t *tun) getGatewaysFromRoute(r *netlink.Route) routing.Gateways {
 	for _, p := range r.MultiPath {
 		// If this route is relevant to our interface and there is a gateway then add it
 		if p.LinkIndex == link.Attrs().Index {
-			gwAddr, ok := getGatewayAddr(r.Gw, r.Via)
+			gwAddr, ok := getGatewayAddr(p.Gw, p.Via)
 			if ok {
 				if t.isGatewayInVpnNetworks(gwAddr) {
 					gateways = append(gateways, routing.NewGateway(gwAddr, p.Hops+1))


6ixfalls:
Page says signed, not sure why the status hasn't updated.

nbrownus:
According to my research `RTA_VIA` is  more or less synonymous with `RTA_GATEWAY` and according to the `rtenetlink` man page it's a "Gateway in different AF", so this makes sense.

I've never tried to run a configuration like this but I was able to reproduce your situation in 2 ways:
```
ip -4 route add 192.168.98.1/32 via inet6 fd99::101 dev nebula1
```

and

```
ip nexthop add id 10 via inet6 fd99::101 dev nebula1
ip route add 192.168.98.0/24 nhid 10
```

Not sure why I have to use a nexthop object to get a larger ipv4 route in place.

How are you configuring your network?

6ixfalls:
I'm hitting the first issue, mostly due to my use case of Nebula as a BGP backbone. The BGP routes are exported to the kernel with IPv6 nexthops, since the BGP peering is over IPv6 only and I have extended next hop enabled.

nbrownus:
@6ixfalls I pushed a small change to the branch and verified that it's working on my own network. Mind giving it a test on your side?

Also verified this works in config
```
tun:
  unsafe_routes:
    - route: 192.168.98.0/24
      via: fd99::102
```

6ixfalls:
Unfortunately, it appears that the new image is not working for me.

6ixfalls:
Seems to have completely borked the routing, since every route now errors with `Ignoring route update, invalid gateway or via address` and `no gateways` (even ones that worked prior to this change).

nbrownus:
@6ixfalls hmm, I will re-review, I also sent you an email to see if we can sync up on Slack or Discord to knock this out.

PR #1525, opened 2025-11-11
Don't log every blocklisted fingerprint

nbrownus:
Currently we log every blocklisted fingerprint on its own line. Deployments with many blocked fingerprints spams the logs with these every time they reload/restart.

To quiet things down I am suggesting we just log the count.

### DIFF
--- a/pki.go
+++ b/pki.go
@@ -523,9 +523,13 @@ func loadCAPoolFromConfig(l *logrus.Logger, c *config.C) (*cert.CAPool, error) {
 		return nil, fmt.Errorf("error while adding CA certificate to CA trust store: %s", err)
 	}
 
-	for _, fp := range c.GetStringSlice("pki.blocklist", []string{}) {
-		l.WithField("fingerprint", fp).Info("Blocklisting cert")
-		caPool.BlocklistFingerprint(fp)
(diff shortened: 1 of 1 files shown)


pki.go
```diff
@@ -523,9 +523,13 @@ func loadCAPoolFromConfig(l *logrus.Logger, c *config.C) (*cert.CAPool, error) {
...
-	for _, fp := range c.GetStringSlice("pki.blocklist", []string{}) {
-		l.WithField("fingerprint", fp).Info("Blocklisting cert")
-		caPool.BlocklistFingerprint(fp)
+	bl := c.GetStringSlice("pki.blocklist", []string{})
+	if len(bl) > 0 {
+		for _, fp := range bl {
+			caPool.BlocklistFingerprint(fp)
+		}
+
+		l.WithField("fingerprintCount", len(bl)).Info("Blocklisted certificates")
```
jasikpark, line 532: Would it be worth it to keep the existing logs behind a log_level = debug if check?

nbrownus: This mostly shows up in defined when we ask folks for logs which puts the logging into debug mode.

jasikpark: Ah, good point.

PR #1533, opened 2025-11-19
Stab at better logging when a relay is being used

### LINKED ISSUE #1187: Change log handling for `remote` when a relay is in use.
With #1173 we had some undesirable behavior with logging when using relays. We should find all log locations that are now logging `Invalid Addr` instead of `nil` for remote and change them to be less scary.

nbrownus:
Fixes #1187 

This changes the log output for udp addresses when a relay was used to send the packet.
Text mode: `from="10.0.0.128:4242 (relayed)"`
json mode: `"from":{"relay":"10.0.0.128:4242"}`

Before this change it was:
Text mode: `udpAddr="invalid AddrPort"`
json mode: `"udpAddr":""`

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -294,9 +294,9 @@ func (v ViaSender) String() string {
 
 func (v ViaSender) MarshalJSON() ([]byte, error) {
 	if v.IsRelayed {
-		return json.Marshal(m{"direct": v.UdpAddr})
+		return json.Marshal(m{"relay": v.UdpAddr})
 	}
-	return json.Marshal(m{"relay": v.UdpAddr})
+	return json.Marshal(m{"direct": v.UdpAddr})
 }
 
 type cachedPacket struct {


handshake_ix.go
```diff
@@ -99,11 +99,11 @@ func ixHandshakeStage0(f *Interface, hh *HandshakeHostInfo) bool {
 	return true
 }
 
-func ixHandshakeStage1(f *Interface, addr netip.AddrPort, via *ViaSender, packet []byte, h *header.H) {
```
JackDoan, line 102: this is also a nice step towards using different types to represent overlay vs underlay IPs, which I think could make nebula much easier to read :D

handshake_ix.go
```diff
@@ -329,37 +329,39 @@ func ixHandshakeStage1(f *Interface, addr netip.AddrPort, via *ViaSender, packet
...
-					f.l.WithField("vpnAddrs", existing.vpnAddrs).WithField("udpAddr", addr).
+					f.l.WithField("vpnAddrs", existing.vpnAddrs).WithField("from", via).
 						WithField("handshake", m{"stage": 2, "style": "ix_psk0"}).WithField("cached", true).
 						Info("Handshake message sent")
 				}
 				return
 			} else {
-				if via == nil {
-					f.l.Error("Handshake send failed: both addr and via are nil.")
+				if via.relay == nil {
```
JackDoan, line 363: is this different from `!via.IsRelayed`? I guess I should keep reading

hostmap.go
```diff
@@ -276,9 +278,25 @@ type HostInfo struct {
 }
 
 type ViaSender struct {
+	UdpAddr   netip.AddrPort
```
JackDoan, line 281: so when IsRelayed is true, is UdpAddr the address of the target or the relay?

hostmap.go
```diff
@@ -276,9 +278,25 @@ type HostInfo struct {
...
 	relayHI   *HostInfo // relayHI is the host info object of the relay
 	remoteIdx uint32    // remoteIdx is the index included in the header of the received packet
 	relay     *Relay    // relay contains the rest of the relay information, including the PeerIP of the host trying to communicate with us.
+	IsRelayed bool      // IsRelayed is true if the packet was sent through a relay
```
JackDoan, line 285: might be better to have this as a method so it can never fall out of sync with `relay`?

outside.go
```diff
@@ -19,21 +19,21 @@ const (
 	minFwPacketLen = 4
 )
 
-func (f *Interface) readOutsidePackets(ip netip.AddrPort, via *ViaSender, out []byte, packet []byte, h *header.H, fwPacket *firewall.Packet, lhf *LightHouseHandler, nb []byte, q int, localCache firewall.ConntrackCache) {
+func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte, h *header.H, fwPacket *firewall.Packet, lhf *LightHouseHandler, nb []byte, q int, localCache firewall.ConntrackCache) {
```
JackDoan: does dropping the pointer on `via` cause a copy?

outside.go
```diff
@@ -126,31 +134,34 @@ func (f *Interface) readOutsidePackets(ip netip.AddrPort, via *ViaSender, out []
...
 		if err != nil {
-			hostinfo.logger(f.l).WithError(err).WithField("udpAddr", ip).
+			hostinfo.logger(f.l).WithError(err).WithField("from", via).
 				WithField("packet", packet).
 				Error("Failed to decrypt lighthouse packet")
 			return
 		}
 
-		lhf.HandleRequest(ip, hostinfo.vpnAddrs, d, f)
+		//NOTE: via should never be a relayed from here
```
JackDoan: while it can't ever happen, the protocol doesn't actually forbid this. We've hand-waved about one day letting people set relay information statically, which could (I think) result in this happening?

handshake_ix.go
nbrownus: The original code had a double check to ensure there was actually a relay there. Seems like a fine thing to keep checking for.

hostmap.go
nbrownus: Relay, need to be careful to check `IsRelayed`

hostmap.go
nbrownus: Sure, at that point I would stick it in another package and not export the underlying variable as well.

outside.go
nbrownus: `*ViaSender` in `master` is not optimized but is only used for an actual relayed packet and we've been pretty up front that relays aren't fully performance tuned.

Given that, I wanted to avoid the performance implication of a guaranteed heap allocation/gc work for every packet but I am also not settled on this being the final implementation. Reusing a `*ViaSender` is probably pretty straight forward, just haven't chewed on it yet since I wasn't sure if this is the approach we wanted to take here.

outside.go
nbrownus: A relayed connection to a lighthouse is strange and certainly not doing you any favors with NAT traversal but it could be entirely valid for advanced use cases.

I was mainly trying to remind myself that anything in `readOutsidePackets` that didn't also accept a `via` likely has protections against an invalid `netip.AddrPort` but with the new code it needs to instead check `IsRelayed`

hostmap.go
```diff
@@ -276,9 +278,25 @@ type HostInfo struct {
...
+		return fmt.Sprintf("%s (relayed)", v.UdpAddr)
+	}
+	return v.UdpAddr.String()
+}
+
+func (v *ViaSender) MarshalJSON() ([]byte, error) {
+	if v.IsRelayed {
+		return json.Marshal(m{"direct": v.UdpAddr})
+	}
+	return json.Marshal(m{"relay": v.UdpAddr})
```
wadey: Is this logic backwards?

interface.go
```diff
@@ -279,7 +279,7 @@ func (f *Interface) listenOut(i int) {
 	nb := make([]byte, 12, 12)
 
 	li.ListenOut(func(fromUdpAddr netip.AddrPort, payload []byte) {
-		f.readOutsidePackets(fromUdpAddr, nil, plaintext[:0], payload, h, fwPacket, lhh, nb, i, ctCache.Get(f.l))
```
wadey, line 282: does this add an allocation in the hot path?

nbrownus: It doesn't escape, at least currently. I asked Jack to run a perf test on both `struct` vs` *struct` commits I have here. I expect `*struct` is maybe a tiny fraction faster, we would just need to be careful not to have it escape in future commits.

outside.go
nbrownus: Jack ran a some tests of `*ViaSender` vs `ViaSender` and found `ViaSender` is a bit better for performance

hostmap.go
nbrownus: Yep, fixed

interface.go
nbrownus: Jack ran a some tests of `*ViaSender` vs `ViaSender` and found `ViaSender` is a bit better for performance

handshake_ix.go
wadey: this line checks for `via.relay` but the follow lines use `via.relayHI`. Should we be checking for that instead?

brad-defined (approved):

PR #1535, opened 2025-11-19
improve nebula-cert sign version auto-select

JackDoan:


cmd/nebula-cert/sign.go
```diff
@@ -279,21 +283,20 @@ func signCert(args []string, out io.Writer, errOut io.Writer, pr PasswordReader)
...
 		if len(v4Networks) != 1 {
 			return newHelpErrorf("invalid -networks definition: v1 certificates can only have a single ipv4 address")
 		}
 
-		if version == cert.Version1 {
-			// If we are asked to mint a v1 certificate only then we cant just ignore any v6 addresses
-			if len(v6Networks) > 0 {
-				return newHelpErrorf("invalid -networks definition: v1 certificates can only be ipv4")
-			}
+		// If we are asked to mint a v1 certificate only then we cant just ignore any v6 addresses
```
nbrownus: Comment language is a bit strange since we can't mint v1+v2 in a single go

JackDoan: ah right I forgot I changed that

PR #1538, opened 2025-11-20
Make 0.0.0.0/0 and ::/0 not mean any address family, add any for that

nbrownus:
This scenario is only present on `master` since v1.9.x and prior would refuse ipv6 traffic.

Currently a cidr of `0.0.0.0/0` or `::/0` results in a short circuit `any` for both ipv4 or ipv6 addresses. This is unnecessary and could lead to surprises for network operators.

This change makes it so that `0.0.0.0/0` means only all ipv4 addresses and `::/0` means only all ipv6 addresses.

This also adds support for the special `cidr: any` and `localCidr: any` which actually means "any ip family and address". The main reason for this is in unsafe routes, when you desire to expose the entire `localCidr` networks. If we did not support `localCidr: any` then you would need to create 2 rules, one for ipv4 `0.0.0.0/0` and another for ip6 `::/0`.

### DIFF
--- a/firewall.go
+++ b/firewall.go
@@ -23,7 +23,7 @@ import (
 )
 
 type FirewallInterface interface {
-	AddRule(incoming bool, proto uint8, startPort int32, endPort int32, groups []string, host string, addr, localAddr string, caName string, caSha string) error
+	AddRule(incoming bool, proto uint8, startPort int32, endPort int32, groups []string, host string, cidr, localCidr string, caName string, caSha string) error
 }
 
 type conn struct {


firewall.go
```diff
@@ -248,30 +248,19 @@ func NewFirewallFromConfig(l *logrus.Logger, cs *CertState, c *config.C) (*Firew
...
-	sIp := ""
-	if ip.IsValid() {
-		sIp = ip.String()
-	}
-	lIp := ""
-	if localIp.IsValid() {
-		lIp = localIp.String()
-	}
-
+func (f *Firewall) AddRule(incoming bool, proto uint8, startPort int32, endPort int32, groups []string, host string, cidr, localCidr, caName string, caSha string) error {
```
johnmaguire, line 251: nit: cidr, localCidr here, but addr, localAddr in the interface.

firewall.go
```diff
@@ -786,30 +773,34 @@ func (fr *FirewallRule) addRule(f *Firewall, groups []string, host string, ip, l
...
 		}
-		err := nlc.addRule(f, localCIDR)
+		err := nlc.addRule(f, localCidr)
 		if err != nil {
 			return err
 		}
 		fr.Hosts[host] = nlc
 	}
 
-	if ip.IsValid() {
```
johnmaguire, line 796: For future reviewers - this just tells you if it's a zero-value or not. This is not true validation per se. We won't return new errors we didn't return before for an "invalid" IP.

johnmaguire (dismissed):
LGTM, and test coverage looks good.

johnmaguire (approved):

firewall.go
jasikpark: https://pkg.go.dev/net/netip#Addr.IsValid strange they didn't follow the pattern in https://pkg.go.dev/time#Time.IsZero which is much more clear.

Issue #1542, opened 2025-12-02
Changelog out of date


phlummox:
The CHANGELOG.md seems to be out of date.

The last version it has an entry for is 1.9.4, here:

https://github.com/slackhq/nebula/blob/64f202fa171db446573057f38555d43921566a24/CHANGELOG.md?plain=1#L17

So it's missing entries for versions 1.9.5, 1.9.6 and 1.9.7.

brad-defined:
The master branch has been heavily updated for some major changes, and in the meantime, maintenance has been occurring on the release-1.9 branch. Please refer to the release-1.9 branch changelog for releases 1.9.5-1.9.7.

https://github.com/slackhq/nebula/blob/release-1.9/CHANGELOG.md

PR #1534, opened 2025-11-19
Update changelog for v1.10.0

nbrownus:
v1.10 specific changes
- [x] 56067af (origin/master, origin/HEAD, master) Stab at better logging when a relay is being used (#1533)
- [x] 64f202f Make 0.0.0.0/0 and ::/0 not mean any address family, add any for that (#1538)
- [x] 6d7cf61 improve nebula-cert sign version auto-select (#1535)
- [x] 83ae807 No need to clear counter 0 (#1537)
- [x] 12cf348 (bits-counter-9) feat: support via gateway for v6 multihop for v4 routes (#1521)
- [x] 7aff313 Relax the restriction on routines from the config (#1531)
- [x] 297767b warn user if they configure a firewall rule that will allow way more traffic than you might expect (#1513)
- [x] 99faab5 Fix a potential bug with udp ipv4 only on darwin (#1532)
- [x] 27ea667 add more tests around bits counters (#1441)
- [x] 4df8bcb nebula-cert: support reading CA passphrase from env (#1421)
- [x] a89f951 Firewall types and cross-stack subnet stuff (#1509)
- [x] 52f1908 Don't log every blocklisted fingerprint (#1525)
- [x] 48f1ae9 switch to go.yaml.in/yaml (#1478)
- [x] 97b3972 honor remote_allow_list in hole punch response (#1186)
- [x] 0f305d5 don't block startup on failure to configure SSH (#1520)
- [x] 01909f4 try to make certificate addition/removal reloadable in some cases (#1468)
- [x] 45c1d3e Support for multi proto tun device on OpenBSD (#1495)
- [x] eb89839 Support for multi proto tun device on NetBSD (#1492)
- [x] fb7f0c3 Use x/net/route to manage routes directly (#1488)
- [x] b1f53d8 Support IPv6 tunneling in FreeBSD (#1399)
- [x] 8824eea helper functions to more correctly marshal curve 25519 public keys (#1481)
- [x] 1ea5f77 update to go 1.25, use the cool new ECDSA key marshalling functions (#1483)
- [x] 4cdeb28 Set CKA_VALUE_LEN attribute in DeriveNoise (#1482)
- [x] 5cccd39 update RemoteList.vpnAddrs when we complete a handshake (#1467)
- [x] 65cc253 prevent linux from assigning ipv6 link-local addresses (#1476)
- [x] 73cfa7b add firewall tests for ipv6 (#1451)
- [x] 768325c cert-v2 chores (#1466)
[...]

PR #1544, opened 2025-12-03
update CHANGELOG.md, point readers to release-1.9 branch

phlummox:
It seems development for the last few versions is happening on the https://github.com/slackhq/nebula/tree/release-1.9 branch - this amends the CHANGELOG.md in the master branch, so it's clearer to readers where the changes for those versions are documented (as per #1542).

(It's also possible to get the CHANGELOG entries by going from a particular release on the Releases page, to the corresponding repo tag, and then going to the CHANGELOG at that point in the repo - but it might be helpful to note this on the master branch.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


nbrownus:
This is handled by #1534 which we plan to merge very soon

PR #1552, opened 2025-12-09
cert.MarshalSigningPublicKeyToPEM should emit the 'ECDSA' variant of the banner

JackDoan:
This makes it possible to distinguish "this P256 key is for signing" and "this P256 key is for key agreement" when they are in their PEM form. This will make your FIPS auditor(s) happier.

### DIFF
--- a/cert/cert_v1_test.go
+++ b/cert/cert_v1_test.go
@@ -100,12 +100,18 @@ AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
 AAAAAAAAAAAAAAAAAAAAAAA=
 -----END NEBULA P256 PUBLIC KEY-----
 `)
+
+	pubP256KeyPemCA := []byte(`-----BEGIN NEBULA ECDSA P256 PUBLIC KEY-----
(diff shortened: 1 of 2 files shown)


nbrownus:
I agree that this is the technically correct thing to do for the banner but it does make some things suck about unmarshalling. To be clear I highly doubt there are many use cases, if any, of un/marshalling a ECDSA P256 public key with our banners but I would also push for a larger change to make this interface even cleaner.

Basically I would advocate for removing:
- `MarshalSigningPrivateKeyToPEM`
- `UnmarshalSigningPrivateKeyFromPEM`
- `MarshalSigningPublicKeyToPEM`
- (there is no `UnmarshalSigningPublicKeyFromPEM` to remove but if there were...)

And then change the return of the following functions to either indicate they are signing keys (`bool`) or the banner constant (both?):
- `UnmarshalPrivateKeyFromPEM`
- `UnmarshalPublicKeyFromPEM`

And then make these functions handle all types of keys:
- `MarshalPublicKeyToPEM`
- `MarshalPrivateKeyToPEM`

JackDoan:
Sounds like a great issue to file for 1.11 ;)

wadey (approved):

PR #1556, opened 2025-12-11
Refactor CA pool handling to use streaming

maggie44:
Refactors the CA pool handling to reduce memory usage when loading large config files on low memory devices:

```
Before (full-slice parsing):

100 CAs: 4.13 ms/op, 1.46 MB, 1,596 allocs
250 CAs: 9.90 ms/op, 8.75 MB, 3,576 allocs
500 CAs: 20.8 ms/op, 33.4 MB, 6,840 allocs
1000 CAs: 45.5 ms/op, 129 MB, 13,347 allocs
5000 CAs: 337 ms/op, 3.14 GB, 65,415 allocs

After (streaming parser):

100 CAs: 4.72 ms/op, 166 KB, 1,493 allocs
250 CAs: 9.61 ms/op, 330 KB, 3,300 allocs
500 CAs: 18.96 ms/op, 667 KB, 6,314 allocs
1000 CAs: 37.0 ms/op, 1.36 MB, 12,343 allocs
5000 CAs: 187.37 ms/op, 6.78 MB, 60,495 allocs
```

### DIFF
--- a/cert/ca_pool.go
+++ b/cert/ca_pool.go
@@ -1,11 +1,13 @@
 package cert
 
 import (
+	"bytes"
+	"encoding/pem"
 	"errors"
 	"fmt"
+	"io"
 	"net/netip"
 	"slices"
-	"strings"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


JackDoan:
Hi @maggie44!

I'm still reading through this, but I'm really curious about your use-case. How many CAs do you typically need to load (any why so many?)

One of the sort of unwritten assumptions throughout Nebula is that while there should always be more than one CA, there are probably are much fewer than 100. If you do have that many, it'd be good for us to know so we can consider it moving forward!

On the other hand, if you just cranked `N` up to show the difference in the benchmark, that would make sense too.

maggie44:
Hi @JackDoan,

For my use case, the cranking of N to levels like 5000 is mostly to show the difference. But there is value in having that many CAs, it allows one Lighthouse to serve many different isolated Nebula networks. Hosts accept traffic from the Lighthouse (CA: LH1), and Lighthouse accepts traffic from all hosts (CA: H1, CA: H2, CA: H3), but hosts don't accept traffic from each other, allowing small groups of networks from one lighthouse.  

The managing of the CAs for handling traffic is very efficient and looks like it could comfortably handle that many CAs. The reading of the CAs though is memory intensive, hence the change.

JackDoan:
Hi again, sorry to leave you in limbo for so long. 

I was re-reading this, and was kind of surprised to see that Golang doesn't have a pem.Decode that takes an io.Reader.  I yelled at an LLM and had it make me a split function for bufio.Scanner that splits on pem headers. To be honest, I'm not sure it's actually better, but I'm curious to hear your thoughts

Link: https://github.com/slackhq/nebula/compare/stream-cas-scanner?expand=1

Benchmark (on my machine, so ignore the ns/op):
```
goos: linux
goarch: amd64
pkg: github.com/slackhq/nebula
cpu: AMD Ryzen Threadripper 7970X 32-Cores          
BenchmarkReloadConfigWithCAs
BenchmarkReloadConfigWithCAs/100CAs
BenchmarkReloadConfigWithCAs/100CAs-64         	     267	   4045093 ns/op	  134651 B/op	    1487 allocs/op
BenchmarkReloadConfigWithCAs/250CAs
BenchmarkReloadConfigWithCAs/250CAs-64         	     127	   9331825 ns/op	  301770 B/op	    3292 allocs/op
BenchmarkReloadConfigWithCAs/500CAs
BenchmarkReloadConfigWithCAs/500CAs-64         	      64	  18384344 ns/op	  571463 B/op	    6295 allocs/op
BenchmarkReloadConfigWithCAs/1000CAs
BenchmarkReloadConfigWithCAs/1000CAs-64        	      31	  36674287 ns/op	 1112922 B/op	   12304 allocs/op
BenchmarkReloadConfigWithCAs/5000CAs
BenchmarkReloadConfigWithCAs/5000CAs-64        	       6	 182310119 ns/op	 5304321 B/op	   60336 allocs/op
PASS
```

maggie44:
It looks like both do the job, and I doubt this is an area of code you will look back at often.

maggie44:
Any progress on deciding which approach?

maggie44:
Following up again to see if we can merge this one? It’s been a while now and seems like we have good approaches ready to go.

maggie44:
Another nudge to see if this is something we could look to merge, it's 4 months since the contribution. 

cc. @JackDoan

johnmaguire:
Hi @maggie44 - we're planning to merge a fix, likely #1644. Leaving this open as we finalize a decision.

maggie44:
> Hi @maggie44 - we're planning to merge a fix, likely #1644. Leaving this open as we finalize a decision.

Thanks for the update, I will subscribe to the other PR.

johnmaguire:
@maggie44 thank you for the contribution. Merged #1644

PR #1566, opened 2026-01-07
Ensure pubkey coherency when rehydrating a handshake cert

brad-defined:
During handshakes, the peer's public key is included in the noise protocol portion of the packet. To minimize handshake packet size, the redundant pubkey information is removed from the peer's certificate portion of the packet.

During signature validation, the pubkey is taken from the handshake an rehydrated into the certificate. This extra processing creates opportunity for verification confusion. This PR strictly enforces in the certificate unmarshal functions that, if a public key is passed in (obtained from the noise handshake by callers), then the certificate being unmarshalled must not also have a public key present.

### DIFF
--- a/handshake_ix.go
+++ b/handshake_ix.go
@@ -171,6 +171,7 @@ func ixHandshakeStage1(f *Interface, via ViaSender, packet []byte, h *header.H) 
 		f.l.WithField("from", via).
 			WithField("handshake", m{"stage": 1, "style": "ix_psk0"}).
 			WithField("cert", remoteCert).Info("public key mismatch between certificate and handshake")
+		return
 	}
 
 	if remoteCert.Certificate.Version() != ci.myCert.Version() {
@@ -546,6 +547,7 @@ func ixHandshakeStage2(f *Interface, via ViaSender, hh *HandshakeHostInfo, packe
 		f.l.WithField("from", via).
 			WithField("handshake", m{"stage": 2, "style": "ix_psk0"}).
 			WithField("cert", remoteCert).Info("public key mismatch between certificate and handshake")
+		return true
 	}
 
 	if len(remoteCert.Certificate.Networks()) == 0 {


cert/cert_v1.go
```diff
@@ -435,14 +435,19 @@ func unmarshalCertificateV1(b []byte, publicKey []byte) (*certificateV1, error)
...
+	// If a public key is passed in as an argument, the certificate pubkey must be empty
+	// and the passed-in pubkey copied into the cert.
 	if len(publicKey) > 0 {
-		nc.details.publicKey = publicKey
+		if len(nc.details.publicKey) != 0 {
+			return nil, ErrCertPubkeyPresent
+		}
+		nc.details.publicKey = make([]byte, len(publicKey))
+		copy(nc.details.publicKey, publicKey)
 	}
```
JackDoan, line 451:
```suggestion
	} else {
	    copy(nc.details.publicKey, rc.Details.PublicKey)
	}
```

maybe this, to more closely resemble cert-v2? I'm happy either way though.

We could also technically avoid the allocation on line 429 in the handshake case. If we replace the allocate+copy on 447-448, with a simple slice assignment, the cert's public key will have the same backing memory as the noise public key, which feels warm and fuzzy to me in some, but not all, ways.

brad-defined: I specifically wanted to avoid sharing the memory. All sorts of things are getting copied around in this code besides, keeping a separate copy here for the cert struct jives with that approach.

brad-defined: I took a stab at your suggestion.

cert/cert_v2.go
```diff
@@ -593,6 +593,11 @@ func unmarshalCertificateV2(b []byte, publicKey []byte, curve Curve) (*certifica
 	var rawPublicKey cryptobyte.String
 	if len(publicKey) > 0 {
 		rawPublicKey = publicKey
```
nbrownus: I think we should do a copy here as well and can flip the order to check if the raw cert has a public key first

cert/cert_v1.go
JackDoan: Makes sense, I like what you cooked up!

cert/cert_v2.go
brad-defined: PR updated

cert/cert_v2.go
```diff
@@ -592,7 +592,13 @@ func unmarshalCertificateV2(b []byte, publicKey []byte, curve Curve) (*certifica
 	// Maybe grab the public key
 	var rawPublicKey cryptobyte.String
 	if len(publicKey) > 0 {
-		rawPublicKey = publicKey
+		rawPublicKey = make(cryptobyte.String, len(publicKey))
+		copy(rawPublicKey, publicKey)
+		// If a public key is passed in, then the handshake certificate must
+		// not have a public key present
+		if input.PeekASN1Tag(TagCertPublicKey) {
```
JackDoan, line 597: super nit: it would be more efficient to check this before we allocate + copy

handshake_ix.go
```diff
@@ -166,6 +167,12 @@ func ixHandshakeStage1(f *Interface, via ViaSender, packet []byte, h *header.H)
 		return
 	}
 
+	if !bytes.Equal(remoteCert.Certificate.PublicKey(), ci.H.PeerStatic()) {
+		f.l.WithField("from", via).
```
JackDoan, line 171: this logs but doesn't seem to abort? It feels like a place where we'd want to bail, right?

handshake_ix.go
```diff
@@ -535,6 +542,11 @@ func ixHandshakeStage2(f *Interface, via ViaSender, hh *HandshakeHostInfo, packe
 		e.Info("Invalid certificate from host")
 		return true
 	}
+	if !bytes.Equal(remoteCert.Certificate.PublicKey(), ci.H.PeerStatic()) {
+		f.l.WithField("from", via).
+			WithField("handshake", m{"stage": 2, "style": "ix_psk0"}).
+			WithField("cert", remoteCert).Info("public key mismatch between certificate and handshake")
```
JackDoan, line 549: same as above

cert/cert_v2.go
JackDoan: you are 2 fast 4 me

handshake_ix.go
JackDoan: If we take this branch and the cert somehow passed `Recombine` and `VerifyCertificate`, something is very wrong. I guess you technically could use that as an argument either for "we should bail on the handshake" _or_ "we don't need to perform this check".

brad-defined: thanks, good call. I can't test this code path b/c I can't possibly get into this situation with the modified behavior of Recombine.

brad-defined: pr updated

handshake_ix.go
brad-defined: pr udpated

PR #1577, opened 2026-01-16
Allow a IPv6-only overlay to carry IPv4 unsafe_routes via SNAT

JackDoan:
How does it work? Looking at traffic flow is probably the easiest way to explain:

None of the hosts below have an IPv4 address on their nebula interface, but they have the appropriate routes set to carry IPv4 traffic through nebula anyway.

```
13:05:52.813046 nebula0 Out IP [underlay IP here] > 9.9.9.9: ICMP echo request, id 4, seq 320, length 64
----------below is on the unsafe_router---------
* nebula receives the packet, and SNATs [underlay IP here] to an IP in the link-local space, to avoid potential conflicts, while still allowing a return path.*
13:05:52.813262 nebula1 In  IP 169.254.55.96 > 9.9.9.9: ICMP echo request, id 4, seq 320, length 64
* here you can see Linux masquerading the link-local address once it's been routed onto the uplink*
13:05:52.813292 uplink0 Out IP 10.0.0.2 > 9.9.9.9: ICMP echo request, id 4, seq 320, length 64
13:05:52.817848 uplink0 In  IP 9.9.9.9 > 10.0.0.2: ICMP echo reply, id 4, seq 320, length 64
* reverse the masquerade for the reply *
13:05:52.817869 nebulatest1 Out IP 9.9.9.9 > 169.254.55.96: ICMP echo reply, id 4, seq 320, length 64
* nebula internally un-SNATs the link-local IP back to the correct underlay IP for the original client *
----------above is on the unsafe_router---------
13:05:52.818078 IP 9.9.9.9 > [underlay IP here]: ICMP echo reply, id 4, seq 320, length 64
```


.github/workflows/smoke/smoke.sh
```diff
@@ -119,11 +120,11 @@ echo
 echo " *** Testing conntrack"
 echo
 set -x
-# host2 can ping host3 now that host3 pinged it first
```
JackDoan: this behavior is no longer true for ICMP, so I switched the test to use UDP, which imo achieves the same goal. Please double check me though!

firewall.go
```diff
@@ -34,6 +50,9 @@ type conn struct {
 	// fields pack for free after the uint32 above
 	incoming     bool
 	rulesVersion uint16
+
+	//for SNAT support
+	snat snatInfo
```
JackDoan: I could probably save a nickel if this was a pointer

JackDoan: update: we need that nickel

firewall.go
```diff
@@ -185,9 +185,8 @@ func NewFirewall(l *logrus.Logger, tcpTimeout, UDPTimeout, defaultTimeout time.D
 		hasUnsafeNetworks = true
 	}
 
-	snatAddr := netip.Addr{}
-	if hasUnsafeNetworks && !hasV4Networks {
-		snatAddr = netip.MustParseAddr("169.254.55.96") //todo this needs to come from the config, or perhaps the tun
+	if !hasUnsafeNetworks || hasV4Networks {
+		snatAddr = netip.Addr{} //disable using the special snat address if it doesn't make sense to use it
```
jasikpark: What does this comment mean? Is the zero value the "special snat address"?

firewall.go
JackDoan: nickel saved

cert/cert_v2.go
```diff
@@ -441,7 +441,7 @@ func (c *certificateV2) validate() error {
 				}
 			} else if network.Addr().Is4() {
 				if !hasV4Networks {
-					return NewErrInvalidCertificateProperties("IPv4 unsafe networks require an IPv4 address assignment: %s", network)
+					//return NewErrInvalidCertificateProperties("IPv4 unsafe networks require an IPv4 address assignment: %s", network)
```
nbrownus: This is just needing #1583 merged in ya?

JackDoan: yeah but I just closed that per our conversation yesterday lmao

I should drop this line instead of commenting it out

firewall.go
```diff
@@ -22,10 +23,29 @@ import (
 	"github.com/slackhq/nebula/firewall"
 )
 
+var ErrCannotSNAT = errors.New("cannot snat this packet")
```
nbrownus: nit: `snat` in the message here but `SNAT` in the message below

firewall.go
```diff
@@ -131,7 +165,7 @@ type firewallLocalCIDR struct {
 
 // NewFirewall creates a new Firewall object. A TimerWheel is created for you from the provided timeouts.
 // The certificate provided should be the highest version loaded in memory.
-func NewFirewall(l *logrus.Logger, tcpTimeout, UDPTimeout, defaultTimeout time.Duration, c cert.Certificate) *Firewall {
+func NewFirewall(l *logrus.Logger, tcpTimeout, UDPTimeout, defaultTimeout time.Duration, c cert.Certificate, snatAddr netip.Addr) *Firewall {
```
nbrownus: Seems we can remove `snatAddr`?

firewall.go
```diff
@@ -207,7 +241,7 @@ func NewFirewallFromConfig(l *logrus.Logger, cs *CertState, c *config.C) (*Firew
 		c.GetDuration("firewall.conntrack.udp_timeout", time.Minute*3),
 		c.GetDuration("firewall.conntrack.default_timeout", time.Minute*10),
 		certificate,
-		//TODO: max_connections
+		netip.Addr{},
```
nbrownus: Looks like we can remove this

overlay/tun.go
```diff
@@ -129,3 +130,76 @@ func selectGateway(dest netip.Prefix, gateways []netip.Prefix) (netip.Prefix, er
 
 	return netip.Prefix{}, fmt.Errorf("no gateway found for %v in the list of vpn networks", dest)
 }
+
+func genLinkLocal() netip.Prefix {
+	octets := []byte{169, 254, 0, 0}
+	_, _ = rand.Read(octets[2:4])
+	if octets[3] == 0 {
+		octets[3] = 1 //please no .0 addresses
```
nbrownus, line 148: Why not just re-roll the `rand.Read` if you get values you don't want?

overlay/tun.go
```diff
@@ -129,3 +130,76 @@ func selectGateway(dest netip.Prefix, gateways []netip.Prefix) (netip.Prefix, er
 
 	return netip.Prefix{}, fmt.Errorf("no gateway found for %v in the list of vpn networks", dest)
 }
+
+func genLinkLocal() netip.Prefix {
```
nbrownus: If this accepts an `io.Reader` we can test the behavior on the rejected values

nbrownus (commented):
`tun.unsafe_origin_address_for_4over6` and `tun.snat_address_for_4over6` are mutually exclusive so we could have just 1 config option, `tun.unsafe_ipv4_snat_address`? The description of the config option is going to have to do all the heavy lifting anyway and I think it makes sense in both the unsafe route subscriber and unsafe router. It would be neat if this was reloadable but I don't think it's a requirement since changing this address could break existing flows and/or listeners.

We also need a config option to entirely opt out of this behavior, something like `tun.allow_unsafe_ipv4_snat` which we can default to `true`, this should be reloadable.

overlay/device.go
```diff
@@ -11,6 +11,9 @@ type Device interface {
 	io.ReadWriteCloser
 	Activate() error
 	Networks() []netip.Prefix
+	UnsafeNetworks() []netip.Prefix
```
nbrownus, line 14: This is a list of unsafe networks from the local certificate and is only used to calculate `prepareSnatAddr`. I don't think we need to store the slice on the tun device or expose it here.

overlay/device.go
```diff
@@ -11,6 +11,9 @@ type Device interface {
...
+	UnsafeIPv4OriginAddress() netip.Prefix
```
nbrownus, line 15: Matching my previous comment about maybe combining these two addresses in config to just 1, we could also simplify this and merge `UnsafeIPv4OriginAddress` and `SNATAddress` into just `SNATAddress`

overlay/tun_android.go
```diff
@@ -91,6 +96,18 @@ func (t *tun) Networks() []netip.Prefix {
...
+func (t *tun) UnsafeNetworks() []netip.Prefix {
+	return t.UnsafeNetworks()
+}
+
+func (t *tun) UnsafeIPv4OriginAddress() netip.Prefix {
+	return t.unsafeIPv4Origin
+}
+
+func (t *tun) SNATAddress() netip.Prefix {
+	return netip.Prefix{}
```
nbrownus, line 108: I know some folks want to use android devices as unsafe routers

overlay/tun_darwin.go
```diff
@@ -545,6 +557,18 @@ func (t *tun) Networks() []netip.Prefix {
...
+func (t *tun) UnsafeNetworks() []netip.Prefix {
+	return t.unsafeNetworks
+}
+
+func (t *tun) UnsafeIPv4OriginAddress() netip.Prefix {
+	return t.unsafeIPv4Origin
+}
+
+func (t *tun) SNATAddress() netip.Prefix {
+	return netip.Prefix{}
```
nbrownus, line 569: Darwin as a router?

overlay/tun_freebsd.go
```diff
@@ -446,6 +453,18 @@ func (t *tun) Networks() []netip.Prefix {
...
+func (t *tun) UnsafeNetworks() []netip.Prefix {
+	return t.unsafeNetworks
+}
+
+func (t *tun) UnsafeIPv4OriginAddress() netip.Prefix {
+	return t.unsafeIPv4Origin
+}
+
+func (t *tun) SNATAddress() netip.Prefix {
+	return netip.Prefix{}
```
nbrownus, line 465: freebsd as a router?

overlay/tun_ios.go
```diff
@@ -69,6 +72,8 @@ func (t *tun) reload(c *config.C, initial bool) error {
 		return nil
 	}
 
+	t.unsafeIPv4Origin = prepareUnsafeOriginAddr(t, t.l, c, routes)
```
nbrownus, line 75: We likely need to assert this address comes from config since I don't think we can assign it to the device after it's running.

overlay/user.go
```diff
@@ -36,6 +36,17 @@ type UserDevice struct {
 	inboundWriter *io.PipeWriter
 }
 
+func (d *UserDevice) UnsafeNetworks() []netip.Prefix {
+	return nil
+}
+
+func (d *UserDevice) SNATAddress() netip.Prefix {
+	return netip.Prefix{}
```
nbrownus, line 44: I imagine we would want SNAT to work for user devices

overlay/device.go
JackDoan: The reason I have them separate is on a host that _is_ an unsafe_router that also is a client of a different unsafe router, they need to be different.

overlay/tun.go
JackDoan: it just needs to be _different_, not actually random, I'm not fussed about the entropy loss here.

overlay/tun.go
JackDoan: ah that's a good idea

overlay/tun_android.go
JackDoan: can Android devices set masquerade rules?

overlay/tun_darwin.go
JackDoan: "scoped for the next iteration"

overlay/tun_freebsd.go
JackDoan: "scoped for the next iteration"

overlay/tun_linux.go
```diff
@@ -415,6 +454,15 @@ func (t *tun) Activate() error {
 		return fmt.Errorf("failed to run tun device: %s", err)
 	}
 
+	//todo hmmmmmm
+	//pretty sure this is avoidable
```
JackDoan: it is, remove this block

overlay/tun_linux.go
```diff
@@ -427,6 +475,26 @@ func (t *tun) setMTU() {
...
+func (t *tun) setSnatRoute() error {
+	dr := &net.IPNet{
+		IP:   t.snatAddr.Masked().Addr().AsSlice(),
+		Mask: net.CIDRMask(t.snatAddr.Bits(), t.snatAddr.Addr().BitLen()),
+	}
+
+	nr := netlink.Route{
+		LinkIndex: t.deviceIndex,
+		Dst:       dr,
+		//todo do we need these other options?
```
JackDoan: it might be good to uncomment the MTU and MSS flags?

overlay/tun_android.go
nbrownus: Yep

firewall.go
JackDoan: resolving bc this no longer exists, thankfully

firewall.go
JackDoan: yep

firewall.go
JackDoan: yyyyyyyep

overlay/tun_android.go
JackDoan: "scoped for the next iteration"

overlay/user.go
JackDoan: I could see a user device being a client/origin, but I'm not sure they'd be able to be routers?

overlay/device.go
JackDoan: In principle I agree, buuuut

1) it's a property of the tun struct, which is implemented per-platform
2) I didn't want to copy/paste prepareSnatAddr per-platform
3) That means it needs to be a method on the interface
4) That means it needs to be public

I could use a bool, but I figured this had the potential to be more useful one day

overlay/device.go
JackDoan: has to do with intricacies of how Linux does packet delivery, and the way we're abusing link-local addresses for SNAT. If we have the address on the tun, Linux doesn't want to deliver a packet numbered with that address to the tun if it came from another interface. But, for whatever reason, it will happily forward the packet to the tun if it thinks the tun is not the final destination of the packet.

overlay/tun_android.go
```diff
@@ -76,6 +79,8 @@ func (t *tun) reload(c *config.C, initial bool) error {
 		return nil
 	}
 
+	t.unsafeIPv4Origin = prepareUnsafeOriginAddr(t, t.l, c, routes)
```
nbrownus, line 82: Same story as iOS, we will need to assign this in kotlin and tell nebula about it

overlay/tun.go
```diff
@@ -22,22 +24,22 @@ func (e *NameError) Error() string {
 }
 
 // TODO: We may be able to remove routines
-type DeviceFactory func(c *config.C, l *logrus.Logger, vpnNetworks []netip.Prefix, routines int) (Device, error)
+type DeviceFactory func(c *config.C, l *logrus.Logger, vpnNetworks []netip.Prefix, unsafeNetworks []netip.Prefix, routines int) (Device, error)
```
nbrownus, line 27: Cert reload does not reject changing unsafe networks in the certificate so we can technically reload into and out of being an unsafe router. Since we are getting the unsafe networks independently from the cert here, we are locked into a point in time behavior that requires a restart to fix. Would be nice to log this in `pki.go` if we detect the situation (unsafe networks changed, includes ipv4 unsafe networks, networks lacks an ipv4 network)

JackDoan: thank you for sharing your truth

overlay/tun_linux.go
```diff
@@ -400,7 +433,13 @@ func (t *tun) Activate() error {
 	//set route MTU
 	for i := range t.vpnNetworks {
 		if err = t.setDefaultRoute(t.vpnNetworks[i]); err != nil {
-			return fmt.Errorf("failed to set default route MTU: %w", err)
+			return fmt.Errorf("failed to set default route MTU for %s: %w", t.vpnNetworks[i], err)
+		}
+	}
+	//TODO snat and be snatted
```
nbrownus: dead comment

overlay/tun_linux.go
```diff
@@ -427,6 +466,26 @@ func (t *tun) setMTU() {
 	}
 }
 
+func (t *tun) setSnatRoute() error {
```
nbrownus, line 469: Would be nice to comment why we don't establish this address on the tun device itself as well

hostmap.go
```diff
@@ -224,6 +224,9 @@ const (
 	NetworkTypeVPNPeer
 	// NetworkTypeUnsafe is a network from Certificate.UnsafeNetworks()
 	NetworkTypeUnsafe
+	// NetworkTypeUncheckedSNATPeer is used to indicate traffic we're willing to route, but never deliver to a NetworkTypeVPN
```
nbrownus: I think `NetworkTypeUnverifiedSNATPeer` might be more clear as this is indicating the packet came from a flow using a `UnsafeIPv4OriginAddress` which can not be cryptographically verified, this is not to say that we did not cryptographically verify the source sent the packet, just that the address is not one encoded in the peers certificate.

interface.go
```diff
@@ -56,6 +56,7 @@ type Interface struct {
 	inside                overlay.Device
 	pki                   *PKI
 	firewall              *Firewall
+	snatAddr              netip.Addr
```
nbrownus: I think this is dead code

interface.go
```diff
@@ -344,6 +346,7 @@ func (f *Interface) reloadFirewall(c *config.C) {
 		f.l.WithError(err).Error("Error while creating firewall during reload")
 		return
 	}
+	fw.SetSNATAddressFromInterface(f)
```
nbrownus, line 348: Since we can't reload into or out of this behavior, is there a point in setting this since `SetSNATAddressFromInterface` doesn't cause a re-check of the conntrack entries in the firewall tables?

firewall.go
```diff
@@ -414,62 +450,224 @@ var ErrInvalidRemoteIP = errors.New("remote address is not in remote certificate
...
+	fp.RemotePort = oldPort
+	return ErrCannotSNAT
+}
+
+func (f *Firewall) applySnat(data []byte, fp *firewall.Packet, c *conn, hostinfo *HostInfo) error {
+	if !f.snatAddr.IsValid() {
+		return ErrCannotSNAT
+	}
+	if f.snatAddr == fp.LocalAddr { //a packet that came from UDP (incoming) should never ever have our snat address on it
+		return ErrSNATIdentityMismatch
```
nbrownus: We should make this a more aggressive error that's different from an invalid snat state

overlay/tun.go
JackDoan: turns out we have this pre-existing, so reloading unsafenetwork changes has never actually functioned (but only for the host itself!) 

```
func (f *Interface) reloadFirewall(c *config.C) {
	//TODO: need to trigger/detect if the certificate changed too
	if c.HasChanged("firewall") == false {
		f.l.Debug("No firewall config change detected")
		return
	}
```

JackDoan: I added log lines to try to warn the user if unsafenetworks changes during a reload, but we might wanna open an issue for this if we intent to support the behavior

interface.go
JackDoan: it's a great hook to do that when we want to support it though :D

overlay/tun_linux.go
JackDoan: added inside t.addRoutes

overlay/user.go
JackDoan: I think we need this but this PR is already the stuff of nightmares and I'd like to kick this can

overlay/tun_android.go
JackDoan: if we inject a value into the config file we're gucci :brain:

JackDoan:
some quirks deferred via #1620

JackDoan:
> `tun.unsafe_origin_address_for_4over6` and `tun.snat_address_for_4over6` are mutually exclusive so we could have just 1 config option, `tun.unsafe_ipv4_snat_address`? The description of the config option is going to have to do all the heavy lifting anyway and I think it makes sense in both the unsafe route subscriber and unsafe router. It would be neat if this was reloadable but I don't think it's a requirement since changing this address could break existing flows and/or listeners.

they're not mutually exclusive, but we do need better names :sob: 

> We also need a config option to entirely opt out of this behavior, something like `tun.allow_unsafe_ipv4_snat` which we can default to `true`, this should be reloadable.


we def need better names for the config options

examples/config.yml
```diff
@@ -283,6 +283,15 @@ tun:
...
   #use_system_route_table_buffer_size: 0
+  #
+  # When using the feature to carry IPv4 unsafe_routes through a IPv6-only overlay network (4over6),
+  # the system still needs to use IPv4 addresses to pass that traffic. 
+  # By default, they are randomly chosen out of 169.254.0.0/16 at startup, but if you need them to be static, you can set them here.
+  # Setting these options is not required for use of the 4over6 feature.
+  # this is the address that a client will use as a source IP for packets sent to the router
+  #unsafe_origin_address_for_4over6: 169.254.55.96
+  # this is the IP the router will use internally to hairpin-NAT 4over6 traffic before it hits the operating system.
+  #snat_address_for_4over6: 169.254.12.34
```
johnmaguire, line 294: Will we support disabling 4over6? (`4over6_enabled`?)

overlay/tun.go
```diff
@@ -129,3 +131,85 @@ func selectGateway(dest netip.Prefix, gateways []netip.Prefix) (netip.Prefix, er
...
+		out, err := netip.ParseAddr(a)
+		if err != nil {
+			l.WithField("value", a).WithError(err).Warn("failed to parse tun.unsafe_origin_address_for_4over6, will use a random value")
+		} else if !out.Is4() || !out.IsLinkLocalUnicast() {
+			l.WithField("value", out).Warn("tun.unsafe_origin_address_for_4over6 must be an IPv4 address")
+		} else if out.IsValid() {
+			return netip.PrefixFrom(out, 32)
+		}
+	}
+	return genLinkLocal(nil)
```
johnmaguire, line 184: Is there an advantage to making these randomized as opposed to choosing some specific hardcoded value as a default? One thought is that it may make for an OK "[magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)#Format_indicator)" that is useful in debugging unexpected packets.

overlay/tun.go
```diff
@@ -129,3 +131,85 @@ func selectGateway(dest netip.Prefix, gateways []netip.Prefix) (netip.Prefix, er
...
+	if !needed {
+		return netip.Prefix{}
+	}
+
+	if a := c.GetString("tun.snat_address_for_4over6", ""); a != "" {
+		out, err := netip.ParseAddr(a)
+		if err != nil {
+			l.WithField("value", a).WithError(err).Warn("failed to parse tun.snat_address_for_4over6, will use a random value")
+		} else if !out.Is4() || !out.IsLinkLocalUnicast() {
+			l.WithField("value", out).Warn("tun.snat_address_for_4over6 must be an IPv4 address")
```
johnmaguire, line 209:
```suggestion
			l.WithField("value", out).Warn("tun.snat_address_for_4over6 must be a link-local IPv4 address")
```

johnmaguire (commented):
Awesome work!

examples/config.yml
JackDoan: Yes I need to add this

overlay/tun.go
```diff
@@ -129,3 +131,85 @@ func selectGateway(dest netip.Prefix, gateways []netip.Prefix) (netip.Prefix, er
...
+	}
+	out, _ := netip.AddrFromSlice(octets)
+	return netip.PrefixFrom(out, 32)
+}
+
+// prepareUnsafeOriginAddr provides the IPv4 address used on IPv6-only clients that need to access IPv4 unsafe routes
+func prepareUnsafeOriginAddr(d Device, l *logrus.Logger, c *config.C, routes []Route) netip.Prefix {
+	if !d.Networks()[0].Addr().Is6() {
+		return netip.Prefix{} //if we have an IPv4 assignment within the overlay, we don't need an unsafe origin address
+	}
```
wadey, line 160: This only checks the first Network in the list, is this correct? should we be checking if any of the Networks in the list are ipv4? Is the ipv4 address always first?

overlay/tun.go
```diff
@@ -129,3 +131,85 @@ func selectGateway(dest netip.Prefix, gateways []netip.Prefix) (netip.Prefix, er
...
+			needed = true //todo should this only apply to unsafe routes? almost certainly
+			break
+		}
+	}
+	if !needed {
+		return netip.Prefix{}
+	}
+
+	//todo better config name for sure
+	if a := c.GetString("tun.unsafe_origin_address_for_4over6", ""); a != "" {
```
wadey, line 174: remove the todo before merging

overlay/tun_linux.go
```diff
@@ -170,6 +187,11 @@ func (t *tun) reload(c *config.C, initial bool) error {
 		return nil
 	}
 
+	if initial {
+		t.unsafeIPv4Origin = prepareUnsafeOriginAddr(t, t.l, c, routes) //todo MUST be different from t.snatAddr!
```
wadey, line 191: Shall we handle this "todo" before merge?

firewall.go
```diff
@@ -414,62 +456,224 @@ var ErrInvalidRemoteIP = errors.New("remote address is not in remote certificate
...
+
+// Drop returns an error if the packet should be dropped, explaining why. It
+// returns nil if the packet should not be dropped.
+func (f *Firewall) Drop(fp firewall.Packet, pkt []byte, incoming bool, h *HostInfo, caPool *cert.CAPool, localCache firewall.ConntrackCache) error {
 	table := f.OutRules
 	if incoming {
 		table = f.InRules
 	}
 
+	snatmode := fp.IsIPv4() && h.HasOnlyV6Addresses() && f.snatAddr.IsValid()
```
wadey, line 627: Drop() gets called for every packet, should we cache some of this so we dont have to recalculate it every packet? Like the loop in HasOnlyV6Addresses?

firewall_test.go
```diff
@@ -1532,3 +1550,59 @@ func resetConntrack(fw *Firewall) {
...
+			LocalPort:  10,
+			RemotePort: 90,
+			Protocol:   firewall.ProtoUDP,
+			Fragment:   false,
+		}
+		require.NoError(t, setup.fw.Drop(p, nil, true, h, cp, nil))
+	})
+	//t.Run("allow inbound unsafe route", func(t *testing.T) {
+	//	t.Parallel()
+	//	unsafePrefix := netip.MustParsePrefix("192.168.0.0/24")
```
wadey, line 1591: Is there a plan to finish this test or should we remove the commented out code?

JackDoan:
I do not recommend doing this

PR #1583, opened 2026-01-20
remove the requirement for a cert-v2 certificate to have an IPv4 Network to have IPv4 UnsafeNetworks

JackDoan:


JackDoan:
doin this all together I think

PR #1588, opened 2026-01-27
 Report if Nebula start fails because of tun device name

JackDoan:


overlay/tun_freebsd.go
```diff
@@ -266,7 +266,7 @@ func newTun(c *config.C, l *logrus.Logger, vpnNetworks []netip.Prefix, _ bool) (
 		}
 
 		// Set the device name
-		ioctl(fd, syscall.SIOCSIFNAME, uintptr(unsafe.Pointer(&ifrr)))
+		_ = ioctl(fd, syscall.SIOCSIFNAME, uintptr(unsafe.Pointer(&ifrr)))
```
JackDoan, line 269: this does not abort on freebsd, but it does on Linux. Is that intended?

nbrownus: Linux being where we started, it was imperative for us to have stats reported with the correct tun device name. I believe there are others that feel similarly.

As for why freebsd is more lenient, the name is the more restricted `tunN` and choosing a unique one is a minefield. The BSDs don't require the name matches the config as a result.

JackDoan: neat!

PR #1593, opened 2026-01-29
use wg tun library; batching improvements

jrwren:
* conntrack purging


firewall.go
```diff
@@ -492,14 +498,17 @@ func (f *Firewall) inConns(fp firewall.Packet, h *HostInfo, caPool *cert.CAPool,
 	conntrack := f.Conntrack
 	conntrack.Lock()
 
-	// Purge every time we test
-	ep, has := conntrack.TimerWheel.Purge()
-	if has {
-		f.evict(ep)
+	// Periodic purge instead of every lookup (major CPU savings)
```
nbrownus, line 501: I think at worst what this will do is allow flows to stay in conntrack beyond their expiry and every 1024th packet will have more latency than the rest.

I think it's safe to play a little looser with the expiry times since we have the rule version test for firewall re-evaluation and I see little point in re-evaluating the rules if the flow had expired but not purged since we have a packet matching the expired flow, it wouldn't be in conntrack if it wasn't allowed (and the fw rules version matched). In the end I expect some memory bloat, the most degenerate would be something like an `nmap` on an otherwise mostly idle host, but that exists today as well, would just recover slower.

If it is in fact a nice perf gain then I would suggest making it an isolated PR that we can merge sooner than the rest of this code.

nbrownus: Also for background, in the early days we tried 2 other approaches for purging.

1. A goroutine that ticked through and purged immediately. This caused some packets to be very delayed while that routine had the conntrack lock, depending on how many entries needed purging. The uneven per packet performance was not desirable.
2. Each packet evaluation checked and purged the entire list if there were any to purge. Same situation with wildly uneven per packet performance.

Amortizing the cost of conntrack maintenance into every packet was the best way to get reliable per packet performance.

jrwren: Thanks @nbrownus This is exactly the kind of feedback for which I was looking.

jrwren: This is why I shouldn't give Claude Code too much freedom.

jrwren:
I'm really not happy with how these changes perform

PR #1596, opened 2026-02-03
add stats.pprof flag to expose pprof on stats listener

jrwren:


jasikpark (commented):
lgtm

stats.go
```diff
@@ -119,6 +122,14 @@ func startPrometheusStats(l *logrus.Logger, i time.Duration, c *config.C, buildV
 		startFn = func() {
 			l.Infof("Prometheus stats listening on %s at %s", listen, path)
 			http.Handle(path, promhttp.HandlerFor(pr, promhttp.HandlerOpts{ErrorLog: l}))
+			if enablePprof {
+				l.Info("pprof endpoints enabled on stats listener")
+				http.HandleFunc("GET /debug/pprof/", pprof.Index)
```
JackDoan, line 127:
```suggestion
```

snip this, let's limit it to these "safer" profiles for now. The heap-ier profiles like /heap or /goroutines can leak heap data, and therefore keys

examples/config.yml
```diff
@@ -318,6 +318,11 @@ logging:
   #subsystem: nebula
   #interval: 10s
 
+  # enables pprof endpoints on the stats listener
+  # endpoints: /debug/pprof/, /debug/pprof/cmdline, /debug/pprof/profile,
+  #            /debug/pprof/symbol, /debug/pprof/trace
+  #pprof: false
```
JackDoan, line 324: it could be nice to make this a configurable list

JackDoan (changes requested):
I really want this feature, but taking a pprof is just such a different privilege level than looking at stats. 

IMO we should wade into this slowly, maybe by adding the pprof listener as a separate http ServeMux on a different port? Maybe even a localhost only bind, to start out?

There's also the "problem" of allowing non-root users on the local machine to take a heap or goroutine pprof, which could expose the live ekey/dkeys for connections, or even the contents of pki.key.

All of that said, I really really really wanna be able to take a /debug/pprof/profile without mucking with ssh!

jrwren:
> I really want this feature, but taking a pprof is just such a different privilege level than looking at stats.

I was thinking that it would be off by default and only enabled under debug situations where none of the drawbacks you described would be applicable.

It sounds like you have plans for leaving it on to take profiles on an ad-hoc basis?

I was really thinking about it as a development debugging feature and nothing more.

PR #1601, opened 2026-02-06
don't panic on bad ed25519 key lengths

JackDoan:
I think it's impossible (or very contrived) to create this condition, but defending against it is cheap.


jasikpark (commented):
lgtm

wadey (approved):

PR #1602, opened 2026-02-09
connection-track ICMP traffic

JackDoan:
Today, all individual ICMP flows are tracked together. This change splits tracking of ICMP per-flow, in order to facilitate future features

### DIFF
--- a/outside_test.go
+++ b/outside_test.go
@@ -167,7 +167,7 @@ func Test_newPacket_v6(t *testing.T) {
 	}
 
 	icmp := layers.ICMPv6{
-		TypeCode: 0x55,
+		TypeCode: layers.ICMPv6TypeEchoRequest,
 		Checksum: 0x1234,
 	}
 


.github/workflows/smoke/smoke.sh
```diff
@@ -37,17 +37,18 @@ docker run --name host4 --device /dev/net/tun:/dev/net/tun --cap-add NET_ADMIN -
 sleep 1
 
 # grab tcpdump pcaps for debugging
-docker exec lighthouse1 tcpdump -i nebula1 -q -w - -U 2>logs/lighthouse1.inside.log >logs/lighthouse1.inside.pcap &
+docker exec lighthouse1 tcpdump -i tun0 -q -w - -U 2>logs/lighthouse1.inside.log >logs/lighthouse1.inside.pcap &
 docker exec lighthouse1 tcpdump -i eth0 -q -w - -U 2>logs/lighthouse1.outside.log >logs/lighthouse1.outside.pcap &
-docker exec host2 tcpdump -i nebula1 -q -w - -U 2>logs/host2.inside.log >logs/host2.inside.pcap &
+docker exec host2 tcpdump -i tun0 -q -w - -U 2>logs/host2.inside.log >logs/host2.inside.pcap &
```
JackDoan, line 42: this makes the logs actually work

.github/workflows/smoke/smoke.sh
```diff
@@ -119,11 +120,11 @@ echo
 echo " *** Testing conntrack"
 echo
 set -x
-# host2 can ping host3 now that host3 pinged it first
-docker exec host2 ping -c1 192.168.100.3
-# host4 can ping host2 once conntrack established
-docker exec host2 ping -c1 192.168.100.4
-docker exec host4 ping -c1 192.168.100.2
+
+# host2 speaking to host4 on UDP 4000 should allow it to reply, when firewall rules would normally not permit this
```
JackDoan, line 124: please note that the above ping test is no longer valid and has been removed

JackDoan:
I'm also considering dropping support for the `Code` argument in firewall rules. As best as I can tell, it doesn't actually work, and filtering on `Type` would make more sense if we wanted to support that.

routing/balance.go
```diff
@@ -12,7 +12,12 @@ import (
 //   - https://github.com/skeeto/hash-prospector
 //     [16 21f0aaad 15 d35a2d97 15] = 0.10760229515479501
 func hashPacket(p *firewall.Packet) int {
-	x := (uint32(p.LocalPort) << 16) | uint32(p.RemotePort)
+	var x uint32
+	if p.Protocol == firewall.ProtoICMP {
+		x = 0 //Don't attempt to use ICMP's code/id/etc to balance
```
jrwren: why not?

JackDoan: only because we weren't doing it before, and I wasn't confident enough to change the behavior :sweat_smile:

nbrownus: This only applies to an unsafe route with multiple gateways. Not using the icmp code means it's guaranteed 2 hosts doing ICMP will always use the same gateway.

JackDoan: I've been convinced to let ICMP get hashed too, so that if TCP is "50% broken" due to a dead router, ICMP also will be

outside.go
```diff
@@ -329,7 +329,7 @@ func parseV6(data []byte, incoming bool, fp *firewall.Packet) error {
 		switch proto {
 		case layers.IPProtocolICMPv6, layers.IPProtocolESP, layers.IPProtocolNoNextHeader:
 			fp.Protocol = uint8(proto)
-			fp.RemotePort = 0
+			fp.RemotePort = 0 //we don't attempt to parse ICMPv6 because we don't SNAT it
```
nbrownus: This will lead to maybe some confusion.

In previous nebula, If I do not allow inbound icmp and I ping a remote host then that remote can also ping me while the conntrack entry is alive. This changes that behavior for ipv4 but keeps it for ipv6.

JackDoan: good point, pushed this for v6 as well

firewall/packet.go
```diff
@@ -22,7 +22,11 @@ const (
 type Packet struct {
 	LocalAddr  netip.Addr
 	RemoteAddr netip.Addr
-	LocalPort  uint16
+	// LocalPort is the destination port for incoming traffic, or the source port for outgoing.
+	// For ICMP, it's "code".
```
nbrownus: We can remove the ICMP bit here, ya?

outside.go
```diff
@@ -423,34 +434,38 @@ func parseV4(data []byte, incoming bool, fp *firewall.Packet) error {
...
-		if fp.Fragment || fp.Protocol == firewall.ProtoICMP {
-			fp.RemotePort = 0
-			fp.LocalPort = 0
-		} else {
-			fp.LocalPort = binary.BigEndian.Uint16(data[ihl : ihl+2])
-			fp.RemotePort = binary.BigEndian.Uint16(data[ihl+2 : ihl+4])
-		}
+	}
+
+	if fp.Fragment {
```
nbrownus, line 462: Can we recombine this with the above if statement to avoid double checking

JackDoan: is ok

PR #1608, opened 2026-02-12
Apply `go fix`

wadey:
Apply `go fix` with go1.26.0

### DIFF
--- a/firewall.go
+++ b/firewall.go
@@ -804,10 +804,8 @@ func (fr *FirewallRule) isAny(groups []string, host string, cidr string) bool {
 		return true
 	}
 
-	for _, group := range groups {
-		if group == "any" {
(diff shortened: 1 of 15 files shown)


firewall.go
```diff
@@ -804,10 +804,8 @@ func (fr *FirewallRule) isAny(groups []string, host string, cidr string) bool {
 		return true
 	}
 
-	for _, group := range groups {
-		if group == "any" {
-			return true
-		}
+	if slices.Contains(groups, "any") {
```
nbrownus, line 807: Gets turned into, more or less:
```
func Index[S ~[]E, E comparable](s S, v E) int {
	for i := range s {
		if v == s[i] {
			return i
		}
	}
	return -1
}
```

PR #1609, opened 2026-02-12
firewall: icmp no longer requires a port spec

JackDoan:


firewall.go
nbrownus, line 275: Seems like a good idea to force start and end port to any in the AddRule as well, could simplify the FromConfig func as a result (ignore port parsing if icmp)

JackDoan: one could argue that a special case like that makes the function less simple, but I'll take a look

JackDoan: Wait, I did this:

>could simplify the FromConfig func as a result (ignore port parsing if icmp)

see firewall.go:358

PR #1622, opened 2026-03-06
add sshd.sandbox_dir config option

jrwren:
Sanitize SSH profile paths (ssh.go:514,683,719) — restrict os.Create(a[0]) to a safe directory. Add a config option in the config file to specify the sandbox directory.

### DIFF
--- a/ssh.go
+++ b/ssh.go
@@ -192,7 +192,8 @@ func attachCommands(l *logrus.Logger, c *config.C, ssh *sshd.SSHServer, f *Inter
 	// sandboxDir defaults to a dir in temp. The intention is that end user will
 	// create this dir as needed. Overriding this config value to "" allows
 	// writing to anywhere in the system.
-	sandboxDir := c.GetString("sshd.sandbox_dir", "/var/tmp/nebula-debug")
+	defaultDir := filepath.Join(os.TempDir(), "nebula-debug")
+	sandboxDir := c.GetString("sshd.sandbox_dir", defaultDir)
 
 	ssh.RegisterCommand(&sshd.Command{
 		Name:             "list-hostmap",
--- a/ssh.go
+++ b/ssh.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/netip"
 	"os"
+	"path/filepath"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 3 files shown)


ssh.go
```diff
@@ -188,6 +189,11 @@ func configSSH(l *logrus.Logger, ssh *sshd.SSHServer, c *config.C) (func(), erro
 }
 
 func attachCommands(l *logrus.Logger, c *config.C, ssh *sshd.SSHServer, f *Interface) {
+	// sandboxDir defaults to a dir in temp. The intention is that end user will
+	// create this dir as needed. Overriding this config value to "" allows
+	// writing to anywhere in the system.
+	sandboxDir := c.GetString("sshd.sandbox_dir", "/var/tmp/nebula-debug")
```
johnmaguire: /var/tmp/nebula-debug is probably not the right path on Windows - not sure about other systems like Darwin/BSDs.

jrwren: Thank you. I was thinking /var/tmp might be nicer than /tmp, but since there is no cross platform version of that, $TMP/$TEMP must suffice.  Please take another look.

ssh.go
```diff
@@ -506,13 +519,39 @@ func sshListLighthouseMap(lightHouse *LightHouse, a any, w sshd.StringWriter) er
...
+	}
+
+	// Clean and resolve the path relative to the sandbox directory
+	if !filepath.IsAbs(filePath) {
+		filePath = filepath.Join(sandboxDir, filePath)
+	}
+	cleaned := filepath.Clean(filePath)
+
+	// Ensure the resolved path is within the sandbox directory
+	if !strings.HasPrefix(cleaned, filepath.Clean(sandboxDir)+string(filepath.Separator)) && cleaned != filepath.Clean(sandboxDir) {
```
johnmaguire: Nit: the && case here is interesting and a good thought but not clearly explained in the error and easy to gloss over. A separate conditional with a separate error may be a slight improvement.

johnmaguire (dismissed):

ssh.go
jrwren: That is better. TY.

johnmaguire (approved):

PR #1634, opened 2026-03-18
Fix sshd goroutine leak and other cleanup

nbrownus:
Fixes a couple of memory leaks in the ssh server:

- `exec` and interactive sessions (that don't use `logout`) would leak a go routine, the session, and session tracking.
- Interactive sessions that use `logout` can double close the session and leak a go routine.
- ~~An authentication failure would leak the TCP socket.~~
- Removed 2 unused fields on the `SSHServer` struct (`helpCommand` and `certChecker`).

Confirmed by checking the number of go routines, then running 10k `exec` commands, and checking the number of go routines again. Pre change, I had 10 routines before `exec`, and 10k+ after `exec`. 

Closes #1633

### DIFF
--- a/sshd/server.go
+++ b/sshd/server.go
@@ -16,16 +16,13 @@ type SSHServer struct {
 	config *ssh.ServerConfig
 	l      *logrus.Entry
 
-	certChecker *ssh.CertChecker
-
 	// Map of user -> authorized keys
 	trustedKeys map[string]map[string]bool
 	trustedCAs  []ssh.PublicKey
 
 	// List of available commands
-	helpCommand *Command
-	commands    *radix.Tree
-	listener    net.Listener
+	commands *radix.Tree
+	listener net.Listener
 
 	// Locks the conns/counter to avoid concurrent map access
 	connsLock sync.Mutex
@@ -184,7 +181,11 @@ func (s *SSHServer) run() {
 
 		if err != nil {
 			l := s.l.WithError(err).WithField("remoteAddress", c.RemoteAddr())
-			if conn != nil {
+			if conn == nil {
+				// conn is nil when the handshake failed before authentication
+				// close the raw TCP connection to avoid leaking the file descriptor.
+				c.Close()
+			} else {
 				l = l.WithField("sshUser", conn.User())
 				conn.Close()
 			}
--- a/sshd/server.go
+++ b/sshd/server.go
@@ -181,11 +181,7 @@ func (s *SSHServer) run() {
 
 		if err != nil {
 			l := s.l.WithError(err).WithField("remoteAddress", c.RemoteAddr())
-			if conn == nil {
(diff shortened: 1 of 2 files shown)


jrwren (dismissed):

sshd/server.go
brad-defined, line 176: I vaguely recall hearing that this call must happen in a different goroutine, as otherwise it may block the for loop during a bad connection handshake and prevent another connection from being established.

sshd/server.go
```diff
@@ -184,7 +181,11 @@ func (s *SSHServer) run() {
 
 		if err != nil {
 			l := s.l.WithError(err).WithField("remoteAddress", c.RemoteAddr())
-			if conn != nil {
+			if conn == nil {
+				// conn is nil when the handshake failed before authentication
```
brad-defined: NewServerConn [says](https://pkg.go.dev/golang.org/x/crypto/ssh#NewServerConn):
`NewServerConn starts a new SSH server with c as the underlying transport. It starts with a handshake and, if the handshake is unsuccessful, it closes the connection and returns an error. The Request and NewChannel channels must be serviced, or the connection will hang.`

So, it will close the connection itself?

JackDoan: indeed (inside NewServerConn)

```
	perms, err := s.serverHandshake(&fullConf)
	if err != nil {
		c.Close()
		return nil, nil, nil, err
	}
```

nbrownus: Thanks, removed that part.

sshd/server.go
nbrownus: Yeah we should move this

sshd/session.go
```diff
@@ -82,6 +87,7 @@ func (s *session) handleRequests(in <-chan *ssh.Request, channel ssh.Channel) {
 			cErr := ssh.Unmarshal(req.Payload, &payload)
 			if cErr != nil {
 				req.Reply(false, nil)
+				channel.Close()
```
brad-defined, line 90: IIUC the `channel` is returned by newChannel.Accept() in handleChannels, and it must have Close called on it to free up resources.

In this for loop, there are code paths that do not result in channel.Close() being called. For instance, when the in ssh.Request golang chan is emptied, the for loop will terminate, and in that case channel.Close() is not invoked.

I think these channel.Close() calls should be removed, and a single defer channel.Close() added to the tope of handleRequests...

jasikpark:
context from claude for commit 2:

> The second commit references an upstream fix in golang.org/x/crypto/ssh (commit 4e5a261, December 2023) that ensures `NewServerConn` closes the underlying `net.Conn` on all error return paths, not just after failed handshake. This matches `NewClientConn` behavior and eliminates the need for explicit `c.Close()` on handshake errors.
> 
> Reference: https://github.com/golang/crypto/commit/4e5a26183ecb4f9a0f85c8f8dbe7982885435436

PR #1635, opened 2026-03-20
fix: guard QueryCert against panic on short/empty QNAME

gnesher:
## Summary
- `QueryCert` slices `data[:len(data)-1]` to strip a trailing dot, which panics when `data` is empty (slice bounds `[:-1]`)
- Added a `len(data) < 2` guard to return early for inputs shorter than a minimal valid `x.` form
- While `miekg/dns` currently rejects wire-format packets that would produce an empty QNAME, Nebula should not rely on library behavior for crash safety

## Test plan
- [x] `TestQueryCert_short_inputs` — verifies `QueryCert("")` and `QueryCert(".")` return `""` without panic
- [x] `TestParseQuery_TXT_empty_and_root_qname` — verifies `parseQuery` handles root zone `"."` and empty `Name` TXT queries without panic, returning NXDOMAIN
- [x] Existing `TestParsequery` and `Test_getDnsServerAddr` still pass

Discovered this while looking for security vulnerabilities, this isn't a bug bounty level issue but still seemed worth a fix (done via llm)

### DIFF
--- a/dns_server.go
+++ b/dns_server.go
@@ -241,6 +241,9 @@ func (d *dnsServer) Query(q uint16, data string) (netip.Addr, bool) {
 }
 
 func (d *dnsServer) QueryCert(data string) string {
+	if len(data) < 2 {
+		return ""
+	}
 	ip, err := netip.ParseAddr(data[:len(data)-1])
 	if err != nil {
 		return ""
--- a/dns_server_test.go
+++ b/dns_server_test.go
@@ -16,6 +16,19 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+type stubDNSWriter struct{}
+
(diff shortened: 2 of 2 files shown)


JackDoan:
Hey @gnesher, sorry we let this sit so long, nice find! I'm happy to merge this, but we need to make golangci-lint happy. I can let you get it if you want, or if it's cool with you I can push to your branch

gnesher:
> Hey @gnesher, sorry we let this sit so long, nice find! I'm happy to merge this, but we need to make golangci-lint happy. I can let you get it if you want, or if it's cool with you I can push to your branch

I'm out for the next week or so, if it's no trouble please push it in

nbrownus (approved):

PR #1640, opened 2026-03-30
SSH handshake in goroutine and defer close

brad-defined:
- SSH handshakes happen in the Accept goroutine, so a slow handshake could block ssh access to other clients
- defer s.Close in handleRequests ensures all resources are cleaned up in all cases
- This is my artisanal take on #1634

### DIFF
--- a/sshd/server.go
+++ b/sshd/server.go
@@ -2,10 +2,10 @@ package sshd
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"net"
-	"sync"
 
 	"github.com/armon/go-radix"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


sshd/session.go
nbrownus: We should remove this close since the defer handles it

nbrownus: Close can block and leak a goroutine if called multiple times which fights against the desire in this PR.

nbrownus: This will double close with handleRequests and trigger a deadlock on the exitChan

brad-defined:
PR updated. Moved the defer to an earlier channel reader loop so that multiple Close() calls can be avoided, and also made it safe to call Close() multiple times.

sshd/server.go
```diff
@@ -175,44 +175,45 @@ func (s *SSHServer) run() {
...
-
-		conn, chans, reqs, err := ssh.NewServerConn(c, s.config)
-		fp := ""
-		if conn != nil {
-			fp = conn.Permissions.Extensions["fp"]
-		}
-
-		if err != nil {
-			l := s.l.WithError(err).WithField("remoteAddress", c.RemoteAddr())
+		go func(c net.Conn) {
```
nbrownus, line 179: There's a chance we race a `Stop()` here and leak the newly finished handshake on a closed server. The conn will stay alive until the client disconnects on its own.

brad-defined: great point! I didn't want to touch that thing, but I think a context tree is a workable approach?

sshd/session.go
```diff
@@ -170,10 +169,8 @@ func (s *session) dispatchCommand(line string, w StringWriter) {
 	}
 
 	_ = execCommand(c, args[1:], w)
-	return
 }
 
 func (s *session) Close() {
-	s.c.Close()
-	s.exitChan <- true
+	s.cancel()
```
nbrownus, line 176: nit: s.c.Close() before s.Cancel() to gracefully disconnect the client, should be safe without it but the user experience changes without it (TCP close vs SSH session close)

brad-defined: PR updated

nbrownus (approved):

PR #1642, opened 2026-03-30
refactor readOutsidePackets

wadey:
The layout of this method is confusing and relies on certain parts to return early for things to work correctly.

Change the ordering of the logic so that we do this:

- Handle unencrypted packets
- Decrypt packet
- Handle encrypted packets

This way, nothing can sneak through unencrypted to where it shouldn't be.


outside.go
```diff
@@ -52,23 +52,24 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
...
-			if !f.decryptToTun(hostinfo, h.MessageCounter, out, packet, fwPacket, nb, q, localCache) {
-				return
-			}
 		case header.MessageRelay:
 			// The entire body is sent as AD, not encrypted.
 			// The packet consists of a 16-byte parsed Nebula header, Associated Data-protected payload, and a trailing 16-byte AEAD signature value.
 			// The packet is guaranteed to be at least 16 bytes at this point, b/c it got past the h.Parse() call above. If it's
 			// otherwise malformed (meaning, there is no trailing 16 byte AEAD value), then this will result in at worst a 0-length slice
 			// which will gracefully fail in the DecryptDanger call.
+			if !f.handleEncrypted(ci, via, h) {
```
JackDoan, line 70: I think this is a great time to rename this function to something like `windowCheck`

outside.go
JackDoan, line 95: We should add `default` cases for every switch, especially user-controlled values

outside.go
```diff
@@ -119,114 +119,78 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
 						// Forward this packet through the relay tunnel
 						// Find the target HostInfo
 						f.SendVia(targetHI, targetRelay, signedPayload, nb, out, false)
-						return
 					case TerminalType:
 						hostinfo.logger(f.l).Error("Unexpected Relay Type of Terminal")
```
JackDoan, line 123: this feels like it should early-return?

outside.go
```diff
@@ -119,114 +119,78 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
...
 					}
 				} else {
 					hostinfo.logger(f.l).WithFields(logrus.Fields{"relayTo": relay.PeerAddr, "relayFrom": hostinfo.vpnAddrs[0], "targetRelayState": targetRelay.State}).Info("Unexpected target relay state")
```
JackDoan, line 126: same here?

outside.go
```diff
@@ -119,114 +119,78 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
...
 
-		//TODO: assert via is not relayed
-		lhf.HandleRequest(via.UdpAddr, hostinfo.vpnAddrs, d, f)
+	// All remaining packets are encrypted
+	if !f.handleEncrypted(ci, via, h) {
+		return
+	}
 
-		// Fallthrough to the bottom to record incoming traffic
+	out, err = f.decrypt(hostinfo, h.MessageCounter, out, packet, h, nb)
```
JackDoan, line 146: pretty minor but it would be neat+cheap to assert hostinfo non-nil here to make my IDE happy

outside.go
```diff
@@ -52,23 +52,24 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
...
-		if !f.handleEncrypted(ci, via, h) {
-			return
-		}
-
 		switch h.Subtype {
-		case header.MessageNone:
-			if !f.decryptToTun(hostinfo, h.MessageCounter, out, packet, fwPacket, nb, q, localCache) {
-				return
-			}
 		case header.MessageRelay:
```
JackDoan, line 64: Hmm this is a sucky problem. I'm tempted to suggest scooting this down under line 142 as an `if h.Type == header.Message && h.Subtype == header. MessageRelay`?

outside.go
```diff
@@ -119,114 +119,78 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
...
-		f.relayManager.HandleControlMsg(hostinfo, d, f)
+		f.relayManager.HandleControlMsg(hostinfo, out, f)
 
 	default:
-		f.messageMetrics.Rx(h.Type, h.Subtype, 1)
 		hostinfo.logger(f.l).Debugf("Unexpected packet received from %s", via)
 		return
 	}
 
 	f.handleHostRoaming(hostinfo, via)
```
JackDoan, line 193: it's a little repetitive, but I kinda wanna move this into each "success" case for handling a message. I think it would be best if once we hit a `case` statement, that's terminal, and we never come out of it.

We could even enforce this at compile time by making `readOutsidePackets` return... something (`return struct{}{}` is cute but ugly) and removing the bottom-level return, to force us to demonstrate we handle every possible case

outside.go
```diff
@@ -119,114 +119,78 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
...
-				WithField("packet", packet).
-				Error("Failed to decrypt test packet")
-			return
-		}
+	case header.LightHouse:
+		//TODO: assert via is not relayed
+		lhf.HandleRequest(via.UdpAddr, hostinfo.vpnAddrs, out, f)
 
+	case header.Test:
 		if h.Subtype == header.TestRequest {
```
JackDoan, line 171: guh this is also hard to resolve structurally, but because we look at `subtype` we probably want to not count "invalid" subtypes? Maybe?

outside.go
```diff
@@ -499,28 +463,16 @@ func (f *Interface) decrypt(hostinfo *HostInfo, mc uint64, out []byte, packet []
 	return out, nil
 }
 
-func (f *Interface) decryptToTun(hostinfo *HostInfo, messageCounter uint64, out []byte, packet []byte, fwPacket *firewall.Packet, nb []byte, q int, localCache firewall.ConntrackCache) bool {
+func (f *Interface) sendDecryptToTun(hostinfo *HostInfo, messageCounter uint64, out []byte, packet []byte, fwPacket *firewall.Packet, nb []byte, q int, localCache firewall.ConntrackCache) bool {
```
JackDoan, line 466: hmmmmmmmmmmm I'm not crazy about naming this `send` because it's on the rx path 😭 also it doesn't decrypt anymore

 (and we can drop `messageCounter`)

`maybeWriteToTun`?

outside.go
JackDoan, line 486: not a behavior change but I realized it's kinda funky that we won't roam on valid traffic from a tunnel that the firewall rejects. Is that on purpose?

JackDoan (approved):
This is definitely better & I'm down if you wanna merge it, but I think we can/should go further

outside.go
wadey: then we would have to special case line 75 I think, because we dont do the windowcheck on relay packets?

wadey: sorry, the windowcheck is the same, but the decryption is different and we would have to special case that

wadey: yeah I see what you are saying now, let me do that

outside.go
wadey: why "maybe"? because of the firewall?

wadey: maybe `handleOutsideMessagePacket` ? to match `handleOutsideRelayPacket`

outside.go
wadey: hmmm, I think not intentional

wadey: fixed in latest commit

outside.go
wadey: fixed to be repetitive :+1:

outside.go
wadey: the latest refactor reduced this function to one call site, so I just inlined it and its much clearer now

outside.go
nbrownus: This is going to be logged at an error level in readOutsidePackets

outside.go
```diff
@@ -568,10 +518,11 @@ func (f *Interface) decryptToTun(hostinfo *HostInfo, messageCounter uint64, out
 				"reason", dropReason,
 			)
 		}
-		return false
+		// Return true because the packet was valid, we just blocked it at the
+		// firewall
+		return true
```
nbrownus: I think this is fine just calling out that a dropped firewall packet counts as inbound for ConnectionManager now, which I think is technically correct since it decrypted fine.

outside.go
```diff
@@ -44,215 +43,198 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
...
+
+		return
+	}
+
+	out, err = f.decrypt(hostinfo, h.MessageCounter, out, packet, h, nb)
+	if err != nil {
+		hostinfo.logger(f.l).Error("Failed to decrypt packet",
+			"error", err,
+			"from", via,
+			"packet", packet,
```
nbrownus: Might be more helpful to log `"header", h` instead of packet here.

outside.go
```diff
@@ -44,215 +43,198 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
...
+		f.handshakeManager.HandleIncoming(via, packet, h)
+		return
+
+	case header.RecvError:
+		f.handleRecvError(via.UdpAddr, h)
+		return
+	}
+
+	// Relay packets are special
+	isMessageRelay := (h.Type == header.Message && h.Subtype == header.MessageRelay)
```
JackDoan, line 87: not actionable: I think it's academically interesting that this is distinct from via.IsRelayed

outside.go
```diff
@@ -34,7 +34,6 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
 		return
 	}
```
nbrownus, line 42: Would be helpful to add in a h.Version != 1 check

outside.go
wadey: hmmmm, not sure the best way to deal with that unless we make that a constant error we can compare to? or dont return an error for that case

wadey: should be a rare case because we do a window check earlier before we get to here

outside.go
```diff
@@ -44,215 +58,198 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
...
-				"error", err,
-				"from", via,
-				"packet", packet,
-			)
-			return
-		}
-
-		if h.Subtype == header.TestRequest {
+		switch h.Subtype {
+		case header.TestRequest:
```
nbrownus, line 152: `header.TestReply` is used to detect a working a tunnel via `f.connectionManager.In(hostinfo)`, without it test packets don't help keep a tunnel alive. This could be the source of e2e test failures.

outside.go
wadey: done, above

outside.go
nbrownus: We could return `ok bool` instead of `error` and log internally like we do with the counter check debug log.

outside.go
```diff
@@ -44,215 +58,188 @@ func (f *Interface) readOutsidePackets(via ViaSender, out []byte, packet []byte,
 		}
 	}
 
+	// don't keep Rx metrics for message type, since you can see those in the tun metrics
+	if h.Type != header.Message {
```
nbrownus, line 71: Currently we have a few pre metrics failures, do we want to record those as well so they don't muddy the calculated assumed header.Message stats?

wadey: hmm, the only cases above this are pretty invalid packets it looks like. We don't want to record metrics for invalid types etc. I think this okay?

wadey: Unless we want a catch all "invalid packets" counter, maybe `type` `invalid`?

wadey: oh we do have an rxUnknown counter, we could increment that: https://github.com/slackhq/nebula/blob/1b616aa9ad4822c416140f5926050aa38dbff181/message_metrics.go#L14

wadey: added rxInvalid

nbrownus (approved):

PR #1644, opened 2026-03-30
Refactor CA pool handling to use streaming

johnmaguire:
Supersedes #1556 - this is Jack's proposal with some extra test cases & error handling updated to match the original code.

Why this approach?

1. No nested loop with risk of future performance issues
2. No manual buffer management
3. PEM boundary logic is isolated and testable

AddCAFromPEM is no longer called from within Nebula, but may be called by external consumers.


pki_hup_benchmark_test.go
```diff
@@ -0,0 +1,121 @@
...
+	"time"
+
+	"github.com/slackhq/nebula/cert"
+	cert_test "github.com/slackhq/nebula/cert_test"
+	"github.com/slackhq/nebula/config"
+	"github.com/slackhq/nebula/test"
+	"github.com/stretchr/testify/require"
+)
+
+func BenchmarkReloadConfigWithCAs(b *testing.B) {
```
johnmaguire, line 20: Do we want to keep this benchmark around, or remove it?

util/pem.go
```diff
@@ -0,0 +1,58 @@
+package util
```
jrwren: Can we put the contents of this file in the cert package instead of the util package?

I'm maybe too much team https://go.dev/blog/package-names but I don't see any value in placing this single function here instead of in the cert package where it is used. It could even be unexported there.

util/pem_test.go
```diff
@@ -0,0 +1,83 @@
+package util
```
jrwren: same comment as the non-test. Move all of these test to cert package.

jrwren (changes requested):
A thought. Lemme know if I'm wrong.

util/pem.go
johnmaguire: Yep, I try to avoid packages/files named lib and util. Carried this over from the other branch and initially waffled on pem vs. cert (and considered an xpem) but left it. Moving it to cert I think makes sense.

pki_hup_benchmark_test.go
JackDoan: I dislike that it's in the root of the project, but, oh well

JackDoan (approved):

jrwren (approved):

PR #1649, opened 2026-04-05
Claude/add unsafe dns routes again

matanbaruch:
This pull request introduces support for "unsafe DNS routes," allowing the system to route traffic over Nebula to non-Nebula nodes based on DNS hostnames rather than static IPs. The implementation includes configuration, parsing, management, and periodic resolution of these DNS-based routes. The changes also include comprehensive validation and unit tests for the new feature.

Key changes:

**Unsafe DNS Routes Feature:**

* Added support for `unsafe_dns_routes` in the configuration, allowing users to specify routes using DNS hostnames, with options for periodic resolution cadence and DNS lookup timeout. Configuration is documented in `examples/config.yml`.
* Introduced a new `DnsRouteManager` in `overlay/dns_route.go` that manages the lifecycle, periodic resolution, and storage of resolved DNS routes. It updates the route table when DNS resolutions change.
* Added parsing logic for `unsafe_dns_routes` in `overlay/route.go`, including validation for hostnames, gateways, MTU, and metric fields.

**Integration and Lifecycle Management:**

* Integrated the DNS route manager into the main overlay initialization and shutdown flow (`main.go`), ensuring it is started and stopped appropriately.
* Modified `overlay/tun.go` to manage the global DNS route manager, include resolved DNS routes in the route list, and support dynamic reloads when the configuration changes. [[1]](diffhunk://#diff-ddf4df35f7f197415d05f5bf21f68f6f92a9cf5c73df157fb90c7160ae51c70eR4) [[2]](diffhunk://#diff-ddf4df35f7f197415d05f5bf21f68f6f92a9cf5c73df157fb90c7160ae51c70eR16-R17) [[3]](diffhunk://#diff-ddf4df35f7f197415d05f5bf21f68f6f92a9cf5c73df157fb90c7160ae51c70eL45-R48) [[4]](diffhunk://#diff-ddf4df35f7f197415d05f5bf21f68f6f92a9cf5c73df157fb90c7160ae51c70eR63-R99)

**Testing and Validation:**

* Added comprehensive unit tests in `overlay/dns_route_test.go` to validate parsing and error handling for the new DNS route configuration.

These changes collectively enable dynamic, DNS-based routing in Nebula, increasing flexibility and ease of configuration for environments where target IPs may change.

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -282,6 +282,33 @@ tun:
     #  metric: 100
     #  install: true
 
+  # Unsafe DNS routes allows you to route traffic over nebula to non-nebula nodes based on DNS hostnames
+  # The hostnames will be periodically resolved and routes updated automatically
+  # This is similar to unsafe_routes but uses DNS hostnames instead of static IP CIDRs
+  # NOTES:
+  # * The nebula certificate of the "via" node(s) *MUST* have the resolved IPs as subnets in its certificate
+  # * DNS resolution happens periodically based on `unsafe_dns_routes_cadence` (default: 30s)
+  # * Each resolved IP gets a /32 (IPv4) or /128 (IPv6) route automatically
+  # * This setting is reloadable
+  #
+  # unsafe_dns_routes:
+  # - host: example.com
+  #   via: 192.168.100.99
+  #   mtu: 1300
+  #   metric: 100
+  # - host: api.example.org
+  #   via:
+  #     - gateway: 10.0.0.1
+  #       weight: 10
+  #     - gateway: 10.0.0.2
+  #       weight: 5
+
+  # Cadence for DNS resolution of unsafe_dns_routes (default: 30s)
+  #unsafe_dns_routes_cadence: 30s
+
+  # Timeout for DNS lookups for unsafe_dns_routes (default: 5s)
+  #unsafe_dns_routes_lookup_timeout: 5s
+
   # On linux only, set to true to manage unsafe routes directly on the system route table with gateway routes instead of
   # in nebula configuration files. Default false, not reloadable.
   #use_system_route_table: false
--- a/main.go
+++ b/main.go
@@ -141,6 +141,14 @@ func Main(c *config.C, configTest bool, buildVersion string, logger *logrus.Logg
 				tun.Close()
 			}
 		}()
+
+		// Initialize DNS route manager for unsafe_dns_routes
+		err = overlay.InitializeDnsRouteManager(ctx, c, l, pki.getCertState().myVpnNetworks)
+		if err != nil {
+			return nil, util.ContextualizeIfNeeded("Failed to initialize DNS route manager", err)
+		}
+
+		defer overlay.CloseDnsRouteManager()
 	}
 
 	// set up our UDP listener
--- a/overlay/dns_route.go
+++ b/overlay/dns_route.go
@@ -1,0 +1,168 @@
+package overlay
+
+import (
+	"context"
+	"net"
+	"net/netip"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/slackhq/nebula/config"
+)
+
+// DnsRouteManager manages DNS-based routes that need periodic resolution
+type DnsRouteManager struct {
+	ctx            context.CancelFunc
+	l              *logrus.Logger
+	dnsRoutes      []DnsRoute
+	resolvedRoutes atomic.Pointer[[]Route]
+	cadence        time.Duration
+	lookupTimeout  time.Duration
+	mutex          sync.RWMutex
+}
+
+// NewDnsRouteManager creates a new DNS route manager
+func NewDnsRouteManager(ctx context.Context, l *logrus.Logger, c *config.C, dnsRoutes []DnsRoute) (*DnsRouteManager, error) {
+	if len(dnsRoutes) == 0 {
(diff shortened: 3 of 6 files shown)


JackDoan:
Hi @matanbaruch!

First of all, thanks for the contribution. This is a really neat idea, and solves some problems I've personally had.

That said, given how the majority of DNS traffic is still vulnerable to MITM attacks,  I fear that relying on DNS for this configuration could create some potentially unintuitive security risks for less cautious users.

At least until we can come up with a more secure way to solve this problem, I think it would be best to address this with a cronjob (or similar) outside Nebula, that queries DNS on a cadence, updates the config, and sends SIGHUP to Nebula. (& if you do this, I would strongly recommend using DoH, and sanity checking the results before you use them. For example, you'd probably want to reject any destination within your overlay network prefix(es))

JackDoan (changes requested):
see above

matanbaruch:
@JackDoan Thanks for the thoughtful review! You make a great point about the DNS MITM risk.

I did some research into how other overlay networks solved this exact problem:

- **Tailscale (App Connectors)**: DNS resolution happens on the App Connector node (routing peer), not the client. DNS queries travel over WireGuard (encrypted), so only the App Connector → upstream DNS path is exposed.
- **NetBird (DNS Interceptor)**: They actually started with client-side resolution (like my PR) and hit **exactly the security issue you described** — [netbirdio/netbird#2419](https://github.com/netbirdio/netbird/issues/2419) showed that local DNS manipulation could redirect traffic through the routing peer to arbitrary destinations. They redesigned to resolve DNS on the routing peer side, with queries forwarded over the tunnel.

Both projects converged on the same architecture: **DNS resolution must happen on a trusted node, not the client, with queries encrypted over the tunnel.**

This got me thinking — Nebula's **lighthouse** already has all the building blocks for this:

1. **Encrypted channel** — lighthouse messages travel over the Nebula tunnel (encrypted, authenticated)
2. **Request/response protocol** — protobuf-based messaging with defined types (`HostQuery`/`HostQueryReply`, etc.)
3. **Built-in DNS server** — `dns_server.go` already handles DNS resolution infrastructure
4. **Trusted admin-controlled node** — same trust model as Tailscale's App Connector

The idea would be:
- Add `DNSRouteQuery`/`DNSRouteQueryReply` message types to the lighthouse protocol
- Clients send hostname resolution requests to the lighthouse over the tunnel
- The lighthouse resolves DNS (could use DoH), validates results (reject overlay prefix IPs), and returns them encrypted
- Client installs routes from the trusted response

```
Client Node                          Lighthouse (trusted)
┌──────────┐                         ┌──────────────────┐
│ Wants to  │── DNSRouteQuery ──────►│ Resolves via     │
│ reach     │   (over Nebula tunnel) │ system DNS / DoH │
│ example.com                        │                  │
│           │◄─ DNSRouteReply ───────│ Returns resolved │
│ Installs  │   (encrypted)          │ IPs + validates  │
│ routes    │                        │ (rejects overlay │
│ for IPs   │                        │  prefix IPs)     │
└──────────┘                         └──────────────────┘
```

Code-wise, ~65% of the current PR (parsing, config, tests, integration) carries over — the main change is moving DNS resolution into a lighthouse handler (~250-300 lines of new/changed code).

Would you be open to this approach? Happy to write up a more detailed design doc if there's interest, or I can also go the external companion tool route if you'd prefer keeping this entirely out of Nebula core.

johnmaguire:
@matanbaruch Thanks for your research. We're continuing to noodle on this on our side.

To date, the Lighthouse has been explicitly kept out of the "trust model" for clients. Meaning, if a Lighthouse returns bad data (e.g. the wrong underlay IP address for a given Nebula node), this does not impact the security of the network: the node will fail to handshake due to cert mismatch and the clients will not try to handshake again (at least for a while.)

This is different from Tailscale and Netbird which use combined coordination+control servers which are responsible for doling out authentication keys to peers. In this model, the coordination server is in the trust model for clients. I fear that we can not safely defer DNS resolution to the Lighthouse given Nebula's existing security posture.

matanbaruch:
@johnmaguire Thanks, you're right, I withdraw the Lighthouse approach. With `HostQuery`, a lying Lighthouse causes a failed handshake. With DNS routes to non-Nebula hosts, there's no cert to catch a lie. Fundamentally different.

How would you feel about a **companion binary** (following the `nebula-cert` pattern) that handles this externally? It would resolve DNS via **DoH/DoT** (to address @JackDoan's MITM concern), update `tun.unsafe_routes` in the config, and SIGHUP Nebula to reload. DNS stays completely outside Nebula core, no trust model changes.

Happy to put together a PR if there's interest.

johnmaguire:
@matanbaruch The reason we want to avoid adding this to Nebula at the moment is less about code surface in the `nebula` binary in particular and more about our awareness of potential security gotchas/pitfalls here that require careful attention. Merging a `nebula-dns-routes` binary would mean we are still on the hook for ensuring the security and reliability of this system - and ultimately this isn't something we have the bandwidth to take on at this exact moment.

But, I don't think know that there is any technical need to ship the binary in this repo, alongside nebula. Therefore, I would suggest building such a companion app to your own needs, and - if you are open to it - perhaps sharing it on your own Github profile. This would allow for community experimentation of this deployment model without imposing risk on other Nebula users in the meantime.

matanbaruch:
@johnmaguire 
I guess you are right.

Thanks for the details and attention for this requirements.

I'll close this for now.

PR #1653, opened 2026-04-14
Additional e2e tests to assert current handshake behavior

nbrownus:
I am reworking handshake_ix.go into a state machine that can operate on additional message patterns. As a first pass I want to get some comprehensive handshake behavior coverage in the e2e tests so that we can assert the state machine doesn't change behavior.

There are a few remaining gaps which are difficult or near impossible to test and they are noted here.

I've run these tests a few thousand times without failure so they shouldn't be flappy.

### DIFF
--- a/e2e/handshake_manager_test.go
+++ b/e2e/handshake_manager_test.go
@@ -1,0 +1,565 @@
+//go:build e2e_testing
+// +build e2e_testing
+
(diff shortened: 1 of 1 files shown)


jrwren (approved):
github needs the its-beautiful rainbow crying unicorn

PR #1654, opened 2026-04-14
udp_linux: wrap socket operations with syscall.RawConn for clean teardown

JackDoan:
remove runtime.LockOSThread() because it makes things worse now

remove the "custom" Write() method from tun_linux.go, the stdlib path via os.File performs better

We should change our guidance around number of routines, ~2 per thread (that you wish to use for Nebula) seems to be about right now

### DIFF
--- a/udp/udp_linux.go
+++ b/udp/udp_linux.go
@@ -60,7 +60,6 @@ func NewListener(l *logrus.Logger, ip netip.Addr, port int, multi bool, batch in
 		_ = udpConn.Close()
 		return nil, err
 	}
-	udpConn.LocalAddr().String()
 
 	//steal the socket's fd for sending
 	var sockFd uintptr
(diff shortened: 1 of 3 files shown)


udp/udp_linux.go
```diff
@@ -18,58 +19,66 @@ import (
...
+	lc := net.ListenConfig{}
+	if multi {
+		lc.Control = setReusePort
 	}
-	syscall.ForkLock.RLock()
-	fd, err := unix.Socket(af, unix.SOCK_DGRAM, unix.IPPROTO_UDP)
-	if err == nil {
-		unix.CloseOnExec(fd)
+	network := "udp"
+	if ip.Is4() { //todo I am not a fan of this, would prefer a default where we always bind dual-stack unless given a non-wildcard v4 addr
```
JackDoan: @nbrownus opinion on this? I'd prefer to change the default to bind dual-stack on 0.0.0.0, but I guess we can do that separate

nbrownus (approved):

PR #1656, opened 2026-04-15
Handshake state machine

nbrownus:
Pulls noise + payload + cert handshake logic into a new `handshake` package with a unit-testable `Machine`, and rewrites the manager around it.

This is using a hand built protowire marshaller which is basically as fast as and has fewer allocations than the now deprecated gogo project. Happy to move it back to using the gogo generated code but will need to move that into its own package first. 

Deletes `handshake_ix.go`.

I did add some additional e2e tests in #1653 to help gain confidence in this change

## Follow ups

- Support for additional message patterns (XX, pqIX, pqXX, for example)
- `(eKey, dKey)` ordering is IX-shaped (preserved from master). Revisit when 3+ message patterns land.
- 3+ message patterns will need a "most recent outgoing" pointer on `HandshakeHostInfo` so we send the correct handshake bytes.
- `NebulaCipherState` can be removed.


handshake/credential.go
```diff
@@ -0,0 +1,52 @@
...
+	"crypto/rand"
+
+	"github.com/flynn/noise"
+	"github.com/slackhq/nebula/cert"
+)
+
+// Credential holds everything needed to participate in a handshake
+// at a given cert version. Version and Curve are read from Cert; the public
+// half of the static keypair likewise comes from Cert.PublicKey().
+type Credential struct {
```
JackDoan, line 13: but I wanted to rename certificates to credentials!

jk this is a good name, nice

handshake/payload.go
```diff
@@ -0,0 +1,152 @@
...
+// Wire format is protobuf-compatible with NebulaHandshake{Details: NebulaHandshakeDetails{...}}.
+type Payload struct {
+	Cert           []byte
+	InitiatorIndex uint32
+	ResponderIndex uint32
+	Time           uint64
+	CertVersion    uint32
+}
+
+// Proto field numbers for NebulaHandshakeDetails
```
JackDoan, line 25: if you want to land this, we should remove the old `NebulaHandshakeDetails` stuff too

handshake/payload.go
```diff
@@ -0,0 +1,152 @@
...
+	fieldCert           = 1 // bytes
+	fieldInitiatorIndex = 2 // uint32
+	fieldResponderIndex = 3 // uint32
+	fieldTime           = 5 // uint64
+	fieldCertVersion    = 8 // uint32
+)
+
+// MarshalPayload encodes a handshake payload in protobuf wire format compatible
+// with NebulaHandshake{Details: NebulaHandshakeDetails{...}}.
+// The result is appended to out, which may be nil.
```
JackDoan:
```suggestion
// returns out (which may be nil), with marshalled Payload appended to it
```
nit?

handshake/payload.go
```diff
@@ -0,0 +1,152 @@
...
+	out = protowire.AppendBytes(out, details)
+
+	return out
+}
+
+// UnmarshalPayload decodes a protobuf-encoded NebulaHandshake message.
+func UnmarshalPayload(b []byte) (Payload, error) {
+	var p Payload
+
+	for len(b) > 0 {
```
JackDoan, line 71: would we want to set a defensive upper-bound on `b` here?

handshake/payload.go
```diff
@@ -0,0 +1,152 @@
...
+		b = b[n:]
+
+		switch {
+		case num == 1 && typ == protowire.BytesType:
+			details, n := protowire.ConsumeBytes(b)
+			if n < 0 {
+				return p, errInvalidHandshakeMessage
+			}
+			b = b[n:]
+			if err := unmarshalPayloadDetails(&p, details); err != nil {
```
JackDoan, line 85: from a security POV, I don't like that this function could be called multiple times per-`b`, and still mutate the same `p`. I'm not smart enough to describe the bugs this could cause off the top of my head, but it makes me nervous.

JackDoan: I'm pretty sure some proto libs might permit this, but this feels like a good opportunity to stop allowing that.

handshake/payload.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		b = b[n:]
+
+		switch {
+		case num == fieldCert && typ == protowire.BytesType:
+			v, n := protowire.ConsumeBytes(b)
+			if n < 0 {
+				return errInvalidHandshakeDetails
+			}
+			p.Cert = append([]byte(nil), v...)
```
JackDoan, line 122: copying this out of the message is good, but it might be nice to be able to recycle this buffer?

If we do, we must make sure to set its len to zero at the top of this function

handshake/payload.go
```diff
@@ -0,0 +1,152 @@
...
+			p.Time = v
+			b = b[n:]
+		case num == fieldCertVersion && typ == protowire.VarintType:
+			v, n := protowire.ConsumeVarint(b)
+			if n < 0 {
+				return errInvalidHandshakeDetails
+			}
+			p.CertVersion = uint32(v)
+			b = b[n:]
+		default:
```
JackDoan, line 164: we don't error out if we get an "expected" tag of the wrong type. Not a bad thing, but I don't recall if that's what proto does.

handshake/payload.go
nbrownus: We do by the side affect of being in noise, 64k is the max for a valid noise message and we couldn't get here without having a valid noise message.

handshake/payload.go
nbrownus: It's documented protobuf behavior https://protobuf.dev/programming-guides/encoding/#last-one-wins

JackDoan: well ok

JackDoan (dismissed):

handshake/payload.go
JackDoan: deferred

JackDoan (dismissed):
niiiiice

JackDoan (dismissed):

JackDoan (dismissed):

handshake/machine_test.go
```diff
@@ -0,0 +1,650 @@
...
+
+func TestMachineAESCipher(t *testing.T) {
+	ca, _, caKey, _ := ct.NewTestCaCert(
+		cert.Version2, cert.Curve_CURVE25519, time.Time{}, time.Time{}, nil, nil, nil,
+	)
+	caPool := ct.NewTestCAPool(ca)
+	// AES cipher test requires building the handshake state with AES, which
+	// is done by the manager in production. Our test helper always uses
+	// chachapoly. This test verifies the Machine works regardless of cipher
+	// by running a standard handshake (chachapoly).
```
wadey: this confuses me, so is this test actually testing AES cipher or not? If not, why is it called that?

JackDoan (dismissed):

wadey (dismissed):
LGTM, its a big diff but I think clearer than what existed before! one question about one of the comments

wadey (approved):

PR #1375, opened 2025-04-02
Remove more os.Exit calls and give a more reliable wait for stop function

nbrownus:

PR #1658, opened 2026-04-15
Remove more os.Exit calls and give a more reliable wait for stop function (attempt 2)

JackDoan:
replaces #1375

### DIFF
--- a/overlay/tun_linux.go
+++ b/overlay/tun_linux.go
@@ -686,15 +686,24 @@ func (t *tun) updateRoutes(r netlink.RouteUpdate) {
 func (t *tun) Close() error {
 	if t.routeChan != nil {
 		close(t.routeChan)
-	}
-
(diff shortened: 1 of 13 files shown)


overlay/tun_linux.go
```diff
@@ -85,7 +84,7 @@ func newTunFromFd(c *config.C, l *logrus.Logger, deviceFd int, vpnNetworks []net
 }
 
 func newTun(c *config.C, l *logrus.Logger, vpnNetworks []netip.Prefix, multiqueue bool) (*tun, error) {
-	fd, err := unix.Open("/dev/net/tun", os.O_RDWR, 0)
+	fd, err := unix.Open("/dev/net/tun", os.O_RDWR|unix.O_NONBLOCK, 0)
```
JackDoan, line 87: smoke seems to want this, benchmarking now

PR #1660, opened 2026-04-15
tun_linux.go: stdlib too slow, but can't use blocking IO and clean shutdown

JackDoan:


overlay/tun_linux.go
```diff
@@ -24,9 +25,61 @@ import (
...
+			{Fd: int32(shutdownFd), Events: unix.POLLIN},
+		},
+	}
+}
+
+func (r *tunFd) Read(buf []byte) (int, error) {
+	for {
+		n, err := unix.Read(r.fd, buf)
+		if err != nil {
+			if err == unix.EAGAIN {
```
nbrownus: We can get EINTR and EBADF here as well.

EINTR - continue
EBADF - I'd think turning it into os.ErrClosed is fine, keeps the logs clean from spam

overlay/tun_linux.go
nbrownus, line 811: Not sure if you want to handle it in this PR but we aren't actually closing anything beyond the primary reader, interface.go needs to call it for each reader or we track the ancestry and chain the closes.

overlay/tun_linux.go
```diff
@@ -24,9 +25,61 @@ import (
...
+				continue
+			}
+			return 0, err
+		}
+		return n, nil
+	}
+}
+
+func (r *tunFd) Write(buf []byte) (int, error) {
+	return unix.Write(r.fd, buf)
```
nbrownus, line 115: - EBADF (convert to os.ErrClosed)
- EAGAIN (retry probably)
- EINTR (retry)

overlay/tun_linux.go
```diff
@@ -145,8 +207,7 @@ func newTunGeneric(c *config.C, l *logrus.Logger, file *os.File, vpnNetworks []n
 		l:                         l,
 	}
 
-	err := t.reload(c, true)
-	if err != nil {
+	if err := t.reload(c, true); err != nil {
```
nbrownus: We should close our fds here as well, happy to handle in another PR. Now that we are trying to properly clean up instead of relying on the OS to do it, we have some fds to close.

PR #1661, opened 2026-04-16
Remove more os.Exit calls and give a more reliable wait for stop function (attempt 3)

JackDoan:
replaces #1375, #1658 

notes from Slack because I'm a mongrel who provided no description

The broad strokes summary is Nebula was using blocking IO on the tun file

This is great for performance, and not a problem when you get restarted by the init system murdering your process

But if you want to restart in-process, you need to unblock those threads somehow. And if they're blocking reads, you can only do that (on linux) by signalling their thread ID (which we don't have) or by feeding them something to read (which we agreed was too gross, because they're tuns, we'd have to open a socket and send a packet to nowhere)

So, we made the tun non-blocking, which comes with other compromises wrt performance. The one we landed on was to do a blocking read on a poll, which lets us wait for one or more file-descriptors to be "ready to read".  When we get a packet, the thread wakes and reads the tun fd, and when we want to shutdown, we write 0x1 to the "shutdown fd", which unblocks the reader threads and tells them its time to pack up and go home

Why didn't you just use os.File which would let you use go's netpoller? Because it benchmarked noticeably worse than what we ended up doing. It has locks and is thread safe, we don't and aren't.

### DIFF
--- a/cmd/nebula-service/main.go
+++ b/cmd/nebula-service/main.go
@@ -78,8 +78,16 @@ func main() {
 	}
 
 	if !*configTest {
-		ctrl.Start()
-		ctrl.ShutdownBlock()
+		wait, err := ctrl.Start()
+		if err != nil {
+			util.LogWithContextIfNeeded("Error while running", err, l)
+			os.Exit(1)
+		}
+
+		go ctrl.ShutdownBlock()
+		wait()
+
+		l.Info("Goodbye")
 	}
 
 	os.Exit(0)
--- a/cmd/nebula-service/main.go
+++ b/cmd/nebula-service/main.go
@@ -85,7 +85,12 @@ func main() {
 		}
 
 		go ctrl.ShutdownBlock()
-		wait()
+
+		if err := wait(); err != nil {
+			l.WithError(err).Error("Nebula stopped due to fatal error")
+			l.Info("Goodbye")
+			os.Exit(2)
+		}
 
 		l.Info("Goodbye")
 	}
--- a/cmd/nebula-service/main.go
+++ b/cmd/nebula-service/main.go
@@ -88,7 +88,6 @@ func main() {
 
 		if err := wait(); err != nil {
 			l.WithError(err).Error("Nebula stopped due to fatal error")
-			l.Info("Goodbye")
 			os.Exit(2)
 		}
 
--- a/cmd/nebula/main.go
+++ b/cmd/nebula/main.go
@@ -72,9 +72,17 @@ func main() {
 	}
 
 	if !*configTest {
-		ctrl.Start()
+		wait, err := ctrl.Start()
+		if err != nil {
+			util.LogWithContextIfNeeded("Error while running", err, l)
+			os.Exit(1)
+		}
+
+		go ctrl.ShutdownBlock()
 		notifyReady(l)
-		ctrl.ShutdownBlock()
+		wait()
+
+		l.Info("Goodbye")
 	}
 
 	os.Exit(0)
--- a/cmd/nebula/main.go
+++ b/cmd/nebula/main.go
@@ -83,7 +83,6 @@ func main() {
 
 		if err := wait(); err != nil {
 			l.WithError(err).Error("Nebula stopped due to fatal error")
-			l.Info("Goodbye")
 			os.Exit(2)
 		}
 
--- a/cmd/nebula/main.go
+++ b/cmd/nebula/main.go
@@ -80,7 +80,12 @@ func main() {
 
 		go ctrl.ShutdownBlock()
 		notifyReady(l)
-		wait()
+
+		if err := wait(); err != nil {
+			l.WithError(err).Error("Nebula stopped due to fatal error")
+			l.Info("Goodbye")
+			os.Exit(2)
+		}
 
 		l.Info("Goodbye")
 	}
--- a/control.go
+++ b/control.go
@@ -2,9 +2,11 @@ package nebula
 
 import (
 	"context"
+	"errors"
 	"net/netip"
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
 
 	"github.com/sirupsen/logrus"
@@ -12,6 +14,16 @@ import (
 	"github.com/slackhq/nebula/header"
 	"github.com/slackhq/nebula/overlay"
 )
+
+type RunState int
+
+const (
+	Stopped  RunState = 0 // The control has yet to be started
+	Started  RunState = 1 // The control has been started
+	Stopping RunState = 2 // The control is stopping
+)
+
+var ErrAlreadyStarted = errors.New("nebula is already started")
 
 // Every interaction here needs to take extra care to copy memory and not return or use arguments "as is" when touching
 // core. This means copying IP objects, slices, de-referencing pointers and taking the actual value, etc
@@ -26,6 +38,9 @@ type controlHostLister interface {
 }
 
 type Control struct {
+	stateLock sync.Mutex
+	state     RunState
+
 	f                      *Interface
 	l                      *logrus.Logger
 	ctx                    context.Context
@@ -49,10 +64,21 @@ type ControlHostInfo struct {
 	CurrentRelaysThroughMe []netip.Addr     `json:"currentRelaysThroughMe"`
 }
 
-// Start actually runs nebula, this is a nonblocking call. To block use Control.ShutdownBlock()
-func (c *Control) Start() {
+// Start actually runs nebula, this is a nonblocking call.
+// The returned function can be used to wait for nebula to fully stop.
+func (c *Control) Start() (func(), error) {
+	c.stateLock.Lock()
+	if c.state != Stopped {
+		c.stateLock.Unlock()
+		return nil, ErrAlreadyStarted
+	}
+
 	// Activate the interface
-	c.f.activate()
+	err := c.f.activate()
+	if err != nil {
+		c.stateLock.Unlock()
+		return nil, err
+	}
 
 	// Call all the delayed funcs that waited patiently for the interface to be created.
 	if c.sshStart != nil {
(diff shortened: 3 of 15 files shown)


udp/udp_generic.go
```diff
@@ -73,25 +71,14 @@ type rawMessage struct {
...
 	for {
 		// Just read one packet at a time
 		n, rua, err := u.ReadFromUDPAddrPort(buffer)
 		if err != nil {
-			if errors.Is(err, net.ErrClosed) {
-				u.l.WithError(err).Debug("udp socket is closed, exiting read loop")
-				return
-			}
-			// Dampen unexpected message warns to once per minute
-			if lastRecvErr.IsZero() || time.Since(lastRecvErr) > time.Minute {
```
JackDoan, line 90: @nbrownus did you remove this on purpose?

control.go
```diff
@@ -71,16 +100,36 @@ func (c *Control) Start() {
 		c.lighthouseStart()
 	}
 
+	c.f.triggerShutdown = c.Stop
+
 	// Start reading packets.
-	c.f.run()
+	c.state = Started
+	c.stateLock.Unlock()
+	return c.f.run()
```
jrwren:
```suggestion
	return c.f.run(), err
```

interface.go
```diff
@@ -237,27 +244,53 @@ func (f *Interface) activate() {
...
 		f.inside.Close()
-		f.l.Fatal(err)
+		return err
 	}
+
+	return nil
 }
 
-func (f *Interface) run() {
+func (f *Interface) run() (func() error, error) {
```
jrwren, line 263:
```suggestion
func (f *Interface) run() (func() error) {
```

interface.go
```diff
@@ -237,27 +244,53 @@ func (f *Interface) activate() {
...
+		})
+	}
+
+	return func() error {
+		f.wg.Wait()
+		if e := f.fatalErr.Load(); e != nil {
+			return *e
+		}
+		return nil
+	}, nil
```
jrwren, line 284:
```suggestion
	}
```

jrwren (commented):
This is my nit request.

Please let me know if I'm wrong.

interface.go
JackDoan: this doesn't return any errors _today_ but it felt short-sighted to exclude the possibility. We're bad at it, but we would like our external interfaces to be stable!

jrwren: that makes sense. thanks for considering and explaining

overlay/tun_linux.go
```diff
@@ -24,9 +25,173 @@ import (
...
+		return os.ErrClosed
+	} else if tunEvents&problemFlags != 0 {
+		return os.ErrClosed
+	}
+	return nil
+}
+
+func (r *tunFile) Read(buf []byte) (int, error) {
+	for {
+		if n, err := unix.Read(r.fd, buf); err == nil {
```
nbrownus, line 139: unix.Read can EINTR as well. Should retry

interface.go
```diff
@@ -237,27 +244,53 @@ func (f *Interface) activate() {
...
+				return err
 			}
 		}
 		f.readers[i] = reader
 	}
 
-	if err := f.inside.Activate(); err != nil {
+	f.wg.Add(1) // for us to wait on Close() to return
+	if err = f.inside.Activate(); err != nil {
 		f.inside.Close()
```
nbrownus, line 256: need to `f.wg.Done()` on error

overlay/tun_linux.go
JackDoan: ah yeah that makes sense, I was avoiding it bc I was thinking getting a signal means we should bail, but that's not always true, and we're nb now

control.go
```diff
@@ -49,10 +64,24 @@ type ControlHostInfo struct {
...
 
-// Start actually runs nebula, this is a nonblocking call. To block use Control.ShutdownBlock()
-func (c *Control) Start() {
+// Start actually runs nebula, this is a nonblocking call.
+// The returned function blocks until nebula has fully stopped and returns the
+// first fatal reader error (if any). A nil error means nebula shut down
+// gracefully; a non-nil error means a reader hit an unexpected failure that
+// triggered the shutdown.
+func (c *Control) Start() (func() error, error) {
+	c.stateLock.Lock()
```
nbrownus, line 77: We could avoid holding a lock for the entire func by transition to `Starting`, similar to how `Stop()` works. Only real impact I see would be avoiding deadlocks in case we sprinkle other `c.stateLock()` calls around in functions called by `Start()`

control.go
```diff
@@ -49,10 +64,24 @@ type ControlHostInfo struct {
...
 }
 
-// Start actually runs nebula, this is a nonblocking call. To block use Control.ShutdownBlock()
-func (c *Control) Start() {
+// Start actually runs nebula, this is a nonblocking call.
+// The returned function blocks until nebula has fully stopped and returns the
+// first fatal reader error (if any). A nil error means nebula shut down
+// gracefully; a non-nil error means a reader hit an unexpected failure that
+// triggered the shutdown.
+func (c *Control) Start() (func() error, error) {
```
nbrownus, line 76: One thing I hedged on in the first pass through was how to deal with the `wait()` func.

Go has like 3 main patterns here, which this isn't really one (it was my hedge on making a choice while I tested)

`os/exec.Cmd` - non blocking `Start()`, blocking `Wait()` on the struct.
`http.Server` - blocking `ListenAndServe()`/`Serve()`, blocking `Shutdown()`, non blocking `Close()`
`Run(ctx) error` - blocks and returns an error

What we have:
- non blocking `Start()`
- non blocking `Stop()`
- an awkward `wait()` func you gotta deal with but you can do it however you want.

We leave it entirely up to the caller, which has some nice properties, but it's also kind of awkward

nbrownus: I think the hedge on the original hedge would be to provide a `Wait() error` on `Control` so that callers don't have to thread `wait` around alongside the Control ref.

control.go
JackDoan: you're not wrong, but I liked that this made it very simple to reason about "what happens if someone calls start then stop really fast, perhaps even before start returns?"

control.go
JackDoan: we could add control.Wait() but the counterargument I think is it's another potential _thing_ for people to get wrong. By returning a wait, we force you to let Start() finish.

udp/udp_generic.go
```diff
@@ -84,7 +84,7 @@ func (u *GenericConn) ListenOut(r EncReader) {
 		if err != nil {
 			if errors.Is(err, net.ErrClosed) {
 				u.l.WithError(err).Debug("udp socket is closed, exiting read loop")
```
nbrownus: We can let the caller log like with darwin

interface.go
```diff
@@ -276,9 +310,16 @@ func (f *Interface) listenOut(i int) {
...
+	err := li.ListenOut(func(fromUdpAddr netip.AddrPort, payload []byte) {
 		f.readOutsidePackets(ViaSender{UdpAddr: fromUdpAddr}, plaintext[:0], payload, h, fwPacket, lhh, nb, i, ctCache.Get(f.l))
 	})
+
+	if err != nil && !f.closed.Load() {
+		f.l.WithError(err).Error("Error while reading inbound packet, closing")
+		f.onFatal(err)
+	}
+
+	f.l.Infof("underlay reader %v is done", i)
```
nbrownus: Seems like more of a debug log line

interface.go
```diff
@@ -292,17 +333,17 @@ func (f *Interface) listenIn(reader io.ReadWriteCloser, i int) {
...
-			f.l.WithError(err).Error("Error while reading outbound packet")
-			// This only seems to happen when something fatal happens to the fd, so exit.
-			os.Exit(2)
+			break
 		}
 
 		f.consumeInsidePacket(packet[:n], fwPacket, nb, out, i, conntrackCache.Get(f.l))
 	}
+
+	f.l.Infof("overlay reader %v is done", i)
```
nbrownus: Another debug line candidate

interface.go
```diff
@@ -477,23 +518,19 @@ func (f *Interface) GetCertState() *CertState {
...
-		if err := r.Close(); err != nil {
-			f.l.WithError(err).Error("Error while closing tun reader")
+			f.l.WithError(err).WithField("writer", i).Error("Error while closing udp socket")
 		}
 	}
 
-	// Release the tun device
-	return f.inside.Close()
+	// Release the tun device (closing the tun also closes all readers)
+	err = f.inside.Close()
```
nbrownus: This will overwrite any errors we encounter closing the individual sockets and may lead to returning nil even though there were errors

cmd/nebula-service/main.go
```diff
@@ -78,8 +78,21 @@ func main() {
...
+		if err != nil {
+			util.LogWithContextIfNeeded("Error while running", err, l)
+			os.Exit(1)
+		}
+
+		go ctrl.ShutdownBlock()
+
+		if err := wait(); err != nil {
+			l.WithError(err).Error("Nebula stopped due to fatal error")
+			l.Info("Goodbye")
```
nbrownus: Thoughts on dropping goodbye?

cmd/nebula/main.go
```diff
@@ -72,9 +72,22 @@ func main() {
...
+			os.Exit(1)
+		}
+
+		go ctrl.ShutdownBlock()
 		notifyReady(l)
-		ctrl.ShutdownBlock()
+
+		if err := wait(); err != nil {
+			l.WithError(err).Error("Nebula stopped due to fatal error")
+			l.Info("Goodbye")
```
nbrownus: Thoughts on dropping goodbye?

cmd/nebula-service/main.go
JackDoan: oh I kept it on purpose lol

control.go
```diff
@@ -49,10 +67,30 @@ type ControlHostInfo struct {
...
+	defer c.stateLock.Unlock()
+	switch c.state {
+	case StateReady:
+		//yay!
+	case StateStopped, StateStopping:
+		return nil, ErrAlreadyStopped
+	case StateStarted:
+		return nil, ErrAlreadyStarted
+	default:
+		return nil, errors.New("invalid state")
```
nbrownus: Could be static like its friends above

control.go
```diff
@@ -49,10 +67,30 @@ type ControlHostInfo struct {
...
+	}
+
 	// Activate the interface
-	c.f.activate()
+	err := c.f.activate()
+	if err != nil {
+		return nil, err
```
nbrownus, line 94: If this fails we get left in a strange state, we use `c.state = StateStopped` down below

nbrownus (approved):

PR #1666, opened 2026-04-20
FreeBSD tun needs to be non blocking as well

nbrownus:
Smokes appear to be timing out due to shutdown being blocked

### DIFF
--- a/overlay/tun_freebsd.go
+++ b/overlay/tun_freebsd.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"io/fs"
 	"net/netip"
+	"os"
 	"sync/atomic"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


overlay/tun_freebsd.go
```diff
@@ -93,86 +94,107 @@ type tun struct {
...
-		err := syscall.Close(t.devFd)
-		if err != nil {
+	if t.f != nil {
+		// Closing the *os.File unblocks any readers via the runtime poller
+		if err := t.f.Close(); err != nil {
 			t.l.WithError(err).Error("Error closing device")
 		}
-		t.devFd = -1
+		// t.f.Close should have handled it for us but let's be extra sure
+		_ = unix.Close(t.fd)
```
JackDoan: why not trust os.File?

JackDoan (dismissed):

overlay/tun_freebsd.go
```diff
@@ -93,107 +94,184 @@ type tun struct {
...
+	closed    atomic.Bool
 }
 
-func (t *tun) Read(to []byte) (int, error) {
-	// use readv() to read from the tunnel device, to eliminate the need for copying the buffer
-	if t.devFd < 0 {
-		return -1, syscall.EINVAL
+// blockOnRead waits until the tun fd is readable or shutdown has been signaled.
+// Returns os.ErrClosed if Close was called.
+func (t *tun) blockOnRead() error {
```
JackDoan, line 108: it would be neat to not have to copy/paste this

overlay/tun_freebsd.go
```diff
@@ -266,21 +364,32 @@ func newTun(c *config.C, l *logrus.Logger, vpnNetworks []netip.Prefix, _ bool) (
...
+		fd:          fd,
+		shutdownR:   shutdownR,
+		shutdownW:   shutdownW,
+		readPoll: [2]unix.PollFd{
+			{Fd: int32(fd), Events: unix.POLLIN},
+			{Fd: int32(shutdownR), Events: unix.POLLIN},
+		},
+		writePoll: [2]unix.PollFd{
+			{Fd: int32(fd), Events: unix.POLLOUT},
+			{Fd: int32(shutdownR), Events: unix.POLLIN},
```
JackDoan, line 384:
```suggestion
			{Fd: int32(shutdownW), Events: unix.POLLIN},
```

was this on purpose?

JackDoan (approved):
assuming shutdownW vs shutdownR wasn't a mistake

overlay/tun_freebsd.go
nbrownus: Yeah, polling on shutdownW doesn't behave in the same way as shutdownR and we want to bail when shutdownR has been closed. `unix.POLLOUT` on the tun fd is going to give us the "ok to write" signal, shutdownR will give us an EOF when the write side of the pipe has been closed.

overlay/tun_freebsd.go
nbrownus: Will try and resolve in a follow up

PR #1667, opened 2026-04-20
No more dns globals, proper cleanup on shutdown

nbrownus:

### DIFF
--- a/dns_server.go
+++ b/dns_server.go
@@ -28,9 +28,10 @@ type dnsServer struct {
 	mux *dns.ServeMux
 
 	// enabled mirrors `lighthouse.serve_dns && lighthouse.am_lighthouse`.
-	// Start and reload consult it so callers don't need to know the gating
-	// rules; Add always populates so entries are ready if DNS is later
-	// enabled via SIGHUP.
+	// Start, Add, and reload consult it so callers don't need to know the
+	// gating rules. When it toggles off via reload, accumulated records are
+	// cleared so a later re-enable starts with a fresh map populated from
+	// new handshakes.
 	enabled atomic.Bool
 
 	serverMu sync.Mutex
@@ -111,6 +112,9 @@ func (d *dnsServer) reload(c *config.C, initial bool) error {
 		if running != nil {
 			d.Stop()
 		}
+		// Drop any records that accumulated while enabled; a later re-enable
+		// will repopulate from fresh handshakes.
+		d.clearRecords()
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


dns_server.go
```diff
@@ -187,24 +186,118 @@ func getDnsServerAddr(c *config.C) string {
...
+	go func() {
+		<-ctx.Done()
+		d.shutdown("shutdown")
+	}()
+
+	for ctx.Err() == nil {
+		if !d.serveWithRetry(ctx, c, maxAttempts, retryWait) {
+			return
+		}
+		// serveWithRetry parked and was woken by a reload — loop to rebind.
```
JackDoan:
```suggestion
		// serveWithRetry parked and was woken by a reload - loop to rebind.
```

I'm a boomer who doesn't like non-ascii in source code 👴

JackDoan: super nit tho, I just wish clod was cooler

JackDoan (dismissed):
kind of funny that this isn't the pattern we used on Control but oh well, still better

dns_server.go
nbrownus: Yeah it ignored my "never use emdash" instructions

dns_server.go
```diff
@@ -1,42 +1,56 @@
...
+
+	mux *dns.ServeMux
+
+	serverMu sync.Mutex
+	server   *dns.Server
+	addr     string
+
+	// wake is used by reload to unpark Run after it has exhausted its retry
+	// budget. Buffered so the callback stays non-blocking and match-depth-1
+	// drops coalesce. We only care that *a* reload happened, not how many.
```
JackDoan, line 35: I'm not smart enough for this comment

dns_server.go
```diff
@@ -187,24 +186,118 @@ func getDnsServerAddr(c *config.C) string {
...
+			return
+		}
+		// serveWithRetry parked and was woken by a reload, loop to rebind.
 	}
 }
 
-func reloadDns(l *logrus.Logger, c *config.C) {
-	if dnsAddr == getDnsServerAddr(c) {
-		l.Debug("No DNS server config change detected")
+// serveWithRetry runs up to maxAttempts bind-and-serve cycles. It returns
```
JackDoan, line 217: we didn't retry before, why now?

JackDoan (dismissed):

dns_server.go
```diff
@@ -1,45 +1,212 @@
...
+	d.serverMu.Lock()
+	running := d.server
+	runningStarted := d.started
+	sameAddr := d.addr == newAddr
+	d.addr = newAddr
+	d.enabled.Store(enabled)
+	d.serverMu.Unlock()
+
+	if initial && wantsDns && !amLighthouse {
+		d.l.Warn("DNS server refusing to run because this host is not a lighthouse.")
```
JackDoan: this feels like it wants to early return

dns_server.go
```diff
@@ -1,45 +1,212 @@
...
+		close(started)
+	}
+
+	if err != nil {
+		d.l.WithError(err).Warn("Failed to run the DNS responder")
+	}
+}
+
+// Stop shuts down the active server, if any. Idempotent.
+func (d *dnsServer) Stop() {
```
JackDoan, line 209: you could argue that stopping+starting the server should trash the list of records, but I can also defend not doing that

dns_server.go
```diff
@@ -1,45 +1,212 @@
...
+	ds.mux = dns.NewServeMux()
+	ds.mux.HandleFunc(".", ds.handleDnsRequest)
+
+	c.RegisterReloadCallback(func(c *config.C) {
+		if err := ds.reload(c, false); err != nil {
+			l.WithError(err).Error("Failed to reload DNS responder from config")
+		}
+	})
+
+	go func() {
```
JackDoan: I feel like this shouldn't be kicked off until .Start is called

JackDoan (approved):

PR #1670, opened 2026-04-21
Make stats a server we can reconfigure and start/stop

nbrownus:
Similar story as #1667 

Follow up PR can swap from using the default registry

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -1,14 +1,16 @@
 package nebula
 
 import (
+	"context"
 	"errors"
 	"fmt"
-	"log"
 	"log/slog"
 	"net"
 	"net/http"
 	"runtime"
 	"strconv"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	graphite "github.com/cyberdelia/go-metrics-graphite"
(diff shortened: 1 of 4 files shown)


stats.go
```diff
@@ -19,109 +21,314 @@ import (
...
+}
+
+// statsConfig is the snapshot of stats-related config that drives the runtime.
+// It is comparable with == so reload can detect "no change" cheaply.
+type statsConfig struct {
+	typ      string
+	interval time.Duration
+
+	// graphite
+	protocol string
```
JackDoan, line 62: it really sucks that this is different from prom's version. Jamming both into sub-structs would make this a lot clearer further down though

stats.go
```diff
@@ -19,109 +21,314 @@ import (
...
-	interval := c.GetDuration("stats.interval", 0)
-	if interval == 0 {
-		return nil, fmt.Errorf("stats.interval was an invalid duration: %s", c.GetString("stats.interval", ""))
+	c.RegisterReloadCallback(func(c *config.C) {
+		if err := s.reload(c, false); err != nil {
+			l.WithError(err).Error("Failed to reload stats from config")
+		}
+	})
+
+	go func() {
```
JackDoan: calling this before start feels like a mistake, vis a vis dns

stats.go
```diff
@@ -19,109 +21,314 @@ import (
...
+	s.runMu.Unlock()
+
+	if listener == nil {
+		return
 	}
 
-	if !configTest {
-		l.Infof("Starting graphite. Interval: %s, prefix: %s, addr: %s", i, prefix, addr)
-		go graphite.Graphite(metrics.DefaultRegistry, i, prefix, addr)
+	s.l.WithField("addr", listener.Addr).Info("Starting prometheus stats listener")
```
JackDoan: bleh it's so weird that this is prom-only

stats.go
JackDoan: also I think it makes a goroutine even if stats is disabled, which is _fine_ but aesthetically unpleasant

wadey (approved):

PR #1672, opened 2026-04-22
Switch to slog, remove logrus

nbrownus:
API break for embedders: `nebula.Main`, `config.NewC`, and every `New*FromConfig` subsystem constructor take `*slog.Logger`.

Visible output changes:
- level=info becomes level=INFO (every log line).
- Trace level prints as level=DEBUG-4.
- `logging.timestamp_format` is ignored; slog emits RFC3339Nano.
- JSON attr order changed from sorted to insertion order.
- Text quoting is sparser (slog omits quotes around :, /, [] when not required).
- Handful of message strings compressed, e.g. "listening on X" -> "listening addr=X".

## Open questions

- Should we drop the trace log level? We use it in a handful of places, all related to allow listing local/remote ip addresses.
	- Logs as `"level":"DEBUG-4"`
	- Requires custom code on our side to thread it through
- Is anyone bothered by the level case change (`info` -> `INFO`)? We could implement our own handler to write lowercase names like logrus did.
- slog doesn't have a simple way to manipulate the time log format. We can disable time but we no longer respect `logging.timestamp_format`.
- Supporting changing the log format live backs us into a corner: `.With()` pays 2x stock at subsystem creation (not per log). If we require a restart for this then we can avoid the shenanigans.


JackDoan (dismissed):

JackDoan (dismissed):

JackDoan (dismissed):

JackDoan (approved):

wadey (approved):
Long overdue, LGTM!

- `timestamp_format` was only added so we can could more accuracy down to microseconds. It looks like the new format has good accuracy by default so it should be fine to drop this config option.

PR #1679, opened 2026-04-22
Connection manager and related move to use slog

nbrownus:
Continuation of https://github.com/slackhq/nebula/pull/1677

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -5,17 +5,16 @@ import (
 	"context"
 	"encoding/binary"
 	"fmt"
+	"log/slog"
 	"net/netip"
 	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/rcrowley/go-metrics"
-	"github.com/sirupsen/logrus"
 	"github.com/slackhq/nebula/cert"
 	"github.com/slackhq/nebula/config"
 	"github.com/slackhq/nebula/header"
-	"github.com/slackhq/nebula/logbridge"
 )
 
 type trafficDecision int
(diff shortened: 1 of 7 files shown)


JackDoan (approved):
```
./relay_manager.go:223:41: resp.InitiatorRelayIndex escapes to heap in (*relayManager).handleCreateRelayResponse:
./relay_manager.go:223:41:   flow: slog.value ← &{storage for resp.InitiatorRelayIndex}:
./relay_manager.go:223:41:     from resp.InitiatorRelayIndex (spill) at ./relay_manager.go:223:41
./relay_manager.go:223:41:     from slog.key, slog.value := "initiatorRelayIndex", resp.InitiatorRelayIndex (assign-pair) at ./relay_manager.go:223:13
./relay_manager.go:223:41:   flow: {temp} ← slog.value:
./relay_manager.go:223:41:     from slog.AnyValue(slog.value) (call parameter) at ./relay_manager.go:223:13
./relay_manager.go:223:41:   flow: ~r0 ← {temp}:
./relay_manager.go:223:41:     from slog.Attr{...} (struct literal element) at ./relay_manager.go:223:13
./relay_manager.go:223:41:     from ~r0 = slog.Attr{...} (assign-pair) at ./relay_manager.go:223:13
./relay_manager.go:223:41:   flow: {storage for ~r0} ← ~r0:
./relay_manager.go:223:41:     from ~r0 (interface-converted) at ./relay_manager.go:223:13
```

this is gonna be better than logrus right? :D

nbrownus:
It should but I gotta get through this transitional period first. We have some knobs we can turn as well if things are more heapy than logrus (not sure how that'd happen)

PR #1687, opened 2026-04-23
Slog finale

nbrownus:
Finish the move to slog, I expect some additional changes as part of final review, happy to make any here as well.


logger.go
```diff
@@ -242,7 +179,8 @@ func parseLogLevel(s string) (slog.Level, error) {
 	case "error":
 		return slog.LevelError, nil
 	case "fatal", "panic":
-		// logrus had these; slog collapses them into error.
+		// Accepted for backwards compatibility with older configs. slog has
```
JackDoan, line 182: Hmm. I feel like we should whine at the user here

nbrownus: We could but I would want to use a `Warn` level log which we'd be disabling :trollface:

JackDoan (approved):

PR #1691, opened 2026-04-24
noise: only type-assert once

JackDoan:
In some of my performance experimentation work, this type assertion was actually showing up in the pprof. It saves us a tiny amount of CPU to do it when we make the struct, rather than per-packet.


nbrownus (approved):
Ran into this in the handshake state machine refactor, looks like a great candidate for removal in that or another PR.

PR #1695, opened 2026-04-24
overlay/tio: KeepAlive writev iovec + payloads through syscall

rawdigits:
Obsolete: Jack landed equivalent or better fixes for both pieces independently:

- `8fd724d` ("fix?"): same hdr_len trust fix this PR's first commit had.
- `c9d5a6e` ("be safer"): cleaner structural fix to `rawWrite` — converted the `unsafe.Pointer` → `uintptr` pattern to a proper `[]unix.Iovec` slice, which lets the compiler keep the iovec rooted automatically without `runtime.KeepAlive`.
- `334dd7a` ("NAPI broken on old kernels and also not helpful"): removes `IFF_NAPI` from the TUN open. This is the actual fix for the kernel UAF I reported in PR comments — `IFF_NAPI` causes `tun_napi_receive` to call `napi_gro_receive` instead of `netif_receive_skb`, which doesn't orphan the skb. On Linux 4.19 (UniFi UXG-Pro), the skb retains TUN-socket ownership through forwarding into a local TCP socket; `tcp_recvmsg` then calls `sock_wfree` against the TUN socket, underflowing `sk_wmem_alloc`. The `kfree_skb` + `tcp_recvmsg` paired warnings I posted earlier all originate from this.

All three bugs reproduced/verified on a UXG-Pro under iperf3 -R; box is now stable on `better-tun-interface@334dd7a` with full GSO/GRO + TSO write-side coalescing.

Closing as superseded. Thanks Jack.

Not filing a separate PR for the IFF_NAPI removal since you've already done it; the analysis is in PR #1695 comments if anyone needs it later.

### DIFF
--- a/overlay/tio/tio_gso_linux.go
+++ b/overlay/tio/tio_gso_linux.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"sync/atomic"
 	"syscall"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


rawdigits:
Update: on the same UXG-Pro (Linux 4.19.152-ui-alpine), after installing a dnclient with *both* commits in this PR applied, the same trace reproduces under `iperf3 -R`:

```
refcount_t: underflow; use-after-free
  sock_wfree -> skb_release_head_state -> kfree_skb
  -> skb_release_data -> __kfree_skb -> tcp_recvmsg ...

refcount_t: addition on 0; use-after-free
  skb_set_owner_w -> sock_alloc_send_pskb
  -> tun_get_user -> tun_chr_write_iter -> do_iter_write
  -> vfs_writev -> do_writev -> __arm64_sys_writev
```

So `runtime.KeepAlive` alone is insufficient -- GC-escape wasn't actually the mechanism on this kernel. The bug is kernel-side: `skb_set_owner_w` is bumping the TUN's internal socket `sk_wmem_alloc` refcount, and finding it already at zero. That's a kernel-owned refcount, not anything nebula's userspace memory touches.

Linux 4.19.x is missing several upstream TUN UAF fixes (most notably commit series starting with `tun: fix use-after-free on eventfd error in tun_sendmsg`). Valid-looking virtio_net_hdr TSO writes on forwarded paths can trip the 4.19 TUN driver's refcount bug.

These two commits are still worth landing as correctness fixes (pointer-through-uintptr without `runtime.KeepAlive` IS a latent bug even if it's not what's crashing this specific kernel), but they do not resolve the full crash on 4.19-based devices like UniFi UXG-Pro. A kernel-version gate on `TUN_F_TSO4|TSO6` (fall back to no-TSO when running on older kernels, similar to the IFF_VNET_HDR open/retry dance already in `tun_linux.go`) would be safer. I can follow up with that as a separate PR if useful.

@JackDoan if you have interest in teaching `newTun` to smoke-test GSO before committing to it -- e.g. open, push a small TSO frame, poll for `WARN` via /dev/kmsg, fall back if detected -- happy to write it up.

JackDoan:
@rawdigits yes please, no TSO allowed for e-waste running anything before 5.10

PR #1696, opened 2026-04-27
Support native Golang "fips140" mode

wadey:
Add support for the "fips140" mode of Go:

- https://go.dev/doc/security/fips140
- https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/5247

You can build with `make fips140`, see the README changes for more info.

Some differences from the boringcrypto builds:

- We switch to using `go:linkname crypto/tls.aeadAESGCMTLS13`, which gives us the fips implementation for both `boringcrypto` and `fips140` modes. This means we also no longer need `-checklinkname=0`
- Go native `fips140` doesn't need CGO_ENABLED=1
- We decide if we should use the fips140 GCM at runtime, if `fips140.Enabled()` is true. If you use the `make release-fips140`, we build with build tag `fips140enforce` which ensures the binary is running with fips140 enabled and that only P256 / AES-GCM is being used. If you don't want this enforce mode, you can build without the build tag.


Makefile
```diff
@@ -215,17 +215,27 @@ ifeq ($(words $(MAKECMDGOALS)),1)
...
+
 bin-docker: bin build/linux-amd64/nebula build/linux-amd64/nebula-cert
 
 smoke-docker: bin-docker
-	cd .github/workflows/smoke/ && ./build.sh
-	cd .github/workflows/smoke/ && ./smoke.sh
-	cd .github/workflows/smoke/ && NAME="smoke-p256" CURVE="P256" ./build.sh
-	cd .github/workflows/smoke/ && NAME="smoke-p256" ./smoke.sh
+	cd .github/workflows/smoke/ && $(GOENV) ./build.sh
+	cd .github/workflows/smoke/ && $(GOENV)./smoke.sh
```
nbrownus: Missing a space after `$(GOENV)`

noise.go
```diff
@@ -23,6 +23,11 @@ func NewNebulaCipherState(s *noise.CipherState) *NebulaCipherState {
 	return &NebulaCipherState{c: x.(cipher.AEAD)}
 }
 
+type cipherAEADDanger interface {
+	EncryptDanger(out, ad, plaintext []byte, n uint64, nb []byte) ([]byte, error)
+	DecryptDanger(out, ad, plaintext []byte, n uint64, nb []byte) ([]byte, error)
```
nbrownus: `plaintext` -> `ciphertext`

Makefile
```diff
@@ -215,17 +215,27 @@ ifeq ($(words $(MAKECMDGOALS)),1)
 	@$(MAKE) service ${.DEFAULT_GOAL} --no-print-directory
 endif
 
+fips140:
```
nbrownus, line 283: `boringcrypto` outputs with a suffix on the folder like: `build/linux-amd64-boringcrypto/nebula`

`fips140` outputs to a colliding path with non `fips140`: `build/linux-amd64/nebula`, seems like a good idea to push this into `build/linux-amd64-fips140/nebula`.

noise.go
```diff
@@ -33,33 +38,43 @@ func NewNebulaCipherState(s *noise.CipherState) *NebulaCipherState {
...
-		//s.n = n
-		nb[0] = 0
-		nb[1] = 0
-		nb[2] = 0
-		nb[3] = 0
-		noiseEndianness.PutUint64(nb[4:], n)
-		out = s.c.Seal(out, nb, plaintext, ad)
-		//l.Debugf("Encryption: outlen: %d, nonce: %d, ad: %s, plainlen %d", len(out), n, ad, len(plaintext))
-		return out, nil
+		switch ce := s.c.(type) {
```
nbrownus: Would be awesome to swap out the interface and avoid the type check for each packet.

Makefile
```diff
@@ -215,17 +215,27 @@ ifeq ($(words $(MAKECMDGOALS)),1)
...
+	@echo > $(NULL_FILE)
+	$(eval GOENV += GOFIPS140=v1.0.0)
```
nbrownus: We will also need to require go 1.26 and include this LDFLAGS: `-X runtime.godebugDefault=fips140=only`

Otherwise fips is enabled but all non fips crypto just keeps working. Will need to couple this with a change to assert `fips140.Enforced()` is true.

noiseutil/fips140.go
```diff
@@ -0,0 +1,78 @@
...
+	// unsafe needed for go:linkname
+	_ "unsafe"
+
+	"github.com/flynn/noise"
+)
+
+// EncryptLockNeeded indicates if calls to Encrypt need a lock
+// This is true for fips140 because the Seal function verifies that the
+// nonce is strictly increasing.
+const EncryptLockNeeded = true
```
nbrownus: The other part of `-X runtime.godebugDefault=fips140=only` is that we will want an `init` check to assert `fips140.Enforced()` here.

Makefile
wadey: good catch, fixed

noise.go
wadey: im confused, when you call this function it is still plaintext

wadey: err nm lol I see

Makefile
wadey: good catch! let me rework this.

wadey: added `release-fips140` and `bin-fips140` for this

noise.go
wadey: Fixed in latest commit

Makefile
wadey: added the default, but not enforcing and instead just logging the state of it. Considering the check.

noiseutil/fips140.go
wadey: Implemented with a build tag `fips140enforce`.

Makefile
wadey: Implemented with a build tag fips140enforce

jrwren (dismissed):

noiseutil/fips140.go
```diff
@@ -0,0 +1,118 @@
...
+	// We can clean this up in the future when NewGCMWithCounterNonce or
+	// NewGCMForQUIC are available:
+	if !bytes.Equal(emptyNonce, nonce) {
+		c.AEAD.Seal([]byte{}, emptyNonce, []byte{}, []byte{})
+	}
+	c.ready = true
+}
+
+func (c *aeadGCMFIPS140Cipher) Seal(dst, nonce, plaintext, additionalData []byte) []byte {
+	if !c.ready {
```
JackDoan, line 102: can we do this as part of `extractFIPSAEAD`? It would avoid a branch & technically also avoids an opportunity to race (thought my understanding is this _should_ always be called under a lock bc of how the fips nonce checker works

JackDoan (dismissed):

noiseutil/fips140.go
wadey: I think we can as long as nothing in our code encrypts with a "0" nonce. I think this is true. Ill test it

wadey: actually, the first WriteMessage that noise does is with nonce 0, so this is a bit tricky to optimize

JackDoan (dismissed):

pki.go
```diff
@@ -241,6 +242,9 @@ func newCipherSuite(curve cert.Curve, pkcs11backed bool, cipher string) (noise.C
 	var dhFunc noise.DHFunc
 	switch curve {
 	case cert.Curve_CURVE25519:
+		if fips140.Enforced() {
+			panic("pki: use of Curve25519 is not allowed in FIPS 140-only mode")
```
nbrownus: Thoughts on returning an error instead of throwing a panic?

wadey: We throw a panic for the other algorithms when fips140.Enforced is turned on. Let me check.

JackDoan: I'm happy with a panic as long as we can show it's not reachable from UDP

wadey: ill change to return an error, that matches what stdlib does in my research

JackDoan:
Do we want to check for and respect `/proc/sys/crypto/fips_enabled` on Linux?

JackDoan (dismissed):

noiseutil/fips140.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// init validates the go:linkname + reflection extraction and the nonce-reuse
+// protection at startup, in every build. cipherAESGCMFIPS140 relies on unexported
+// crypto/tls and crypto/internal/fips140 internals; if a future Go version changes
+// those, this fails fast with a clear message instead of panicking per-handshake
+// (or, worse, silently losing the strictly-increasing nonce check that is the whole
+// point of using this cipher). Because this file has no build tag, this self-test
+// runs even in non-FIPS builds, so the default CI lane catches an incompatible Go.
+func init() {
```
JackDoan: how do you feel about moving this to a _test.go file, and/or a build-tagged file? It'd be nice to avoid paying this startup cost on non-fips builds, especially on low-speed platforms

wadey: @JackDoan  We could wrap the block with `if (boringEnabled || fips140.Enabled)` ?

wadey: See latest commit which addresses this

noiseutil/fips140.go
```diff
@@ -0,0 +1,180 @@
...
+func (c *aeadGCMFIPS140Cipher) Encrypt(out []byte, n uint64, ad, plaintext []byte) []byte {
+	return c.Seal(out, aeadGCMFIPS140CipherNonce(n), plaintext, ad)
+}
+
+func (c *aeadGCMFIPS140Cipher) Decrypt(out []byte, n uint64, ad, ciphertext []byte) ([]byte, error) {
+	return c.Open(out, aeadGCMFIPS140CipherNonce(n), ciphertext, ad)
+}
+
+func (c *aeadGCMFIPS140Cipher) EncryptDanger(out, ad, plaintext []byte, n uint64, nb []byte) ([]byte, error) {
+	binary.BigEndian.PutUint64(nb[4:], n)
```
nbrownus, line 123: We need the exhaustion guard from #1841 here
https://github.com/slackhq/nebula/blob/edc3c5e0188659711b86413ed143c23af0384e3e/noiseutil/aesgcm.go#L28-L30

nbrownus: Probably want to take the `c` guard as well
```
	if c == nil {
		return nil, errors.New("no cipher state available to encrypt")
	}
```

nbrownus (approved):

PR #1703, opened 2026-05-04
Darwin and openbsd in line with the other bsds for tun support

nbrownus:
Gets Darwin and OpenBSD up to snuff with free/netbsd tun implementation.

The main problem is that the BSDs put a 4 byte header on their tun read/write and using readv/writev is an elegant way to deal with it. Since a raw syscall.Syscall(SYS_READV/WRITEV) won't fly on OpenBSD (pinsyscall) or darwin (the SVC trap Apple keeps threatening to disallow), we //go:linkname x/sys/unix's libc-routed readv/writev stubs instead.

I did a very light smoke of Darwin, NetBSD, OpenBSD, and FreeBSD.

### DIFF
--- a/overlay/tun_darwin.go
+++ b/overlay/tun_darwin.go
@@ -23,7 +23,8 @@ import (
 )
 
 type tun struct {
-	io.ReadWriteCloser
+	f           *os.File
+	rc          syscall.RawConn
 	Device      string
 	vpnNetworks []netip.Prefix
 	DefaultMTU  int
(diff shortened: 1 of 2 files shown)


JackDoan (dismissed):
you checked that the lambdas don't escape, right? I have trust issues with that

wadey:
Can we use `golang.org/x/sys/unix` which has `Readv` and `Writev` with Darwin support?

- https://github.com/golang/go/issues/64710

JackDoan (dismissed):

JackDoan (approved):

JackDoan (approved):

PR #1704, opened 2026-05-04
the definitive tun offloads branch

JackDoan:
Make nebula go fast by making fewer syscalls


rawdigits:
Bug: `sendNoMetrics` panics on relayed connections with small packets (e.g. TestReply)

**Panic:**
```
panic: runtime error: slice bounds out of range [:48] with capacity 32

goroutine 57 [running]:
github.com/slackhq/nebula.(*Interface).sendNoMetrics(...)
    inside.go:569
```

**Root cause:** `sendNoMetrics` saves `fullOut := out` before offsetting `out` for the relay header. After `EncryptDanger`, if the buffer was too small (cap < header + ciphertext + AEAD tag), Go allocates a new backing array for `out`. But `fullOut` still points at the original small buffer. Then `fullOut[:header.Len+len(out)]` overflows.

MTU-sized buffers (normal data packets) always have enough capacity so `EncryptDanger` writes in-place — `fullOut` and `out` share the same array and everything works. Small packets like `TestReply` (~32 bytes) don't — `EncryptDanger` reallocates, `fullOut` goes stale, panic.

**Conditions to trigger:** (1) relayed tunnel (no direct path between peers), AND (2) small packet (TestReply health-check ping is the common case). Direct tunnels never hit the relay branch; large packets never trigger reallocation.

**Fix (9 lines, applies cleanly to current PR HEAD):**

```diff
-			f.SendVia(relayHostInfo, relay, out, nb, fullOut[:header.Len+len(out)], true)
+			needed := header.Len + len(out)
+			if needed <= cap(fullOut) {
+				f.SendVia(relayHostInfo, relay, out, nb, fullOut[:needed], true)
+			} else {
+				// EncryptDanger reallocated out into a new backing array, so fullOut
+				// no longer contains the encrypted payload. Build a new relay buffer.
+				relayBuf := make([]byte, needed)
+				copy(relayBuf[header.Len:], out)
+				f.SendVia(relayHostInfo, relay, out, nb, relayBuf, true)
+			}
```

Happy path (buffer big enough) is identical behavior — just guarded by a cap check. The else branch handles the reallocation edge case.

Hit this in production on a host behind residential NAT where peer tunnels are relayed through lighthouses. Gateway routers with public IPs never trigger it because their tunnels are direct.

JackDoan:
@rawdigits fixed this a layer up by making relay outbounds also use the tx batcher, which also puts them on the fast path

firewall/cache.go
```diff
@@ -5,11 +5,15 @@ import (
...
+	"github.com/slackhq/nebula/logging"
 )
 
 // ConntrackCache is used as a local routine cache to know if a given flow
-// has been seen in the conntrack table.
-type ConntrackCache map[Packet]struct{}
+// has been seen in the conntrack table. Keyed on PacketKey (dense form)
+// rather than Packet so the lookup hashes raw bytes instead of the
+// unique.Handle each netip.Addr in Packet carries.
+type ConntrackCache map[PacketKey]struct{}
```
JackDoan: def considering pulling this part of the changeset out and doing it "later". It does help quite a bit with perf, since it makes conntrack map accesses quicker, but it's not really part of offloads per-se

main.go
```diff
@@ -33,6 +36,9 @@ func Main(c *config.C, configTest bool, buildVersion string, l *slog.Logger, dev
 		buildVersion = moduleVersion()
 	}
 
+	//todo no merge
+	go http.ListenAndServe(":6060", nil)
```
JackDoan: don't merge this!

firewall/cache.go
JackDoan: pulled out, diff smaller now. Perf actually hasn't changed much! We might be able to totally drop them

main.go
JackDoan: this is what's causing the e2e fails

pprof_debug.go
```diff
@@ -0,0 +1,33 @@
+//go:build debug
```
JackDoan, line 1: probably shouldn't merge in this PR either but it makes clod shut up when asking for reviews

cpupick/cpupick.go
```diff
@@ -0,0 +1,187 @@
+// Package cpupick chooses which CPUs the tun reader threads pin to when the
```
JackDoan, line 1: I think I'd prefer to pull this out as its own package

main.go
```diff
@@ -220,6 +232,21 @@ func Main(c *config.C, configTest bool, buildVersion string, l *slog.Logger, dev
...
+		// The operator didn't choose pin CPUs, so pick a default set that
+		// prefers performance cores and doesn't stack co-located instances
+		// onto allowed[0]. The bound UDP port keys the per-instance spread:
+		// distinct across instances sharing a box, stable across restarts.
+		// A nil result keeps listenIn's stock allowed[i] fallback.
+		key := uint64(os.Getpid())
+		if ap, err := udpConns[0].LocalAddr(); err == nil && ap.Port() != 0 {
+			key = uint64(ap.Port())
+		}
+		cpuAffinity = cpupick.Default(routines, key, l)
```
wadey, line 266: For cases like Kubernetes Nodes, we can have multiple Nebula instances all with the same `port`, I think this will make them all pick the same CPUs. Not sure if we can key off of something else additional that will make it better for this case. Not a blocker but perhaps a TODO

examples/config.yml
```diff
@@ -172,6 +172,8 @@ listen:
   # allowing for more precise routing decisions based on the packet tags. Default is 0 meaning no mark is set.
   # This setting is reloadable.
   #so_mark: 0
+  # the udp_offloads setting controls if Nebula will attempt to enable GSO and GRO for its UDP socket(s). Linux only, not reloadable.
+  # udp_offloads: true
```
wadey, line 176: Should we also expose a `tun_offloads` flag?

JackDoan: dangit I thought we had this, but we do not. :homer-spinning:

main.go
JackDoan: I could add a config bool to prefer PID? I'd been using just `port` for consistency on benchmark runs, but the k8s case is probably common-ish, and I bet there's others similar to it.

wadey (dismissed):

main.go
wadey: can do in a follow up if we want :+1:

JackDoan: nah it was an easy add, and thinking about it, I think the PID is a better default for most users.

nbrownus (approved):

PR #1708, opened 2026-05-06
Stop leaking goroutines past Control.Stop, consolidate punching in Punchy

nbrownus:
There are 2 commits, the first is the simplest thing we need to do to finalize the remaining goroutines. The second commit drops randomly spawned goroutines for a timerwheel scheduled punch experience. 

- Control.Stop now closes the SSH listener and tears down sessions
- Replace lighthouse fire-and-forget holepunch goroutines with a single timer-wheel worker on Punchy
- Move connection manager's keepalive punch logic into Punchy
- Make punchy.punch reloadable and normalize config change logging
- Drop goleak.IgnoreCurrent() from TestNoGoroutineLeaks

One thing that we should resolve is whether to collapse the punch metrics into a single metric or keep the "lighthouse" based stats split.

### DIFF
--- a/punchy.go
+++ b/punchy.go
@@ -13,12 +13,10 @@ import (
 	"github.com/slackhq/nebula/udp"
 )
 
-const (
-	holepunchTickDuration  = 250 * time.Millisecond
-	holepunchWheelDuration = 60 * time.Second
-)
+// holepunchQueueSize buffers the channel that pending holepunchJobs land on after their delay timer fires.
+const holepunchQueueSize = 64
 
-// holepunchJob is one scheduled item on the holepunch timer wheel.
+// holepunchJob is one scheduled item delivered to the worker goroutine.
 //   - target valid -> send a UDP punch to target. vpnAddr, if set, is the peer's vpn addr carried for log context.
 //   - target invalid, vpnAddr valid -> send an encrypted test packet to vpnAddr (a "punchback").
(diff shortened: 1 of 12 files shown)


punchy.go
```diff
@@ -59,52 +106,143 @@ func (p *Punchy) reload(c *config.C, initial bool) {
...
+
+		nb := make([]byte, 12, 12)
+		out := make([]byte, mtu)
+		empty := []byte{0}
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case now := <-clockSource.C:
```
JackDoan: It is unfortunate that we'll wake a CPU 4x a second here, no matter how often we actually intend to handshake. I know it's a little heapy, but I think the old goroutine-sleep way might be better? A channel with a deep buffer might also be a solution.

nbrownus: Agreed, added a new `Scheduler` to try and speak to the use case here. At worst we could just use `time.AfterFunc` and have pretty similar semantics to the old go routine spawn but with `time.AfterFunc` we at least avoid the stack allocation and go routine exhaustion situation. with the `Scheduler` we also gain zero alloc.

nbrownus: The other option I considered was lazy ticking the TimerWheel but the resolution loss for punchy delays felt bad.

JackDoan (approved):

PR #1715, opened 2026-05-08
Prime some critical stats before the first scrape

### LINKED ISSUE #907: 🐛 BUG: nebula certificate_ttl_seconds emits 0 until intialized

1.7.2

Linux

### Describe the Bug

We are now monitoring our certificate expiries thanks to the new feature built straight into the prometheus metrics. In addition to this, we have set up monitoring rules which let us know if the ttl_seconds falls below a certain threshold.

We started noticing an issue whenever nebula clients are restarted with this monitoring in place, that directly after boot, the cert_ttl_seconds metric is emitting a `0` until the certificate itself is fully initialized within Nebula. This has the unfortunate side effect of firing our monitors every time a nebula client is restarted with this monitoring in place.

```
~# curl -s 127.0.0.1:8090/metrics | grep cert
# HELP nebula_certificate_ttl_seconds certificate.ttl_seconds
# TYPE nebula_certificate_ttl_seconds gauge
nebula_certificate_ttl_seconds 0
~# curl -s 127.0.0.1:8090/metrics | grep cert
# HELP nebula_certificate_ttl_seconds certificate.ttl_seconds
# TYPE nebula_certificate_ttl_seconds gauge
nebula_certificate_ttl_seconds 60803
```

It seems to me it would be better to not emit this metric at all if the value it provides is inaccurate -- prometheus will use the last value scraped until a new value is provided to update it.

### Logs from affected hosts

_No response_

_No response_

nbrownus:
Prime some stats, mainly certificate ttl seconds, before the first prometheus scrape can happen to avoid reporting 0 ttl cert lifetimes.

Adds a test to catch regressions.

Closes #907

### DIFF
--- a/interface.go
+++ b/interface.go
@@ -491,26 +491,34 @@ func (f *Interface) emitStats(ctx context.Context, i time.Duration) {
 	certInitiatingVersion := metrics.GetOrRegisterGauge("certificate.initiating_version", nil)
 	certMaxVersion := metrics.GetOrRegisterGauge("certificate.max_version", nil)
 
+	emit := func() {
+		f.firewall.EmitStats()
+		f.handshakeManager.EmitStats()
+		udpStats()
+
+		certState := f.pki.getCertState()
+		defaultCrt := certState.GetDefaultCertificate()
+		certExpirationGauge.Update(int64(defaultCrt.NotAfter().Sub(time.Now()) / time.Second))
+		certInitiatingVersion.Update(int64(defaultCrt.Version()))
+
+		// Report the max certificate version we are capable of using
+		if certState.v2Cert != nil {
+			certMaxVersion.Update(int64(certState.v2Cert.Version()))
+		} else {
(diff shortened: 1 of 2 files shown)


interface.go
```diff
@@ -481,36 +481,47 @@ func (f *Interface) reloadMisc(c *config.C) {
...
+	udpStats()
 
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			f.firewall.EmitStats()
-			f.handshakeManager.EmitStats()
+			f.emit()
```
jasikpark: I assume moving these prometheus gauge updates before `udpStats()` is intentional?

jasikpark (approved):
lgtm

interface.go
nbrownus: Cleaned it up a bit

wadey (approved):

PR #1717, opened 2026-05-08
Search for config.yaml/yml in both service and cli mode

### LINKED ISSUE #755: FEATURE: default value for `-config`

master

Mac

### Describe the Bug

I want to email my friend a zip file with nebula.exe + config.yml + certs, and tell them "double-click nebula.exe to connect to my network"

service.go already contains code for "if config isn't specified, check to see if there's a config file next to the binary":
https://github.com/slackhq/nebula/blob/4453964e34631cb1fce827eb11360da36f7774bf/cmd/nebula-service/service.go#L53-L59

Could this code be used in the other places, so we default to "check the current folder" instead of "exit with an error"?

Now that I write this out, I realise that I could also include a `.bat` file in the `.zip` which just runs `./nebula.exe -config config.yml`, but I feel like "check the current folder" would be nicer, and `service.go` is doing that already, so really we'd just be making it consistent :)

### Logs from affected hosts

_No response_

_No response_

nbrownus:
Service mode already searched the cwd for `config.yaml` or `config.yml`, this change make the direct cli mode do the same thing.

Closes #755

### DIFF
--- a/cmd/nebula-service/main.go
+++ b/cmd/nebula-service/main.go
@@ -61,9 +61,12 @@ func main() {
 	}
 
 	if *configPath == "" {
-		fmt.Println("-config flag must be set")
-		flag.Usage()
-		os.Exit(1)
+		p, err := config.DefaultPath()
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+		*configPath = p
 	}
 
 	c := config.NewC(l)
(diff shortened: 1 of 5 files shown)


johnmaguire (dismissed):
lgtm

johnmaguire (approved):

wadey (approved):

PR #1726, opened 2026-05-18
Surface previously-swallowed errors in DNS, SSH, and config paths

randomizedcoder:
### Hi 👋

Five places inside `dns_server.go`, `sshd/session.go`, and `config/config.go` quietly swallowed the error returned by a write/reply/stat call. In each case the operator saw normal output or a misleading success while a real failure (UDP write error, channel torn down mid-handshake, addFile failing on an unresolvable path) went unrecorded. This PR resolves all five at the root, in five small bisect-safe commits. Where the calling API can carry an error, it is now propagated; where it cannot (DNS write callbacks, SSH protocol replies inside a goroutine), the failure is now logged via `slog.Warn` and counted in a `rcrowley/go-metrics` counter so it shows up in the existing metrics export.

### How we found these

We ran [`errcheck`](https://github.com/kisielk/errcheck) via golangci-lint as part of a strict static-analysis pipeline downstream of this fork. `errcheck` flags every call whose returned error is discarded. After triaging the full report into "real bugs", "cosmetic (test setup, defer-close)", and "false-positive", five real-bug call sites remained:

```
config/config.go:351:14:  Error return value of `c.addFile` is not checked
dns_server.go:194:14:     Error return value of `w.WriteMsg` is not checked
sshd/session.go:85:13:    Error return value of `req.Reply` is not checked
sshd/session.go:89:13:    Error return value of `req.Reply` is not checked
sshd/session.go:96:21:    Error return value of `channel.SendRequest` is not checked
```

Each line maps to one commit.

### Per-site operator impact

| File:line | Function | Pre-fix operator experience |
|---|---|---|
| `config/config.go:351` | `(c *C).resolve` | A user-specified config path whose `filepath.Abs` fails is silently dropped — the user sees "no config files found" with no hint of the real cause |
| `dns_server.go:194` | `handleDnsRequest` | `WriteMsg` failures (client socket closed, FD exhaustion, UDP send error) are invisible; clients quietly timeout and the operator has no signal |
| `sshd/session.go:85` | `handleRequests` (bad-payload reject) | A malformed exec request rejection that fails to reach the client is invisible — the user sees a hung session, the operator sees nothing |
| `sshd/session.go:89` | `handleRequests` (accept-exec) | Same shape, but the consequence is worse: today the handler still proceeds to dispatch the command and emit exit-status to a client that never confirmed the accept |
| `sshd/session.go:96` | `handleRequests` (exit-status) | The final exit-status notification can fail with no record; the client appears to hang for a moment before close |

The `dns_server.go` case is the one most likely to bite an operator in the wild — a single failing client (e.g. a closed-socket replay or a transient UDP error) is currently silent, so a misbehaving DNS consumer is invisible until it is widespread enough to show up elsewhere.

### Design notes

We thought carefully about how to handle each site:

1. **No retry**, even for the DNS WriteMsg case. UDP DNS responses are stateless: if the client socket is gone, retrying the same response with the same DNS transaction ID is a protocol violation. FD exhaustion is not solvable by retry. The DNS protocol already expects the resolver layer to retry on no-answer, so adding handler-side retry would only delay the eventual `WriteMsg` failure. **Decision: observe, do not retry.**

2. **No new dependencies.** The DNS commit adds a `~20`-line atomic-int-based `dnsWriteWarnLimiter` rather than pulling in `golang.org/x/time/rate`. The limiter coalesces high-rate failure bursts so a runaway client does not flood the journal: the first failure in a 10-second window logs immediately with the count of suppressed messages since the last log, the rest only increment the counter. (The counter is exposed regardless of the log rate-limit.)

3. **`rcrowley/go-metrics` counters everywhere**, following the same dot-separated registry convention the rest of nebula already uses (e.g. `messages.tx.*`). Three new counters:
    - `dns.responses.write_failures`
    - `sshd.reply.errors`
    - `sshd.send_request.errors`

4. **`replyer` / `requester` interfaces** for the SSH side. `*ssh.Request.Reply` and `*ssh.Channel.SendRequest` are not testable in isolation without standing up a full SSH transport; we extracted single-method interfaces and route the call sites through small helpers (`replyAndLog`, `sendRequestAndLog`). `*ssh.Request` and `*ssh.Channel` satisfy them unchanged; tests inject fakes.

5. **One commit bails out, the others observe.** Specifically, the SSH `accept-exec` site (sshd/session.go:89) now branches on the helper's return: if the accept-reply fails, the handler closes the channel and bails without running the user's command — proceeding would dispatch a command for which the client never received the accept and would lead to a confused exit-status that the client cannot correlate. The other SSH sites and the DNS site only observe, because at those points the work is either done or there is no useful follow-up action.

### Commit map

1. `Return addFile errors from config.resolve` — propagates `filepath.Abs` failures through `addFile` to `resolve`. Test injects a fake `filepathAbs` via a package-level var.
2. `Log and meter dns_server WriteMsg failures` — adds the `dnsWriteWarnLimiter` (with a 6-row table-driven test) and routes `WriteMsg` errors through it. Counter `dns.responses.write_failures` increments unconditionally; the Warn is rate-limited to one per 10s.
3. `Log and meter SSH protocol reply failures from session handler` — introduces `sshd/metrics.go`, `sshd/reply.go` (`replyer` interface + `replyAndLog`), and `sshd/reply_test.go` (`TestReplyAndLog`, 4-row table). Routes the bad-payload reject site through the helper.
4. `Bail out of SSH exec handler when accept-reply fails` — routes the accept-reply through `replyAndLog` and branches on its return so the handler stops cleanly when the accept never reaches the client.
5. `Log and meter ssh.Channel.SendRequest failures` — extends `sshd/reply.go` with `requester` + `sendRequestAndLog`, adds `metricSendRequestErrors`, routes the exit-status request through the helper. `TestSendRequestAndLog` is a 4-row table covering success/failure × wantReply.

Mutation-tested: at every commit that adds an observability helper, I removed the metric line and the log line in turn and confirmed the table test catches each mutation. The tests are not tautologies.

All five commits are independently bisect-safe — `go test -count=1 ./...` passes at every SHA on the branch.

### How to run the new tests locally

```
go test -count=1 -v -run 'TestConfig_Resolve_AddFileErrorPropagates' ./config
go test -count=1 -v -run 'TestDNSWriteWarnLimiter|TestDnsServer_HandleDnsRequest_WriteMsgFailure' .
go test -count=1 -v -run 'TestReplyAndLog|TestSendRequestAndLog' ./sshd
```

### Backward compatibility

No CLI flag, config key, or public API changes. The only observable behaviour change is that previously-silent failures now appear in the metrics export and the journal, plus the SSH accept-exec path now cleanly aborts on a broken handshake rather than dispatching a command whose accept never landed. All existing tests continue to pass with no edits.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -335,8 +335,7 @@ func (c *C) resolve(path string, direct bool) error {
 	}
 
 	if !i.IsDir() {
-		c.addFile(path, direct)
-		return nil
+		return c.addFile(path, direct)
 	}
 
 	paths, err := readDirNames(path)
@@ -354,6 +353,13 @@ func (c *C) resolve(path string, direct bool) error {
 	return nil
 }
 
+// filepathAbs is the abs-path resolver used by addFile. It is a package
+// level var so tests can swap in a function that errors to exercise the
+// previously-swallowed propagation path. Production always uses
+// filepath.Abs. Tests that mutate filepathAbs cannot run with
+// t.Parallel().
+var filepathAbs = filepath.Abs
+
 func (c *C) addFile(path string, direct bool) error {
 	ext := filepath.Ext(path)
 
@@ -361,7 +367,7 @@ func (c *C) addFile(path string, direct bool) error {
 		return nil
 	}
 
-	ap, err := filepath.Abs(path)
+	ap, err := filepathAbs(path)
 	if err != nil {
 		return err
 	}
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"testing"
@@ -38,6 +39,34 @@ func TestConfig_Load(t *testing.T) {
 		"new": "hi",
 	}
 	assert.Equal(t, expected, c.Settings)
+}
+
+// TestConfig_Resolve_AddFileErrorPropagates exercises the previously-
+// swallowed addFile error path inside (*C).resolve. With a real
+// filesystem we cannot easily make filepath.Abs error (it only fails
+// when os.Getwd does, which requires deleted-cwd manipulation that is
+// platform-fragile), so we follow the same pattern as
+// cmd/nebula-cert/stdio.go's stdinReader: swap the package-level
+// filepathAbs func for one that always returns an error.
+//
+// Cannot run with t.Parallel() — mutates the package-level
+// filepathAbs var.
+func TestConfig_Resolve_AddFileErrorPropagates(t *testing.T) {
+	dir := t.TempDir()
+	target := filepath.Join(dir, "x.yml")
+	require.NoError(t, os.WriteFile(target, []byte("k: v\n"), 0o644))
+
+	injected := errors.New("injected filepath.Abs failure")
+	origAbs := filepathAbs
+	t.Cleanup(func() { filepathAbs = origAbs })
+	filepathAbs = func(string) (string, error) { return "", injected }
+
+	c := NewC(test.NewLogger())
+	err := c.Load(target)
+
+	require.Error(t, err, "addFile error must propagate from resolve via Load")
+	require.ErrorIs(t, err, injected,
+		"propagated error must wrap the underlying cause so callers can errors.Is")
 }
 
 func TestConfig_Get(t *testing.T) {
--- a/dns_server.go
+++ b/dns_server.go
@@ -10,10 +10,55 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/miekg/dns"
+	"github.com/rcrowley/go-metrics"
 	"github.com/slackhq/nebula/config"
 )
+
+// dnsWriteWarnLogWindow gates the WriteMsg-failure Warn log line. Failures
+// within the window are counted; the next Warn after the window elapses
+// includes the suppressed-count as an slog attr so an operator can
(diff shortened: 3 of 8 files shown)


JackDoan (changes requested):
Please break this up by subsystem (and combine it with the broken up results of your other PRs). Separate PRs for config, ssh, and DNS improvements are easier to review, and don't require cross-changeset context.

randomizedcoder:
Thanks for the thoughtful reviews. I'll get on it.

Regards,
Dave Seddon
+1 415 857 5102

randomizedcoder:
Thanks for the review. As suggested, I've broken this up by subsystem and combined the relevant bits with the broken-up results of #1725.

The three new subsystem PRs are:
- **config**: #1737 — folds the addFile-error fix from this PR together with the stat-error fix from #1725.
- **dns**: #1738 — the DNS subsystem bits from this PR (atomic-int rate limiter for the `WriteMsg`-failure Warn, plus the `dns.responses.write_failures` counter).
- **ssh**: #1739 — combines the `sshd/session.go` sites from this PR with the `ssh.go` swallow sites from #1725 (reworked to log+meter per your suggestion on that PR).

A few cleanups applied during the rewrite:
- Comments trimmed to 1-2 lines max throughout.
- Non-ASCII characters scrubbed (em-dashes replaced with hyphens).
- Test scaffolding minimized — each new test reads like `sshd/reply_test.go` (counter delta + log-buffer match), no large fixture builders.

Closing this PR in favor of the three subsystem PRs. Thanks again!

PR #1733, opened 2026-05-19
Reduce relay handshake log spam

nbrownus:
This PR dampens relay handshake logging in similar fashion as normal handshakes. First attempt is INFO logged, subsequent attempts are DEBUG logged unless there is a new relay vpn address to try.

### DIFF
--- a/relay_manager.go
+++ b/relay_manager.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/netip"
+	"slices"
 	"sync/atomic"
 
 	"github.com/slackhq/nebula/cert"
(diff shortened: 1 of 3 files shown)


relay_manager.go
```diff
@@ -75,20 +87,27 @@ func (rm *relayManager) StartRelays(f *Interface, vpnIp netip.Addr, hostinfo *Ho
...
+
 		// Check the relay HostInfo to see if we already established a relay through
 		existingRelay, ok := relayHostInfo.relayState.QueryRelayForByIp(vpnIp)
 		if !ok {
 			// No relays exist or requested yet.
 			if relayHostInfo.remote.IsValid() {
 				idx, err := AddRelay(rm.l, relayHostInfo, rm.hostmap, vpnIp, nil, TerminalType, Requested)
 				if err != nil {
-					hostinfo.logger(rm.l).Info("Failed to add relay to hostmap", "relay", relay.String(), "error", err)
+					hl.Info("Failed to add relay to hostmap", "relay", relay.String(), "error", err)
```
JackDoan, line 110: not this PRs fault but an err where we don't bail is suspicious to me

relay_manager.go
```diff
@@ -57,14 +58,25 @@ func (rm *relayManager) GetUseRelays() bool {
...
-	hostinfo.logger(rm.l).Info("Attempt to relay through hosts", "relays", hostinfo.remotes.relays)
+	relays := hostinfo.remotes.relays
+	listLevel := slog.LevelDebug
+	prior := hh.lastRelays
+	if !slices.Equal(relays, prior) {
+		listLevel = slog.LevelInfo
+		hh.lastRelays = slices.Clone(relays)
+	}
+	hl := hostinfo.logger(rm.l)
+	hl.Log(context.Background(), listLevel, "Attempt to relay through hosts", "relays", relays)
```
JackDoan, line 76: imo it's potentially confusing to see this always print, and the other lines no longer print

JackDoan: I'd almost rather demote this line to always debug, and keep the others as always info. Is that weird?

nbrownus: This is following the same pattern as regular handshake logs, log the first one at info and only log subsequent attempts at info if there are new relays to try, otherwise debug level.

JackDoan: perhaps I'm out of touch :P

you're good with the slices.Clone call?

JackDoan (approved):

PR #1741, opened 2026-05-25
udp: add io_uring, complementary to #1704

randomizedcoder:
# Summary

Follow-up to discussion #1736. Proposes an io_uring-backed `udp.Conn` that layers on top of the GSO/GRO work in #1704. Built and tested against #1704's HEAD (`c08ed7d` as of 2026-05-24), so this PR can land any time after #1704 is merged — or earlier if you'd prefer to review them together.

First — genuine appreciation for #1704. The 2 Gbit → 10+ Gbit jump is the headline result, and the GSO/GRO + virtio_net + asm checksums + CPU pinning combination is a serious piece of work. Excited to see it proceed.

# Position

io_uring is **complementary**, not competing. @JackDoan's two points in #1736 are correct:

- **GSO/GRO is the bigger throughput lever.** Software UDP_GSO/GRO coalesces ~40 packets into one syscall on each side — io_uring's per-syscall batching can't match that. This PR doesn't try to.
- **TUN can't benefit from io_uring** (no `read_iter`/`write_iter` in the kernel driver). We don't touch TUN here.

What io_uring *does* buy, layered on top of #1704's GSO/GRO:

- **Lower p99 latency** on the UDP socket — async submission/completion, no goroutine parking on `recvmsg`/`sendmsg`.
- **CPU reduction** on the outside Conn under concurrent senders (handshake, lighthouse, punchy, relay all share that socket).
- **Forward path to SQPOLL** (zero user-mode syscalls in the steady state) as a follow-up.

Realistic delta vs the recvmmsg+GSO/GRO baseline is single-digit % CPU and µs-level p99 latency. The throughput headline stays with you.

# What's in this PR

Four commits, organized for sequential review:

1. **`lint: clean up pre-existing findings on #1704`** — 19 mechanical lint fixes across 7 files (intrange, perfsprint, errorlint, commentFormatting, indent-error-flow, S1023) plus removal of an unused `headerCounter` function (looks like leftover from the May 6 "remove udp-level RX reorder buf" commit). Zero behavioural change.

2. **`docs: expand ECT acronym and fix typo`** — Inline RFC 3168 terminology so readers unfamiliar with ECN can follow the RFC 6040 combine logic. Also fixes `catastraphic` → `catastrophic`.

3. **`udp: add io_uring backend with unit tests`** — The main contribution. ~1,500 LOC across 7 new files behind the `iouring` build tag. Reuses #1704's cmsg byte layout exactly (`UDP_SEGMENT`, `IP_TOS`/`IPV6_TCLASS`, `UDP_GRO`), so the tx batcher and RFC 6040 `applyOuterECN` work unchanged.

4. **`udp: add io_uring integration and race tests`** — `-race`-validated concurrent tests plus a contention-curve benchmark.

# Design highlights

- **Build-tag gated** (`-tags iouring`). Default builds are bit-for-bit identical to #1704. The `listen.io_uring` config flag opts in at runtime; falls back gracefully on kernels below 5.6 or on builds without the tag.

- **Multi-ring send** (default 4 shards, max 32). The outside Conn has 5-6 concurrent writer goroutines; a single ring + mutex would serialize their entire send latency including `WaitCQE`. `acquireSendRing` picks a shard via `atomic.Uint32.Add % N`, scans with `TryLock` to skip past hot rings, and yields via `runtime.Gosched` if every shard is briefly held — never a busy spin.

  `BenchmarkIoUringConn_concurrent_WriteTo_rings` on a Threadripper PRO 3945WX:

  | R rings | G goroutines | ns/op | vs R=1 G=1 |
  |---:|---:|---:|---:|
  | 1 | 1 | 8389 | 1.0× |
  | 1 | 4 | 9911 | 0.85× (contention) |
  | 4 | 4 | 2479 | **3.4×** |
  | 8 | 8 | 1291 | 6.5× |
  | 16 | 16 | 756 | **11.1×** |

- **Close coordinates with in-flight senders.** Found and fixed (caught by `-race`) — Close was originally unmapping rings while `WriteTo` was still using them. Now Close waits for `ListenOut` via `sync.WaitGroup` (bounded), and acquires each send ring's mutex before `QueueExit`. `WriteTo`/`writeBatchChunk` re-check `c.closed` after acquiring their ring's mutex.

- **Config knobs** (all under `listen:`, none reloadable since rings are sized at kernel-side setup):

  ```yaml
  io_uring: false                 # master switch (default off)
  io_uring_send_rings: 4          # shard count, 1..32
  io_uring_recv_ring_size: 512    # power of 2, max 32768
  io_uring_send_ring_size: 512    # power of 2, max 32768
  io_uring_recv_slots: 256
  io_uring_send_slots: 256        # per ring
  ```

  Non-power-of-2 ring sizes are rounded up with a warning; oversized slot counts are clamped to ring size.

# Tests

- Unit tests for pure cmsg/sockaddr parsers + option validation.
- Concurrent race tests under `-race` (8 goroutines × 1000 sends; mixed `WriteTo`/`WriteBatch`).
- Close-during-send race test (caught the Close coordination bug).
- Microbenchmarks: `parseRecvCmsgRaw` 6.3 ns/op, `parseSockaddrFromRaw` 2.2 ns/op, both zero-alloc.
- Multi-ring contention benchmark (above).

Passes under `go test -race -tags iouring -count=50 ./udp/...`.

# Two findings worth flagging (not fixed here)

While running comprehensive race tests, found two pre-existing races in the tree that are **not** introduced by this PR. Happy to send separate PRs for these if you'd like:

1. **`overlay/tio/tio_poll_linux.go:160` (Close write) vs `:111` (readOne read)** — `TestPoll_WakeForShutdown_WakesFriends` fails under `go test -race`. Came in via #1704 commits `4b4331b` and `6a46a29`.

2. **`dns_server.go:176` (Start write) vs `dns_server_test.go:201` (read)** — `TestDnsServer_reload_sameAddr_noOp` fails under `go test -race`. Pre-#1704 (originally landed in PR #1667 "No more dns globals").

# Dependency

`github.com/randomizedcoder/giouring v1.0.1` — a fork of `pawelgaczynski/giouring`. Same version `randomizedcoder/gosrt` uses in production. The fork has a small tweak; happy to switch to upstream once that lands.

# Notes

- This PR targets `better-tun-interface-ordering` (#1704's branch) so the diff is just our incremental work. If you force-push #1704 this branch needs a rebase.

- Happy to revise scope, naming, or any design choice based on your feedback.

Thanks for considering — and again, looking forward to seeing #1704 land.


randomizedcoder:
Pushed a follow-up commit ([`f41dd91`](https://github.com/slackhq/nebula/pull/1741/commits/f41dd91)) addressing the first of the two race findings noted above — `overlay/tio/tio_poll_linux.go` `Poll.Close` vs `readOne`.

The fix is a one-line removal of a redundant `t.fd = -1` sentinel write. `t.closed` (`atomic.Bool`) already guards idempotency, so the field write only existed to race with concurrent readers. In-flight syscalls against the closed fd return EBADF, which `readOne`/`Write` already translate to `os.ErrClosed`. `TestPoll_WakeForShutdown_WakesFriends` now passes under `-race`.

The `dns_server.go` race is still outstanding — happy to send a separate PR for that one if you'd like, since it pre-dates #1704 and might land independently.

randomizedcoder:
Pushed a follow-up commit ([`8790f7d`](https://github.com/slackhq/nebula/pull/1741/commits/8790f7d)) addressing the second race finding — the `dns_server.go` `Start` vs test read of `d.server`.

The production code writes `d.server` under `d.serverMu` (in `Start` and `Stop`); the race was three tests reading the field directly without the mutex. Added a test-only `currentServer()` accessor that reads under `d.serverMu` and routed the three `assert.Nil(t, ds.server)` calls through it. `TestDnsServer_reload_sameAddr_noOp` now passes under `-race`.

Both pre-existing races from the original PR description are now addressed. Full `go test -race ./...` is clean except for build-tag-gated paths that aren't run by default.

JackDoan:
Interesting!

Is this actually faster though? Single-stream iperf3 under perf is a decent smoke test (iperf3 from many Nebula hosts is a better real world test though)

Results from `perf stat -d -d -d` would also be interesting to compare 

How's memory usage? One of the things holding me back from io_uring has been the need to reserve many full-size superpacket buffers up front, per routine.

emmibarra:
I +3 approve this MR

PR #1751, opened 2026-06-09
correctly record window counters for relayed packets in a tunnel

JackDoan:


outside.go
```diff
@@ -181,6 +181,13 @@ func (f *Interface) handleOutsideRelayPacket(hostinfo *HostInfo, via ViaSender,
 	if err != nil {
 		return
 	}
+	// Advance the replay window now that the frame is authenticated
+	if !hostinfo.ConnectionState.window.Update(f.l, h.MessageCounter) {
+		if f.l.Enabled(context.Background(), slog.LevelDebug) {
+			hostinfo.logger(f.l).Debug("dropping out of window relay packet", "header", h)
```
JackDoan, line 187: is there a metric we need to .Inc here?

wadey: the metrics should be updated in window.Update, example: https://github.com/slackhq/nebula/blob/b041f306cb2ee8177dfcb9ed8e7f31e6e7fe6564/bits.go#L179

wadey (approved):

Issue #1446, opened 2025-08-05
Feature Request: Post-quantum cryptography


0-wHiTeHand-0:
Hi, this is just a request for the medium/long-term. It would be great to start introducing post-quantum crypto similarly as [Rosenpass](https://rosenpass.eu/) or other projects. Quantum computers are not a threat nowadays (as far as I know), but they will be in the future.

nbrownus:
The best method that I personally know of currently would be to use a noise variant that supports kyber768 or above. The public keys in this scheme are prohibitively large which leaves very little space for the other certificate details to fit within a single handshake packet. `nebula` uses a certificate that is signed by a mutually trusted authority to establish identity characteristics which benefit the built in firewall to allow assertions of traffic flow beyond a simple ip address. Supporting this would require a move to QUIC, TCP, or something else similar at the transport layer, which would introduce backwards incompatibility.

Another option would be similar to what [wireguard offers](https://www.wireguard.com/known-limitations/#post-quantum-secrecy), a per tunnel PSK (which is negotiated using a PSQ technique of your choice) to harden each tunnel. The main downsides is that the security of that key exchange as well as the exchange itself are left entirely up to the network operator.

We can also gain a slight edge with #1273, which provides for a mesh wide PSK to increase security, which is less beneficial than a per tunnel PQS negotiated PSK.

Clancy508:
A third option is the OpenSSH route - use key encapsulation with a post quantum protocol. That leaves auth quantum insecure but is supposed to at least protect against capture and decrypt later type attacks.

Another solution to look into is https://rosenpass.eu/, which is the setup that Netbird uses, although that would either require significant adaptation or implementing a PSK mechanism anyway that Rosenpass can hook into

0-wHiTeHand-0:
Just went through this [NIST announcement](https://nvlpubs.nist.gov/nistpubs/ir/2024/NIST.IR.8547.ipd.pdf). TL;RD conventional cryptographic algorithms should be replaced by post-quantum cryptography by 2030.
In addition to Rosenpass, another example of implementing PQC is Cloudflare's zero trust platform. [They already apply](https://developers.cloudflare.com/ssl/post-quantum-cryptography/) ML-KEM, which is the PQ key agreement selected by NIST (it seems that kyber768 is obsolete).

mossholderm:
Just to keep this up to date, I've seen a couple of articles today about using Merkle Tree certificates to address the cert size issue.   

[...]

PR #1761, opened 2026-06-14
pq: post-quantum PSK handshake (ix_psk2) with pluggable provider

natoopotato:
Built this because I wanted it. Running it in production, no issues anymore, so figured I'd make it available. Take what's useful, ignore the rest. Refs #1446.

## What it is

PQ (harvest-now-decrypt-later) resistance: mix a 32-byte PSK into the Noise handshake. New `ix_psk2` subtype, PSK at placement 2. PSK comes from a pluggable `pq.Provider`. There's a rosenpass sidecar adapter and an embedded one, but the point of the interface is you can plug in an HSM/KMS/other KEM daemon without touching handshake code.

## Design

- **Off by default, byte-identical to stock.** No provider configured (`pki.pq_psk_dir`/`pki.pq_psk_path` unset) ⇒ no PSK ⇒ classical handshake, existing path untouched. 137 deletions total; ~40 lines of pre-existing handshake code modified, the rest is new code behind a subtype guard.
- **Wire-compatible.** Stock nodes ignore the cert-v2 extension and the unknown gossip field, interop on IXPSK0. PQ ↔ vanilla still talk.
- **Not rosenpass-locked.** `pq.Provider` is the contract; file/memory/static/composed providers all implement it.
- **Embedded path** behind `-tags rosenpass_embedded`. Default build never pulls the unaudited Go rosenpass port.

## In it

- `pq/`: provider interface, providers, policy modes (`opportunistic`/`required`), binding validation, multi-epoch skew tolerance, metrics.
- Handshake: `ix_psk2` subtype, PSK injection, prev-epoch swap on AEAD reject (skew healing both directions).
- IXPSK0 degrade (see below).
- Cert-v2 `rosenpassPubKeySha256` binding + `nebula-cert sign` flags.
- Optional lighthouse gossip of rosenpass identity/ports.
- `pq-status` ssh command.
- Operator guide + ~7k lines tests (~0.84 test:code).

## Security model

- `opportunistic` beats the passive HNDL adversary, not an active on-path MITM (who can still force classical). `required` is PQ-or-nothing.
- **IXPSK0 degrade** — flagging this one. Two relay lighthouses with desynced rosenpass PSKs can't complete IXPSK2, and the re-key that'd fix it can't cross the never-formed tunnel: permanent wedge. After repeated *proven* msg2 rejects (not timeouts — a dropper can't trigger it) the initiator drops to IXPSK0 for 60s so the classical tunnel forms, rosenpass re-keys, next attempt upgrades. Availability over strip-resistance, and coupled to the sidecar-over-tunnel setup — core nebula probably shouldn't carry it. Gated to `opportunistic`, every episode metered + alarmed. Easy to drop.
- Binding ships at `warn` (logs mismatch, uses PSK anyway). `enforce` works but isn't production-tested.

## Size

~8.5k prod + ~7k tests. Not one reviewable PR. Happy to split — the generic PSK-provider + mesh-wide IXPSK0 is the smallest standalone slice; degrade and the rosenpass adapter come later and are easy to leave out. Or just cherry-pick, no attribution needed.

Thanks for nebula.


nbrownus:
Thanks for the significant effort here, there's a lot of careful thinking in this. Post quantum is something we're tracking closely and actively working on ourselves. It touches the most fundamental parts of how Nebula communicates, the handshake and the wire protocol, and that's an area where we need to own the design end to end and move deliberately. For that reason we aren't taking outside contributions on the foundational protocol layers right now, and I'd rather be upfront about that than leave a PR of this size sitting open indefinitely. I'm going to close this one, but the work is appreciated and #1446 is the right place to keep shaping where we take it.

PR #1770, opened 2026-06-19
another imagining of an info socket

JackDoan:
not at all merge-ready

I vibed this up with minimal guidance with the objective/expectation of producing the median info-socket-API thing. I wouldn't have named it the host-query-API, but other than that, I think I really like the semantics of this. The robot pointed out that http-on-unix-socket is pretty popular because it's curl-able, self-packetizing, and the `docker` command line tool does it, therefore it can't be that crazy.


JackDoan:
#1821 replaces

PR #1775, opened 2026-06-29
fix(bits): add RWMutex to make Bits safe for concurrent use

rishuranjanofficial:
## Problem

`Bits.Check` and `Bits.Update` are called from multiple goroutines when
`routines: 2` or higher is configured, but `Bits` has no internal
synchronisation. This is a data race on `b.current` and `b.bits[]`,
detectable with `go test -race`.

Concurrent writes to the bitmap can corrupt it, causing valid future
packets to be incorrectly treated as replays and silently dropped.

## Fix

Add `sync.RWMutex` to `Bits`:

- `Check` holds `RLock` — multiple concurrent pre-flight reads are fine.
- `Update` holds `Lock` — all mutations serialised.

The TOCTOU between `Check` and `Update` is intentional and safe:
`Update` is the authoritative dedup step. Two goroutines racing through
`Check` results in at most one redundant decryption; the second
`Update` returns `false` and the duplicate is discarded before
delivery. No packet is ever delivered twice.

No API or protocol changes. All callers in `outside.go` are unchanged.

## Testing

Added `TestBitsConcurrentCheckUpdate` — run `go test -race -run TestBits ./...`
to confirm the race detector is silent after this change.

Note: `TestDnsServer_reload_sameAddr_noOp` reports a pre-existing race
on `dns_server.go` that is present on `master` and unrelated to this PR.

### DIFF
--- a/bits.go
+++ b/bits.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"math"
 	mathbits "math/bits"
+	"sync"
 
 	"github.com/rcrowley/go-metrics"
 )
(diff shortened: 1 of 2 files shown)


johnmaguire:
Closed by #1802

PR #1788, opened 2026-07-07
hostmap: unlink a multi-vpnAddr hostinfo from the shared chain exactly once on delete

nbrownus:
`unlockedDeleteHostInfo` unlinked a hostinfo's shared `next`/`prev` chain once per
vpnAddr instead of once per hostinfo. On the first address it nil'd the shared
pointers; on the second address the node then looked already-detached, so
`isLastHostinfo` went spuriously true and the code dropped that address's map entry
(and disestablished its relays) instead of promoting the live sibling.

This rewrites the delete to: compute `isLastHostinfo` up front, promote the successor
for every address where this hostinfo is the head, then splice it out of the chain a
single time.

The primary/secondary hostinfo chain and the `MaxHostInfosPerVpnIp` cap were added in
#820 specifically to stop hostinfos leaking on re-handshake. The chain contract is
unchanged here; the bug was introduced in #1216 when the single-vpnIp delete was
adapted to per-vpnAddr V2 certs and began unlinking the shared chain once per address.
This restores "unlink exactly once" while keeping #820's guarantees: dead hostinfos are
still fully removed, live siblings are still promoted, and `MaxHostInfosPerVpnIp` still
bounds growth. The behavioral change is scoped strictly to deleting a hostinfo that is
primary for two or more of its own vpnAddrs.

### DIFF
--- a/hostmap.go
+++ b/hostmap.go
@@ -438,43 +438,29 @@ func (hm *HostMap) unlockedMakePrimary(hostinfo *HostInfo) {
 }
 
 func (hm *HostMap) unlockedDeleteHostInfo(hostinfo *HostInfo) {
+	isLastHostinfo := hostinfo.next == nil && hostinfo.prev == nil
+
 	for _, addr := range hostinfo.vpnAddrs {
-		h := hm.Hosts[addr]
-		for h != nil {
-			if h == hostinfo {
-				hm.unlockedInnerDeleteHostInfo(h, addr)
-			}
-			h = h.next
+		if hm.Hosts[addr] != hostinfo {
+			continue
+		}
+		if hostinfo.next != nil {
+			// Promote the next hostinfo in the shared chain to primary for this address
+			hm.Hosts[addr] = hostinfo.next
+		} else {
+			delete(hm.Hosts, addr)
 		}
 	}
-}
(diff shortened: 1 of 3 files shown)


jrwren (approved):

hostmap.go
```diff
@@ -438,43 +438,29 @@ func (hm *HostMap) unlockedMakePrimary(hostinfo *HostInfo) {
...
-	} else {
-		// Relink if we were in the middle of multiple hostinfos for this vpn addr
-		if hostinfo.prev != nil {
-			hostinfo.prev.next = hostinfo.next
+		} else {
+			delete(hm.Hosts, addr)
 		}
+	}
+	if len(hm.Hosts) == 0 {
+		hm.Hosts = map[netip.Addr]*HostInfo{}
```
JackDoan, line 455: why?

nbrownus: It's an old thing, frees the underlying map memory if it grew really large and is now 0

JackDoan: I guess so. I was thinkin if you talked to a lot of hosts before, you'll probably talk to them again, so re-used the memory would be nice. Def not a hill I wanna die on though.

JackDoan (approved):

PR #1789, opened 2026-07-08
fix firewall port/proto bypass in parseV6 from uint8 extension-header length overflow

nbrownus:
Related to #1784


JackDoan (approved):
I hate extension headers

PR #1790, opened 2026-07-08
hostmap: replace the shared next/prev hostinfo chain with independent per-address lists so divergent or overlapping vpnAddr sets cannot corrupt the map

nbrownus:
A hostinfo has a single next/prev pointer pair but conceptually belongs to one list per vpn address it owns, so the shared chain corrupts whenever hostinfos for the same peer carry diverging or overlapping address sets. That flaw produced #1788 and two more latent bugs on the add and makePrimary paths. This replaces the chain with independent per-address lists: Hosts stays map[addr]*HostInfo holding the primary, so the packet hot path is untouched, and a new internal moreHosts map only carries an entry while an address is held by two or more hostinfos, most recent first. Benchmarked against master, hot path lookups and steady state memory are identical, with a small transient cost per address only while duplicate tunnels exist. Along the way it fixes a latent bug where deleting the same hostinfo twice could wipe lighthouse state while a live sibling tunnel still held the address, stops makePrimary from resurrecting a hostinfo that lost a race with teardown, and adds tests that assert the list invariants after every mutation.

### DIFF
--- a/handshake_manager.go
+++ b/handshake_manager.go
@@ -430,14 +430,11 @@ func (hm *HandshakeManager) CheckAndComplete(hostinfo *HostInfo, handshakePacket
 	// Check if we already have a tunnel with this vpn ip
 	existingHostInfo, found := hm.mainHostMap.Hosts[hostinfo.vpnAddrs[0]]
 	if found && existingHostInfo != nil {
-		testHostInfo := existingHostInfo
-		for testHostInfo != nil {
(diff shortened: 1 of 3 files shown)


jrwren (commented):
😵‍💫 🤯

jrwren (approved):

JackDoan (approved):
After much hand-wringing I am forced to conclude this is the best way to do this.

Cope: no more linked list!

PR #1794, opened 2026-07-09
Make Control safe to stop and wait on from any lifecycle state

nbrownus:
This branch makes nebula's `Control` safe to drive as an embedded library through its whole lifecycle. `Stop` is now callable from any state (a never-started `Control` releases everything `Main` acquired and poisons itself so `Start` returns `ErrAlreadyStopped`), a new `Control.Wait` blocks until nebula has fully stopped and returns the first fatal reader error, `Interface.Close` is idempotent, and failed activation or `Main` errors now release the udp sockets and tun fd instead of leaking them. `RebindUDPServer` is gated on run state so a network-change handler racing shutdown can't touch a closed socket, and the iOS tun fd is set non-blocking so shutdown can't hang.

### DIFF
--- a/cmd/nebula-service/service.go
+++ b/cmd/nebula-service/service.go
@@ -67,6 +67,9 @@ func (p *program) Stop(s service.Service) error {
 	}
 
 	p.control.Stop()
+
+	// block until nebula has fully drained before reporting stopped.
+	// error logging is handled by Start.
 	_ = p.control.Wait()
 	return nil
 }
--- a/cmd/nebula-service/service.go
+++ b/cmd/nebula-service/service.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"log"
+	"os"
 
 	"github.com/kardianos/service"
 	"github.com/slackhq/nebula"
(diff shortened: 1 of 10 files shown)


wadey (dismissed):

cmd/nebula-service/service.go
```diff
@@ -40,22 +40,38 @@ func (p *program) Start(s service.Service) error {
...
 }
 
 func (p *program) Stop(s service.Service) error {
 	logger.Info("Nebula service stopping.")
+	if p.control == nil {
+		return nil
+	}
+
 	p.control.Stop()
+	_ = p.control.Wait()
```
jrwren, line 73: Should the error from Wait be returned here?

If not, maybe include a comment as to why not?

jrwren (dismissed):
LGTM. One thought, take it or leave it.

cmd/nebula-service/service.go
nbrownus: Added a comment. Basically this is kardianos api forcing our hand (Stop returns an error) but we want to know if nebula exits without having called Stop and that thing wants to log an error. If we log from both the catch-all and Stop then we get 2 log statements. Doing it this way lets us avoid knowing if we called Stop vs nebula fatal'd to change where the log output happens.

JackDoan (approved):

PR #1797, opened 2026-07-11
warn on uselessly low MTU

JackDoan:


examples/config.yml
```diff
@@ -110,6 +110,15 @@ lighthouse:
...
     #- "1.2.3.4:0" # port will be replaced with the real listening port
 
+  # Locally discovered addresses are checked against the MTU of the link they were found on. If the link cannot fit
+  # a full-size packet from the nebula tun device (`tun.mtu` plus encapsulation overhead, which is larger for relayed
+  # traffic) without fragmenting, a warning is logged.
+  # When omit_low_mtu_addrs is true, addresses whose links cannot fit normal nebula traffic are dropped from
+  # lighthouse reports entirely.
+  # Addresses that can fit normal nebula traffic but not relayed traffic are always still advertised.
+  # This does not apply to addresses listed in advertise_addrs.
+  #omit_low_mtu_addrs: false
```
nbrownus, line 120: Default to true?

hostmap.go
```diff
@@ -869,9 +869,17 @@ func (i *HostInfo) logger(l *slog.Logger) *slog.Logger {
...
+// localAddr is a locally discovered address candidate for lighthouse
+// advertisement, along with details about the link it was found on.
+type localAddr struct {
+	addr    netip.Addr
+	ifName  string
+	linkMTU int // MTU reported for the link, or <= 0 if unknown
+}
+
+func localAddrs(l *slog.Logger, allowList *LocalAllowList) []localAddr {
 	//FIXME: This function is pretty garbage
```
nbrownus, line 881: <3

lighthouse.go
```diff
@@ -905,6 +938,108 @@ func (lh *LightHouse) TriggerUpdate() {
...
+	// The link can carry normal nebula traffic, but relayed traffic (which
+	// adds a second layer of encapsulation) will not fit
+	linkMTUTooSmallForRelay
+	// Even normal nebula traffic will not fit
+	linkMTUTooSmall
+)
+
+const (
+	// Both AES-256-GCM and ChaCha20-Poly1305 append a 16 byte AEAD tag
+	cipherTagLen = 16
```
nbrownus, line 965: We should get this from the current cipher, will be slightly annoying to thread it through though. Keeps us from the future footgun.

lighthouse.go
```diff
@@ -905,6 +938,108 @@ func (lh *LightHouse) TriggerUpdate() {
...
+	}
+
+	direct = tunMTU + header.Len + cipherTagLen + udpHeaderLen + ipHeaderLen
+	relayed = direct + header.Len + cipherTagLen
+	return direct, relayed
+}
+
+// checkLocalLinkMTU classifies e's link MTU, logs when the classification
+// changes, and reports whether e should be advertised to lighthouses.
+func (lh *LightHouse) checkLocalLinkMTU(e localAddr) bool {
```
nbrownus, line 986: Feels like all of this can move to its own package and relieve lighthouse.go of carrying it along.

examples/config.yml
JackDoan: technically a breaking change, figured we could at least give some warning for a release or two

lighthouse.go
JackDoan: where do you wanna draw the line? I agree but its super blurry

examples/config.yml
nbrownus: Fair

lighthouse.go
nbrownus: The stateless stuff. Feels like `header` should have the bits that matter for the header + relay, cipher config has the tag len (using a const there is future pain if we can avoid it now), the bits with state should stay in lighthouse.go, the rest can live somewhere else.

Like the stuff in hostmap.go, no reason for it, just never moved it anywhere else.

PR #1801, opened 2026-07-14
Add <nebula> self-token for bind addresses

johnmaguire:
Problem: It is not possible to listen _only_ on _every_ Nebula IP. You can listen on `0.0.0.0` or `[::]` but this will also listen publicly. You can listen on a specific Nebula IP, but Nebula does not support listing multiple.

Solution:

Introduce a "<nebula>" host token usable in the host position of a listener
config value (e.g. `stats.listen: "<nebula>:8080"`). It expands to every one of
this host's overlay/VPN addresses, binding all of them (one listener each).

"<" and ">" are illegal in hostnames, so the token is safe by construction: it
is always intercepted and expanded before any name resolution, and a call site
that ever forgot to expand it would fail loud rather than silently resolve.

This commit adds the shared resolveSelfListenAddrs helper and wires the first
consumer, stats.listen (Prometheus), refactoring it to serve one listener per
expanded address. Non-token values are unchanged.

Caveats:

- A cert update with new addresses will not reload listeners. I believe this is best solved as a followup PR.
- A tun.disabled host will fail to listen on its nebula IPs - so log warnings if the token is used with them

Explicitly excluded:
- Indexed reference to IPs (e.g. <nebula[0]>) - these could appear in any order, and are not stable
- IPv4 / IPv6 (e.g. <nebula.ipv4>) only listens: this could make sense as a followup.

Alternatives:

- Change the listen fields to accept string or array of string. In the case of array, listen on all. (This would actually be more flexible, but requires the config writer to set the config based on the addresses in the cert.)

### DIFF
--- a/self_listen.go
+++ b/self_listen.go
@@ -2,9 +2,12 @@ package nebula
 
 import (
 	"fmt"
+	"log/slog"
 	"net"
 	"net/netip"
 	"strings"
+
+	"github.com/slackhq/nebula/config"
 )
 
 // nebulaSelfToken is a magic host value usable in the host position of a
@@ -84,4 +87,21 @@ func (p *PKI) vpnAddrs() []netip.Addr {
 		return nil
 	}
 	return cs.myVpnAddrs
+}
+
+// warnSelfTokenWithTunDisabled logs a warning for each listener whose config
+// uses the "<nebula>" self-token while tun.disabled is set. The token expands
+// to this host's overlay addresses, which no interface carries when the tun is
+// disabled, so those listeners cannot bind. The common footgun is enabling DNS
+// with "<nebula>" on a tun-disabled lighthouse.
+func warnSelfTokenWithTunDisabled(l *slog.Logger, c *config.C) {
+	if !c.GetBool("tun.disabled", false) {
+		return
+	}
+	for _, key := range []string{"lighthouse.dns.host", "sshd.listen", "stats.listen"} {
+		if strings.Contains(c.GetString(key, ""), nebulaSelfToken) {
+			l.Warn(`Listener uses the "<nebula>" self-token but tun.disabled is set; overlay addresses are not bindable without a tun interface, so this listener will fail to bind`,
+				"configKey", key)
+		}
+	}
 }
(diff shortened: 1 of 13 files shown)


self_listen.go
```diff
@@ -0,0 +1,107 @@
...
+		addrs[i] = net.JoinHostPort(a.Unmap().String(), port)
+	}
+	return addrs, nil
+}
+
+// vpnAddrs returns this host's overlay/VPN addresses for expanding the
+// "<nebula>" self-token, or nil when the PKI or its cert state isn't available
+// yet. Nil-safe on the receiver so every listener call site (and tests that
+// build a server without a PKI) can call it without its own guard.
+func (p *PKI) vpnAddrs() []netip.Addr {
```
JackDoan: kinda weird having this so far from the PKI struct.

Is nil on no-pki what you want?

stats.go
```diff
@@ -186,6 +188,27 @@ func (s *statsServer) Start() {
...
+// non-token address takes the byte-identical legacy ListenAndServe path; the
+// token expands to one listener per VPN address, all served by the same server.
+func (s *statsServer) servePrometheus(cfg statsConfig, srv *http.Server) bool {
+	addrs, err := resolveSelfListenAddrs(cfg.prom.listen, s.pki.vpnAddrs())
+	if err != nil {
+		s.l.Error("Failed to expand prometheus stats listen address", "error", err)
+		return false
+	}
+
+	// Preserve today's exact behavior for ordinary configs: one server, one
```
JackDoan: tbh I'd rather just have one path

stats.go
```diff
@@ -216,6 +239,65 @@ func (s *statsServer) serveListener(listener *http.Server) bool {
...
+			}
+			return false
+		}
+		listeners = append(listeners, ln)
+	}
+
+	// Per-invocation watcher: ctx cancellation triggers a server shutdown that
+	// closes all served listeners. Closing `done` on exit keeps the watcher
+	// from outliving this call.
+	done := make(chan struct{})
```
JackDoan, line 228: is this recyclable across all 3 of our kinds of "extra server"?

tbh I expect no :/

stats.go
johnmaguire: already done!

brad-defined (approved):

brad-defined (changes requested):
mis-clicked the approve on this one. ignore me.

PR #1803, opened 2026-07-14
Support multiple listen addresses in stats, sshd, and dns

johnmaguire:
Problem: It is not possible to listen _only_ on _every_ Nebula IP. You can listen on 0.0.0.0 or [::] but this will also listen publicly. You can listen on a specific Nebula IP, but Nebula does not currently support listing multiple.

Solution:

stats.listen now accepts either a single "host:port" string (as before) or a
list of them, binding one Prometheus listener per address. This lets a node
serve stats on several addresses, e.g. each of its nebula IPs, without a
wildcard bind.

Adds the shared getListenAddrs helper (string-or-list, backwards compatible)
and refactors the prometheus serving path to bind every configured address up
front and serve them from one http.Server.

Caveats:

- You must be careful to update the IPs in the config if the IPs in the cert change

Alternatives:

- Special `<nebula>` listener for Nebula IPs only - this solution offers less flexibility (see #1801)

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -50,6 +50,8 @@ lighthouse:
   #serve_dns: false
   #dns:
     # The DNS host defines the IP to bind the dns listener to. This also allows binding to the nebula node IP.
+    # host may also be a list of IPs to bind the listener to several addresses (e.g. this node's nebula IPs),
+    # each combined with the port below.
     #host: 0.0.0.0
     #port: 53
   # interval is the number of seconds between updates from this node to a lighthouse.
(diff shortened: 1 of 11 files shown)


examples/config.yml
```diff
@@ -50,6 +50,8 @@ lighthouse:
   #serve_dns: false
   #dns:
     # The DNS host defines the IP to bind the dns listener to. This also allows binding to the nebula node IP.
+    # host may also be a list of IPs to bind the listener to several addresses (e.g. this node's nebula IPs),
```
JackDoan, line 53: if we're changing semantics, it feels like a great time to standardize on ONE way to set an IP+port :D

`listeners: ["something that netip.ParseAddrPort will accept"]` ?

PR #1805, opened 2026-07-16
Fix relay re-establishment for handshake on Disestablised entry

johnmaguire:
`handleOutsideRelayPacket` filled `ViaSender.remoteIdx` with `relay.RemoteIndex`,
an index from the relay peer's index space, but the rescue in
`sendHandshakeResponse` looks that value up in `relayForByIdx`, which is keyed
by local index. The lookup could never hit, so a terminal relay entry left
`Disestablished` by a one-sided teardown stayed `Disestablished` even after a
valid handshake arrived over it. The responder's first transmit then failed
to find an `Established` relay, deleted its only relay entry, and every
subsequent send was silently dropped until dead-tunnel detection forced a
re-handshake.

### DIFF
--- a/outside.go
+++ b/outside.go
@@ -214,7 +214,6 @@ func (f *Interface) handleOutsideRelayPacket(hostinfo *HostInfo, via ViaSender, 
 		via = ViaSender{
 			UdpAddr:   via.UdpAddr,
 			relayHI:   hostinfo,
-			remoteIdx: relay.RemoteIndex,
 			relay:     relay,
 			IsRelayed: true,
 		}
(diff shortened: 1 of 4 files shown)


nbrownus (approved):

outside.go
```diff
@@ -214,7 +214,6 @@ func (f *Interface) handleOutsideRelayPacket(hostinfo *HostInfo, via ViaSender,
 		via = ViaSender{
 			UdpAddr:   via.UdpAddr,
 			relayHI:   hostinfo,
-			remoteIdx: relay.RemoteIndex,
```
brad-defined, line 217: yeah, this doesn't match the comment of the field in the struct. It was supposed to be the idx from the packet, not the relay's remote index.

brad-defined (approved):

PR #1820, opened 2026-07-24
Fold the rebind counter and traffic flags into one atomic word

nbrownus:
Reduces the size of the `HostInfo` struct (shaved 24 bytes off on my computer) and fixes a known/accepted data race with `rebindCount`, which is more interesting now that we use it on more than just iOS.

The `HostInfo` reduction comes from collapsing several `atomic.Bool`s (`in`, `out`, `pendingDeletion`) into a single `atomic.Uint32` (which is what an `atomic.Bool` is under the hood) and reordering the struct to reduce padding.

`rebindCount` safety comes for free: the top 29 bits of that same `atomic.Uint32` now hold the rebind epoch the tunnel last sent under, replacing the old raw `uint8` `lastRebindCount`. Besides fixing the race, this closes the old wrap hole where an idle tunnel that landed on exactly a multiple of 256 rebinds would miss its requery. 29 bits is 2^29 rebinds, which we will never reach.

The rebind comparison now happens inside `connectionManager.Out` in the same atomic operation that records the send: it records outbound traffic and reports whether the epoch moved, and only then do we ask the lighthouse to get the far side punching at us again. Two paths deliberately record traffic without consuming that edge: relayed via-sends (`OutNoRebind`), where the direct path to the relay owns the epoch, and close, which skips the connection manager entirely since the tunnel is torn down right after.

One intentional behavior change: a tunnel formed after a rebind now stamps the current epoch when it is added to the hostmap, so it will not fire a redundant lighthouse query on its first packet. The handshake it just completed already primed the far side with our current address.

No change in throughput or CPU in my linux perf tests (a c8a.16xlarge server with 16 c8a.xlarge clients running iperf/reverse iperf for 5 minutes). I do expect some memory savings on highly connected hosts.


connection_manager.go
```diff
@@ -105,11 +105,18 @@ func (cm *connectionManager) getInactivityTimeout() time.Duration {
 }
 
 func (cm *connectionManager) In(h *HostInfo) {
-	h.in.Store(true)
+	h.markIn()
 }
 
-func (cm *connectionManager) Out(h *HostInfo) {
-	h.out.Store(true)
+// OutNoRebind records outbound traffic without consuming the rebind epoch, for relayed sends: the direct path
```
JackDoan, line 111: huh???

hostmap.go
```diff
@@ -770,6 +766,64 @@ func (i *HostInfo) TryPromoteBest(preferredRanges []netip.Prefix, ifce *Interfac
...
 }
 
+// Bits within HostInfo.state, everything above stateEpochShift is the epoch
+const (
+	stateIn uint32 = 1 << iota
+	stateOut
+	statePendingDeletion
+
+	stateFlags = stateIn | stateOut | statePendingDeletion
+	// The epoch is the top 29 bits, it would take 2^29 rebinds to wrap and we will never get there
```
JackDoan, line 776: it would not cost very much to have an atomic.uint32 for bits, and keep an atomic.uint32 for epoch-countery stuff.

Mixing flags and numbers in the same bit-space when it's not strictly required is never fun

connection_manager.go
nbrownus: Some outbound packets shouldn't check if it needs to perform any actions as a result of a rebind. Relayed packets are currently the only one (previously close tunnel packets also needed it but I yanked that path out).

If closing the tunnel triggers a post rebind activity it will cause a bunch of work for a tunnel that's being destroyed
If a relayed packet triggers a post rebind activity it will similarly do silly things on behalf of a tunnel that isn't directly connected or attempting a direct connection.

JackDoan (approved):

hostmap.go
JackDoan: counterpoint: this is one less atomic op per send on the hot path :D

PR #1821, opened 2026-07-24
Add info_api for local identity lookups

JackDoan:
Programs running alongside nebula have no simple way to ask "who is this vpn address?" when making authorization decisions, e.g. a nebula-aware webapp that wants to identify an inbound connection by its source address instead of presenting a login form. The existing surfaces are the sshd admin interface (no one wants to integrate with SSH) and the lighthouse-only DNS TXT lookup, which returns raw cert JSON over an awkward transport.

This adds an opt-in `host_query` config section that serves a small HTTP+JSON API on a unix socket or tcp address, requiring no client library to consume:

  GET /v1/host?addr=<vpn addr>  identity of the host owning the address
                                (an established peer, or this node).
                                addr may include a port so a server can
                                pass a connection's RemoteAddr through
                                unparsed.
  GET /v1/self                  this node's own identity.

Responses carry the certificate-derived identity only: name, vpn addresses, networks, unsafe networks, groups, fingerprint, issuer, validity window, and cert version.

Binds to a local UNIX socket with http as a transport, similar to how Docker's CLI works

### DIFF
--- a/examples/config.yml
+++ b/examples/config.yml
@@ -231,6 +231,35 @@ punchy:
   # Overriding this to "" is the same as "/" and will allow overwriting any path on the host.
   #sandbox_dir: /var/tmp/nebula-debug
 
+# EXPERIMENTAL: this feature may change or disappear in the future.
+# info_api exposes a small local HTTP+JSON API that lets other programs on
+# this machine resolve a vpn address to its certificate identity (name, vpn
+# addresses, groups, fingerprint, validity), e.g. for making authorization
+# decisions about an inbound connection:
+#   GET /v1/host?addr=<vpn addr>  - identity of the host owning the address: a
+#     peer with an active tunnel, or this node itself. `addr` may include a
+#     port (`192.168.100.7:54321`), which is ignored, so a connection's remote
+#     address can be passed through as is. Returns 404 when the address is
+#     unknown or has no active tunnel.
+#   GET /v1/self                  - this node's own identity.
+# Identity answers can be trusted because nebula drops inbound packets whose
+# source vpn address is not contained in the sender's certificate, so the
+# source address of a connection arriving over the nebula interface is
+# guaranteed to map to the certificate reported here.
+# There is no authentication in this API; restrict access with unix socket
+# file permissions.
+# This whole section is reloadable.
+#info_api:
+  # Toggles the feature
+  #enabled: false
+  # listen accepts a unix socket path as a unix:// URL with an absolute path:
+  #listen: unix:///var/run/nebula-info-api.sock
+  # File mode for the unix socket, as an octal string.
+  # The socket is created by nebula's user; to grant a group of local services
+  # access, place the socket in a directory with appropriate permissions
+  # (e.g. a systemd RuntimeDirectory) and relax this to "0660".
+  #socket_mode: "0600"
+
 # EXPERIMENTAL: relay support for networks that can't establish direct connections.
 relay:
   # Relays are a list of Nebula IP's that peers can use to relay packets to me.
--- a/info_api_server.go
+++ b/info_api_server.go
@@ -1,0 +1,409 @@
+package nebula
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/fs"
+	"log/slog"
+	"net"
+	"net/http"
+	"net/netip"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/slackhq/nebula/cert"
+	"github.com/slackhq/nebula/config"
+)
+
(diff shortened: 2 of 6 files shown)


examples/config.yml
```diff
@@ -231,6 +231,35 @@ punchy:
...
+#     unknown or has no active tunnel.
+#   GET /v1/self                  - this node's own identity.
+# Identity answers can be trusted because nebula drops inbound packets whose
+# source vpn address is not contained in the sender's certificate, so the
+# source address of a connection arriving over the nebula interface is
+# guaranteed to map to the certificate reported here.
+# There is no authentication in this API; restrict access with unix socket
+# file permissions.
+# This whole section is reloadable.
+#info_api:
```
matt-defined, line 252: I don't want to get hung up bike-shedding, but maybe just `api`?

info_api_server.go
```diff
@@ -0,0 +1,409 @@
...
+// programs on this machine resolve a vpn address to its certificate identity (name, groups, networks)
+// for making authorization decisions. Lifecycle works like statsServer: the constructor wires the
+// reload callback, reload records config, Start runs the runtime, Stop tears it down
+type infoAPIServer struct {
+	l       *slog.Logger
+	ctx     context.Context
+	hostMap *HostMap
+	pki     *PKI
+
+	// enabled mirrors `info_api.enabled` so callers of Start don't need to know the gating rules
```
matt-defined, line 35: This comment makes me confused 😂

info_api_server.go
```diff
@@ -0,0 +1,409 @@
...
+	h.runMu.Unlock()
+
+	h.enabled.Store(newCfg.enabled)
+
+	if initial || sameCfg {
+		return nil
+	}
+
+	if running {
+		h.Stop()
```
matt-defined, line 107: I think this can end up hanging the reconfig for multiple seconds. It's probably okay for `Stop` to `Close` instead of `Shutdown`?

info_api_server.go
```diff
@@ -0,0 +1,409 @@
...
+// listen binds the configured unix socket. It also clears a stale socket file left by an unclean
+// exit and applies the configured file mode
+func (h *infoAPIServer) listen(cfg infoAPIConfig) (net.Listener, error) {
+	if fi, err := os.Stat(cfg.addr); err == nil {
+		if fi.Mode()&os.ModeSocket == 0 {
+			return nil, fmt.Errorf("info_api.listen path %s exists and is not a socket, refusing to replace it", cfg.addr)
+		}
+		// a normal shutdown unlinks the socket, so a file here means a previous process exited
+		// uncleanly, remove it so the bind below can succeed
+		if err = os.Remove(cfg.addr); err != nil {
```
matt-defined, line 218: Before removing, we could check to see if anybody's listening on it. Instead of a crash, it might be a misconfiguration where two instances of nebula are pointed at the same location.

info_api_server.go
```diff
@@ -0,0 +1,409 @@
...
+		case <-h.ctx.Done():
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			if err := srv.Shutdown(shutdownCtx); err != nil {
+				h.l.Warn("Failed to shut down info API listener", "error", err)
+			}
+		case <-done:
+		}
+	}()
+	defer close(done)
```
matt-defined, line 183: Maybe put this above the go routine, just so it's closer visually?

info_api_server.go
```diff
@@ -0,0 +1,409 @@
...
+		}
+	}
+
+	ln, err := net.Listen("unix", cfg.addr)
+	if err != nil {
+		return nil, err
+	}
+	// The socket is briefly live with umask-derived permissions before this chmod lands, tolerated
+	// because connections accepted in that window still only reach this read-only API
+	if err = os.Chmod(cfg.addr, cfg.socketMode); err != nil {
```
matt-defined, line 229: We could umask around it, to close the race.

info_api_server.go
```diff
@@ -0,0 +1,409 @@
...
+		return cfg, fmt.Errorf("info_api.socket_mode was not a valid octal file mode: %s", modeStr)
+	}
+	cfg.socketMode = fs.FileMode(mode)
+	return cfg, nil
+}
+
+// parseInfoAPIListen extracts the unix socket path from the info_api.listen config value, which
+// must be a `unix://` URL with an absolute path, e.g. `unix:///var/run/nebula.sock`
+func parseInfoAPIListen(listen string) (addr string, err error) {
+	path, ok := strings.CutPrefix(listen, "unix://")
```
matt-defined, line 402: Should we allow upper case?

examples/config.yml
wadey: I think we talked about maybe having an "admin" api eventually too, exposed differently than this readonly API.

PR #1826, opened 2026-07-27
Don't fail on the batch at the first error

nbrownus:
Rather than pick an error to return, `WriteBatch` now returns the count of packets that actually went out, mirroring `sendmmsg` itself. An error means the call failed (bad arguments, no progress); a short count means specific destinations were undeliverable.

### DIFF
--- a/udp/udp_tester.go
+++ b/udp/udp_tester.go
@@ -171,13 +171,18 @@ func (u *TesterConn) WriteTo(b []byte, addr netip.AddrPort) error {
 		return nil
 	}
 }
-func (u *TesterConn) WriteBatch(bufs [][]byte, addrs []netip.AddrPort, _ []byte) error {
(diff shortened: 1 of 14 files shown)


udp/udp_tester.go
```diff
@@ -171,13 +171,18 @@ func (u *TesterConn) WriteTo(b []byte, addr netip.AddrPort) error {
 		return nil
 	}
 }
-func (u *TesterConn) WriteBatch(bufs [][]byte, addrs []netip.AddrPort, _ []byte) error {
+func (u *TesterConn) WriteBatch(bufs [][]byte, addrs []netip.AddrPort, _ []byte) (int, error) {
+	// An un-sendable destination costs its own packet, never the ones behind it in the batch. This mirrors the
+	// real backends on purpose: e2e runs on TesterConn, so a divergence here hides the production behavior.
```
JackDoan:
```suggestion
```

JackDoan (approved):

PR #1832, opened 2026-07-28
Drop dependency on github.com/cyberdelia/go-metrics-graphite

### LINKED ISSUE #1831: (Re-)consider dependency on github.com/cyberdelia/go-metrics-graphite
Hi!  I help maintain this in Debian.  There is a dependency on https://github.com/cyberdelia/go-metrics-graphite but that package was last updated 10+ years ago.  How important is that package for Nebula?  Relying on old practically unmaintained packages is a bit of a packaging and supply-chain concern.

See also: https://github.com/cyberdelia/go-metrics-graphite/issues/17

Thanks!

jrwren:
## Summary

Removes the dependency on `github.com/cyberdelia/go-metrics-graphite`, which has been unmaintained for 10+ years and is a packaging / supply-chain concern for downstream distributors (Debian, Guix). See #1831.

Nebula only used two symbols from that ~120-line package: the `Config` struct and the `Once()` function. Rather than keep a dead dependency, this inlines just those parts (~65 lines) into a local `graphite.go`, preserving the upstream BSD-2-Clause copyright notice as the file requires.

## Changes

- **`graphite.go`** (new) — inlined `graphiteConfigExport` and `graphiteOnce`, with the upstream license header retained.
- **`stats.go`** — use the local types instead of the imported package.
- **`stats_test.go`** — update a stale comment.
- **`go.mod` / `go.sum`** — dropped via `go mod tidy`.

## Testing

- `go build ./...` and `go vet` pass.
- The graphite export path is exercised by `TestStatsServer_reload_initial_graphite` and `TestStatsServer_Start_graphite_blocksUntilStop` (the latter drives a real TCP sink through `graphiteOnce`), both passing.

Fixes #1831

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,6 @@ require (
 	filippo.io/bigmod v0.1.0
 	github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be
 	github.com/armon/go-radix v1.0.0
-	github.com/cyberdelia/go-metrics-graphite v0.0.0-20161219230853-39f87cc3b432
 	github.com/flynn/noise v1.1.0
 	github.com/gaissmai/bart v0.28.0
 	github.com/gogo/protobuf v1.3.2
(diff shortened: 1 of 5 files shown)


jas4711:
Thanks for this -- any chance it could be reviewed and merged?

wadey (approved):

PR #1598, opened 2026-02-04
tests: Use less Argon2 memory to fix tests on 32-bit systems (#1597)

jas4711:
=== RUN   Test_ca
runtime: out of memory: cannot allocate 2147483648-byte block (2151251968 in use)
fatal error: out of memory
goroutine 22 gp=0x98870e8 m=3 mp=0x9845008 [running]:
...
github.com/slackhq/nebula/cmd/nebula-cert.Test_ca(0x9887328)
	/build/package/package/obj-i686-linux-gnu/src/github.com/slackhq/nebula/cmd/nebula-cert/ca_test.go:181 +0x2102 fp=0x9923f84 sp=0x9923e04 pc=0x8330ce2

Issue #1597, opened 2026-02-04
🐛 BUG: armhf/i386: Test_ca: runtime: out of memory: cannot allocate 2147483648-byte block (2151022592 in use)


jas4711:
### What version of `nebula` are you using? (`nebula -version`)

1.10.2

### What operating system are you using?

Debian

### Describe the Bug

Hi!  I'm updating this package for Debian (we could use testers to get confidence in newer packages, let me know how to reach people interested in this).  On armhf/i386 we got this failure:

```
=== RUN   Test_ca
runtime: out of memory: cannot allocate 2147483648-byte block (2151022592 in use)
fatal error: out of memory
```

See logs here:

https://buildd.debian.org/status/fetch.php?pkg=nebula&arch=armhf&ver=1.10.2%2Bdfsg-1%7Eexp0&stamp=1770158339&raw=0

https://buildd.debian.org/status/fetch.php?pkg=nebula&arch=i386&ver=1.10.2%2Bdfsg-1%7Eexp0&stamp=1770158639&raw=0

Any ideas?

We'll probably just skip this test on armhf/i386 unless there is some other ideas.

Thanks,
Simon

### Logs from affected hosts

```

```

### Config files from affected hosts

```

```

jas4711:
This line is suspicious:

https://github.com/slackhq/nebula/blob/master/cmd/nebula-cert/ca_test.go#L191

The value in the crash backtrace 2147483648 is identical to `2*1024*1024*1024`, although the code says `2*1024*1024`.

Maybe there is an additional unintended implicit `*1024`?

I am trying a rebuild on i386 changing the `2*1024*1024` into merely `2*1024` now.

jas4711:
Even more suspicious line:

https://github.com/slackhq/nebula/blob/master/cmd/nebula-cert/ca.go#L58

```
	cf.argonMemory = cf.set.Uint("argon-memory", 2*1024*1024, "Optional: Argon2 memory parameter (in KiB) used for encrypted private key passphrase")
```

Allocating 2GB will not work well on 32-bit archs.  I don't want to suggest to lower this number, as that seems like a horrible solution.

jas4711:
Debian is now using https://github.com/jas4711/nebula/commit/595e9ee8413f94cf4b13e44842f88234d4be243c to resolve this problem -- it lowers the memory usage for the self-tests only.  I've confirmed Nebula 1.10.2 now builds and passes self tests on i386, more archs to follow.

[...]

Issue #1387, opened 2025-04-12
🐛 BUG: argon2 way too much memory ?


kapouer:
### What version of `nebula` are you using? (`nebula -version`)

1.9.3

### What operating system are you using?

Linux debian/trixie

### Describe the Bug

I get oom error when running the tests:
https://buildd.debian.org/status/fetch.php?pkg=nebula&arch=armhf&ver=1.9.3%2Bdfsg-2&stamp=1744458470&raw=0

it seems 2 * 1024 * 1024 is the default for argon2 memory parameter.
That is in kilobites, so it's a bit too much ? 2GB ?

### Logs from affected hosts

```

```

### Config files from affected hosts

```

```

johnmaguire:
I'm sorry, I misunderstood the issue. I will update the test to use the second recommendation (64 MiB, 3 iterations, 4 lanes.)

kapouer:
Thanks ! From a distribution point of view, it will make it much easier.

PR #1834, opened 2026-07-31
cmd/ca: prevent out-of-memory on 32bit systems

realsn0w:
This is another attempt at fixing #1597 whose original fix (#1598) went stale. 

If 32bit is detected it conditionally alters the argon default params to RFC9106's recommendation for memory-constrained systems, similar to how #1387 was fixed.


realsn0w:
CLA has been signed but the bot doesn't seem to notice

wadey (approved):
thank you!

PR #1839, opened 2026-08-13
Firewall bypass hardening: reject disjoint-network peers & fix test h…

BOSS0exe:
…arness panic

- handshake_manager: new handshakes.reject_on_no_common_networks option; peers whose certificate networks share no subnet with ours are now refused on both the responder (beginHandshake) and initiator (stage-2 completion) paths instead of completing a tunnel with nothing in common. Completing such handshakes by default wastes tunnel state and lets any CA-signed peer reach hosts that only overlap with other subnets, contrary to the intended firewall posture.
- CloseTunnel / tunnel-establishment bypass: new e2e reproducers prove that tunnel establishment and CloseTunnel messages are never evaluated against the configured firewall policy; document the gap in the PR body for maintainer decision.
- e2e/router: formatUdpPacket hardened against malformed/tun packets (use ok-form type assertions) so continuous routing loops no longer panic the flow renderer.
- e2e tests: reject_on_no_common_networks verifier tests, firewall-bypass reproducers, and an unauthenticated handshake-flood resource-exhaustion reproducer (hostmap grows with every distinct valid-cert peer; there is no per-source rate limit on the responder path).

No regressions: the full e2e suite passes.

### DIFF
--- a/e2e/close_tunnel_fw_bypass_test.go
+++ b/e2e/close_tunnel_fw_bypass_test.go
@@ -1,0 +1,236 @@
+//go:build e2e_testing
+// +build e2e_testing
+
+// Reproducer for: CloseTunnel tunnel teardown bypasses firewall policy, and
+// tunnel establishment itself ignores firewall rules.
+//
+// A (server, 10.128.0.1/24) configures its firewall to DROP ALL INBOUND traffic
+// (empty inbound rules), only allowing outbound. B (10.128.0.2/24) sends data
+// first to force a handshake.
+//
+// Evidence:
+//  1. The handshake completes successfully even though the inbound firewall
+//     would drop every data packet from B (outside.go:150 HandleRequest /
+//     handshake_manager.go HandleIncoming never consult the firewall;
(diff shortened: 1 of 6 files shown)


JackDoan:
>handshake_manager: new handshakes.reject_on_no_common_networks option; peers whose certificate networks share no subnet with ours are now refused on both the responder (beginHandshake) and initiator (stage-2 completion) paths instead of completing a tunnel with nothing in common. Completing such handshakes by default wastes tunnel state and lets any CA-signed peer reach hosts that only overlap with other subnets, contrary to the intended firewall posture.

The firewall and the handshake mechanism are intentionally decoupled. In Nebula, host discovery, hole punching, handshakes, tunnel maintenance, etc, are modelled as "layer 2" of the OSI model. The firewall only interacts with "layer 3" (IP traffic).

Allowing tunnels to form between disjoint subnets has useful properties. For instance, you can make use of lighthouses and relays, without concerns over allowing L3 traffic from those lighthouses/relays to bleed into your network.

PR #1840, opened 2026-08-17
parseV6: only walk real ipv6 extension headers, fail closed on unknown protocols

nbrownus:
`parseV6` walked any Next Header value it didn't have an explicit case for as if it were an ipv6 extension header, applying the (len+1)*8 length formula to it. For a real upper layer protocol we don't dissect (SCTP, GRE, IP-in-IP, etc.) that marched into the transport payload and could land on a byte that looks like TCP or UDP, producing a bogus protocol/port that the firewall then trusted, while the host delivered the real, different packet.

This switches the walk to an allow list. We only walk the actual extension headers (Hop-by-Hop, Routing, Destination Options, plus the existing Fragment and AH cases), which matches the set the Linux kernel skips in `ipv6_ext_hdr`. Anything else is treated as terminal: we record its true protocol with no ports and stop, so it can only ever match an any rule, the same way we already handle ESP and NoNextHeader. Truncated extension headers now return an error instead of stumbling forward.

### DIFF
--- a/iputil/packet.go
+++ b/iputil/packet.go
@@ -381,6 +381,11 @@ func IPv6FindUpperProtocol(packet []byte) (nextHeader uint8, offset int, isFragm
 			offset += (int(packet[offset+1]) + 2) << 2
 
 		default:
+			// A prior extension header can declare a length that advances offset past the packet. The terminal
+			// protocol's header isn't actually here, so treat the chain as truncated rather than classifying it.
+			if offset > len(packet) {
+				return nextHeader, offset, isFragment, ErrIPv6CouldNotFindPayload
+			}
 			return nextHeader, offset, isFragment, nil
 		}
 	}
--- a/iputil/packet.go
+++ b/iputil/packet.go
@@ -2,10 +2,15 @@ package iputil
 
 import (
 	"encoding/binary"
+	"errors"
 
 	"golang.org/x/net/ipv4"
 	"golang.org/x/net/ipv6"
 )
+
+// ErrIPv6CouldNotFindPayload is returned when the ipv6 extension header chain is truncated before a terminal
+// upper layer protocol is reached.
+var ErrIPv6CouldNotFindPayload = errors.New("could not find payload in ipv6 packet")
 
 const (
 	// MaxIPv4RejectPacketSize is the largest IPv4 reject packet:
@@ -199,8 +204,8 @@ func ipv4CreateRejectTCPPacket(packet []byte, out []byte) []byte {
 }
 
 func ipv6CreateRejectPacket(packet []byte, out []byte) []byte {
-	proto, offset, isFragment := ipv6FindUpperProtocol(packet)
-	if isFragment {
+	proto, offset, isFragment, err := IPv6FindUpperProtocol(packet)
+	if err != nil || isFragment {
 		return nil
 	}
 	switch proto {
@@ -333,7 +338,18 @@ func ipv6CreateRejectTCPPacket(packet []byte, out []byte, offset int) []byte {
 	return out
 }
 
-func ipv6FindUpperProtocol(packet []byte) (nextHeader uint8, offset int, isFragment bool) {
+// IPv6FindUpperProtocol walks the ipv6 extension header chain and returns the upper layer protocol, the
+// offset it begins at, and whether the packet is a non-first fragment. Only the RFC 8200 and IANA extension
+// headers below are walked. Everything else, including Mobility (135), HIP (139), Shim6 (140), experimental
+// 253/254, and real upper layer protocols like SCTP or GRE, is terminal. Walking those as extension headers
+// is a firewall bypass, so they fail closed. For a non-first fragment the returned protocol is the fragmented
+// protocol and offset points at the fragment header, there is no transport header to locate. Returns
+// ErrIPv6CouldNotFindPayload if packet is smaller than an ipv6 header or the chain is truncated before a
+// terminal protocol is reached.
+func IPv6FindUpperProtocol(packet []byte) (nextHeader uint8, offset int, isFragment bool, err error) {
+	if len(packet) < ipv6.HeaderLen {
+		return 0, 0, false, ErrIPv6CouldNotFindPayload
+	}
 	nextHeader = packet[6]
 	offset = ipv6.HeaderLen
 
@@ -341,30 +357,31 @@ func ipv6FindUpperProtocol(packet []byte) (nextHeader uint8, offset int, isFragm
 		switch nextHeader {
 		case 0, 43, 60: // Hop-by-Hop, Routing, Destination
 			if len(packet) < offset+2 {
-				return nextHeader, offset, isFragment
+				return nextHeader, offset, isFragment, ErrIPv6CouldNotFindPayload
 			}
 			nextHeader = packet[offset]
 			offset += (int(packet[offset+1]) + 1) << 3
 
 		case 44: // Fragment
 			if len(packet) < offset+8 {
-				return nextHeader, offset, isFragment
+				return nextHeader, offset, isFragment, ErrIPv6CouldNotFindPayload
 			}
+			// Non-first fragments carry no transport header, report the fragmented protocol and stop
 			if packet[offset+2] != 0 || packet[offset+3]&0xf8 != 0 {
-				isFragment = true
+				return packet[offset], offset, true, nil
 			}
 			nextHeader = packet[offset]
 			offset += 8
 
 		case 51: // AH
 			if len(packet) < offset+2 {
-				return nextHeader, offset, isFragment
+				return nextHeader, offset, isFragment, ErrIPv6CouldNotFindPayload
 			}
 			nextHeader = packet[offset]
 			offset += (int(packet[offset+1]) + 2) << 2
 
 		default:
-			return nextHeader, offset, isFragment
+			return nextHeader, offset, isFragment, nil
 		}
 	}
 }
--- a/iputil/packet_test.go
+++ b/iputil/packet_test.go
@@ -474,3 +474,57 @@ func TestCreateICMPEchoResponse_IPv6_NotICMPv6(t *testing.T) {
 	result := CreateICMPEchoResponse(packet, out)
 	assert.Nil(t, result)
 }
+
+func Test_IPv6FindUpperProtocol(t *testing.T) {
+	src := net.ParseIP("fd00::1")
+	dst := net.ParseIP("fd00::2")
+
+	// 8 byte extension/transport stand-ins, first byte is the next header, second is the length field
+	extToTCP := []byte{6, 0, 0, 0, 0, 0, 0, 0}        // len 0 -> 8 bytes, next = TCP
+	extToUDP := []byte{17, 0, 0, 0, 0, 0, 0, 0}       // len 0 -> 8 bytes, next = UDP
+	ahToUDP := []byte{17, 0, 0, 0, 0, 0, 0, 0}        // AH len 0 -> (0+2)<<2 = 8 bytes, next = UDP
+	firstFragToUDP := []byte{17, 0, 0, 1, 0, 0, 0, 1} // frag offset 0, M=1, next = UDP
+	nonFirstFrag := []byte{17, 0, 0, 9, 0, 0, 0, 1}   // frag offset non-zero, next = UDP
+	transport := []byte{0, 80, 1, 187, 0, 0, 0, 0}    // stand-in bytes, IPv6FindUpperProtocol never reads ports
+
+	tests := []struct {
+		name         string
+		nextHeader   uint8
+		payload      []byte
+		wantProto    uint8
+		wantOffset   int
+		wantFragment bool
+		wantErr      error
+	}{
+		{"plain udp", 17, transport, 17, ipv6.HeaderLen, false, nil},
+		{"hop-by-hop then tcp", 0, append(extToTCP, transport...), 6, ipv6.HeaderLen + 8, false, nil},
+		{"routing then tcp", 43, append(extToTCP, transport...), 6, ipv6.HeaderLen + 8, false, nil},
+		{"destination then udp", 60, append(extToUDP, transport...), 17, ipv6.HeaderLen + 8, false, nil},
+		{"ah then udp", 51, append(ahToUDP, transport...), 17, ipv6.HeaderLen + 8, false, nil},
+		{"first fragment walks to transport", 44, append(firstFragToUDP, transport...), 17, ipv6.HeaderLen + 8, false, nil},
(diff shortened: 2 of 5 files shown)


outside_test.go
```diff
@@ -675,3 +679,50 @@ func Test_newPacket_v6ExtHeaderOverflow(t *testing.T) {
...
+
+	// Same confusion but with the unknown protocol sitting after a real extension header, so the walk state
+	// has advanced (protoAt != 6) before it hits the terminal default branch. A HopByHop of 8 bytes is walked
+	// correctly, then SCTP at offset 48 must still fail closed rather than be walked into its own payload.
+	chained := make([]byte, 60)
+	chained[0] = 0x60                             // version 6
+	chained[6] = byte(layers.IPProtocolIPv6HopByHop) // NextHeader = HopByHop extension
+	chained[7] = 64                              // hop limit
+	chained[40] = byte(layers.IPProtocolSCTP)    // HopByHop NextHeader = SCTP
+	chained[41] = 0                              // HopByHop length 0 -> 8 bytes, SCTP begins at offset 48
```
jasikpark: This block isn't gofmt-clean — `gofmt -l .` flags the file, and gofmt wants the end-of-line comments aligned with the two `binary.BigEndian` lines below. `Static checks` will fail on this once the runner rate-limiting clears.

— Caleb + Cache

outside.go
```diff
@@ -405,21 +405,25 @@ func parseV6(data []byte, incoming bool, fp *firewall.Packet) error {
...
-			// Safety check, each ipv6 header has to be at least 8 bytes
-			next = 8
+		default:
+			// Anything else is a real upper layer protocol we don't dissect, so stop here and fail closed.
+			// Walking it like an extension header would let a crafted payload forge a protocol/port pair.
+			fp.Protocol = uint8(proto)
+			fp.RemotePort = 0
+			fp.LocalPort = 0
+			fp.Fragment = false
+			return nil
```
jasikpark: For part of this class the change is a relaxation rather than a tightening, and I think it deserves a `[Unreleased]` changelog entry. Pre-fix, an unknown proto either got a forged proto/port (the bug) or ran the walk off the end and returned `ErrIPv6CouldNotFindPayload`, i.e. an unconditional drop. Packets in that second bucket now reach the firewall and are delivered under an `any`/`any` rule.

Since `proto:` only accepts `any|tcp|udp|icmp` (`examples/config.yml:429`), an `any`/`any` rule is the only handle an operator has here. Concretely: anyone tunneling GRE or IP-in-IP over their overlay today whose traffic works by accident through a tcp/udp match will start dropping after upgrade. Same shape as the 1.11.0 `inbound_action`/`outbound_action` note.

— Caleb + Cache

outside_test.go
```diff
@@ -213,11 +213,15 @@ func Test_newPacket_v6(t *testing.T) {
 	assert.Equal(t, uint16(0), p.LocalPort)
 	assert.False(t, p.Fragment)
 
-	// An unknown protocol packet
+	// An unknown protocol packet, we don't dissect it so we fail closed on its true protocol with no ports
 	b = buffer.Bytes()
 	b[6] = 255 // 255 is a reserved protocol number
 	err = newPacket(b, true, p)
-	require.ErrorIs(t, err, ErrIPv6CouldNotFindPayload)
+	require.NoError(t, err)
```
jasikpark, line 233: This assertion is the proof of the behavior change I flagged on the `default` branch — proto 255 goes from `ErrIPv6CouldNotFindPayload` (always dropped) to classified-with-no-ports, so it's now delivered under an `any`/`any` rule.

— Caleb + Cache

outside.go
```diff
@@ -405,21 +405,25 @@ func parseV6(data []byte, incoming bool, fp *firewall.Packet) error {
 		case layers.IPProtocolAH:
 			// Auth headers, used by IPSec, have a different meaning for header length
 			if dataLen <= offset+1 {
-				break
+				return ErrIPv6CouldNotFindPayload
```
jasikpark: nit: truncation reads more naturally as `ErrIPv6PacketTooShort`, which is what the other truncation checks in this function use (`:354` ICMPv6, `:370` TCP/UDP, `:388` fragment). Same for `:415`. Both paths drop, so it only affects the error text — but it did change an existing test assertion, so worth confirming it's deliberate.

— Caleb + Cache

outside.go
```diff
@@ -405,21 +405,25 @@ func parseV6(data []byte, incoming bool, fp *firewall.Packet) error {
...
 			}
 			next = (int(data[offset+1]) + 2) << 2
 
-		default:
-			// Normal ipv6 header length processing
+		case layers.IPProtocolIPv6HopByHop, layers.IPProtocolIPv6Routing, layers.IPProtocolIPv6Destination:
```
jasikpark: nit: worth saying in the comment that Mobility (135), HIP (139), Shim6 (140) and experimental 253/254 are deliberately left out. They're IANA extension headers using this same `(len+1)*8` convention, so the list reads as incomplete otherwise. Excluding them is right — fail-closed, and it matches both `ipv6_ext_hdr` and `iputil.ipv6FindUpperProtocol`.

— Caleb + Cache

outside_test.go
```diff
@@ -675,3 +679,50 @@ func Test_newPacket_v6ExtHeaderOverflow(t *testing.T) {
...
+	// has advanced (protoAt != 6) before it hits the terminal default branch. A HopByHop of 8 bytes is walked
+	// correctly, then SCTP at offset 48 must still fail closed rather than be walked into its own payload.
+	chained := make([]byte, 60)
+	chained[0] = 0x60                             // version 6
+	chained[6] = byte(layers.IPProtocolIPv6HopByHop) // NextHeader = HopByHop extension
+	chained[7] = 64                              // hop limit
+	chained[40] = byte(layers.IPProtocolSCTP)    // HopByHop NextHeader = SCTP
+	chained[41] = 0                              // HopByHop length 0 -> 8 bytes, SCTP begins at offset 48
+	binary.BigEndian.PutUint16(chained[48:50], 0x1100) // SCTP src port, pre-fix forged NextHeader/length bait
+	binary.BigEndian.PutUint16(chained[50:52], 445)    // SCTP dst port, the real target
```
jasikpark: nit: neither the pre- nor post-fix code ever reads bytes 50-52, so "the real target" reads as more significant than it is — pre-fix the walk lands on ports at 56-60, which are zero. Same for the first case at `:42`.

Related: the only assertion in this second case that discriminates pre/post-fix is `Protocol`; the port assertions pass either way. Its real value is proving the ext-header handoff still fails closed, which the comment could say instead of "the walk state has advanced" — the `default` branch never looks at `offset`/`protoAt`.

— Caleb + Cache

jasikpark (commented):
The exploit reproduces. I copied master's `parseV6` over this branch's test file and `Test_newPacket_v6ExtHeaderConfusion` fails exactly as described, forging `Protocol=0x11` with ports 53/53. `go vet`, the full unit suite, and e2e all pass at this head.

The fix also lines `parseV6` up with `iputil.ipv6FindUpperProtocol` (`iputil/packet.go:336`), which already used this allow-list — cases `0, 43, 60`, fragment, AH, terminal `default`. Pre-fix the two walks disagreed on the same packet: the firewall could classify it as UDP while the reject path correctly built an ICMP unreachable for SCTP. Worth mentioning in the body, it's good independent evidence the shape is right.

Worth mentioning too that this silently fixes a second bug: `next` is declared outside the loop and never reset, so the old `break`-on-truncation path advanced the walk by the *previous* header's length. Dropping the `if next <= 0 { next = 8 }` guard is now provably safe since every remaining branch yields `next >= 8`.

One blocker (gofmt), one behavior-change note worth a changelog entry, and three nits inline.

Heads up that CI here is red entirely from `429 Too Many Requests` downloading `actions/setup-go` — the tests never ran, which is why the gofmt break wasn't caught. Needs a re-run.

— Caleb + Cache

outside.go
JackDoan: there are other header types that are technically legal, but that we should probably never ever see

JackDoan: ah yes the mythical "mobility header"

iputil/packet.go
```diff
@@ -333,38 +338,50 @@ func ipv6CreateRejectTCPPacket(packet []byte, out []byte, offset int) []byte {
...
+func IPv6FindUpperProtocol(packet []byte) (nextHeader uint8, offset int, isFragment bool, err error) {
+	if len(packet) < ipv6.HeaderLen {
+		return 0, 0, false, ErrIPv6CouldNotFindPayload
+	}
 	nextHeader = packet[6]
 	offset = ipv6.HeaderLen
 
 	for {
 		switch nextHeader {
 		case 0, 43, 60: // Hop-by-Hop, Routing, Destination
```
JackDoan, line 358: hop-by-hop is required to appear first, and only once. If we think we could benefit from enforcing that, maybe we should?

iputil/packet_test.go
```diff
@@ -474,3 +475,57 @@ func TestCreateICMPEchoResponse_IPv6_NotICMPv6(t *testing.T) {
 	result := CreateICMPEchoResponse(packet, out)
 	assert.Nil(t, result)
 }
+
+func Test_IPv6FindUpperProtocol(t *testing.T) {
```
matt-defined, line 479: Should have a test with multiple headers, to verify the walking.

outside.go
```diff
@@ -332,101 +332,57 @@ func parseV6(data []byte, incoming bool, fp *firewall.Packet) error {
...
+	if isFragment {
+		// Non-first fragments carry no transport header, so we have no ports to read
+		fp.RemotePort = 0
+		fp.LocalPort = 0
+		return nil
+	}
 
+	switch layers.IPProtocol(proto) {
+	case layers.IPProtocolICMPv6:
+		if dataLen < offset+6 {
```
matt-defined: I think the minimum length of an ICMPv6 message is actually just four bytes. All of the defined types actually require 8 bytes, but the experimental types might not. Maybe move this into the switch below and distinguish the cases?

matt-defined (changes requested):
I think we should have the chained test. I feel less strongly about needing to support experimental ICMPv6 types.

iputil/packet_test.go
JackDoan: @matt-defined just to be explicit, do you mean something like:

```
		{"hop-by-hop then routing then tcp"...},
```
?

matt-defined: Yep.

outside.go
```diff
@@ -332,101 +332,57 @@ func parseV6(data []byte, incoming bool, fp *firewall.Packet) error {
...
+			fp.LocalPort = binary.BigEndian.Uint16(data[offset+2 : offset+4])
+		} else {
+			fp.LocalPort = binary.BigEndian.Uint16(data[offset : offset+2])
+			fp.RemotePort = binary.BigEndian.Uint16(data[offset+2 : offset+4])
 		}
 
-		protoAt = offset
-		offset = offset + next
+	default:
+		// ESP, NoNextHeader, and any upper layer protocol we don't dissect (SCTP, GRE, etc.). Fail closed:
```
JackDoan:
```suggestion
	default:
		// don't set ports for protocols Nebula doesn't inspect
```

outside.go
```diff
@@ -332,101 +332,57 @@ func parseV6(data []byte, incoming bool, fp *firewall.Packet) error {
...
+		// keep the true protocol with no ports so it can only match an `any` rule, never a forged port pair.
```
JackDoan:
```suggestion
```

iputil/packet_test.go
nbrownus: Added a hop-by-hop, routing, then tcp case to the walker table. It walks both extension headers and lands on TCP at offset 56, so a walker that mis-advanced between hops fails only on this row.

outside.go
nbrownus: Good catch. Moved the length check into the switch: the base is now offset+4 (the RFC 4443 type/code/checksum minimum) and only the echo branch requires offset+6 for the identifier it reads. Added a test with a 4 byte non-echo message that the old offset+6 check dropped.

outside.go
nbrownus: Called them out explicitly now, 135/139/140 and the experimental 253/254 are named in the IPv6FindUpperProtocol doc as deliberately left out (fail closed), so the omission reads as intentional rather than incomplete.

iputil/packet.go
nbrownus: I don't think enforcing it buys us anything here. A misordered or duplicated HBH isn't terminal, so it can't forge a proto/port; worst case we classify a malformed packet the receiving kernel drops anyway, so an allow verdict has no effect. Enforcing order/count adds per-packet state to the walk and makes us stricter than the kernel's own ipv6_skip_exthdr, which doesn't check position either. Leaving it fail-closed-on-unknown. Happy to add strict ordering as a follow-up if you feel strongly.

matt-defined (dismissed):
👍

iputil/packet.go
```diff
@@ -333,38 +338,50 @@ func ipv6CreateRejectTCPPacket(packet []byte, out []byte, offset int) []byte {
...
 			if len(packet) < offset+2 {
-				return nextHeader, offset, isFragment
+				return nextHeader, offset, isFragment, ErrIPv6CouldNotFindPayload
 			}
 			nextHeader = packet[offset]
 			offset += (int(packet[offset+1]) + 2) << 2
 
 		default:
-			return nextHeader, offset, isFragment
+			return nextHeader, offset, isFragment, nil
```
brad-defined, line 389: iputil/packet.go:383 — medium. The terminal (default) case returns without checking that offset is inside the packet, so a chain whose declared length runs past the end of the buffer is now accepted instead of dropped.
  Every walked case bounds-checks before reading, but the declared length is added to offset unchecked, and the default arm returns err == nil with an out-of-range offset. parseV6's TCP/UDP/ICMPv6 arms happen to catch it via dataLen < offset+4,
  but outside.go:381's default arm reads nothing and returns nil. Verified: a 48-byte packet with [6]=60 (Dest Options), [40]=132 (SCTP), [41]=255 yields IPv6FindUpperProtocol → (132, offset=2088, false, nil) on a 48-byte buffer, and newPacket
  returns nil with Protocol=132. On master this same packet returns ErrIPv6CouldNotFindPayload and is dropped. So the PR's claim that "truncated extension headers now return an error instead of stumbling forward" holds only for the transport
  protocols; for everything else a malformed, truncated packet is now waved through to any proto: any rule and re-emitted over the tunnel (and iputil.CreateRejectPacket will happily build a 96-byte reject for it). Suggested fix: in the default
  case, return ErrIPv6CouldNotFindPayload when offset > len(packet).

nbrownus: Nice find, pushed a fix

brad-defined (approved):

wadey (approved):

PR #1841, opened 2026-08-17
Enforce outbound message counter limits to prevent nonce reuse

nbrownus:
Addresses a report that we never limit the outbound message counter (the per-tunnel AEAD nonce). Nothing stopped it running to 2^64 and wrapping to 0, reusing nonces on a live key. Unreachable in practice.

- `RehandshakeAfterMessages` (2^34): the connection manager re-handshakes to roll keys. This is what fires in practice. Sized to the AES-GCM data-volume bound, not the well-known 2^32 figure (that's the *random-nonce* limit and doesn't apply since we use a counter nonce). At 2^34 even 64KB frames stay under ~2^-36 GCM advantage, which is ~daily rolls on a saturated 100 Gbps jumbo link and far less at normal MTU.
- `RejectAfterMessages` (`MaxUint64 - RejectHeadroom`, 2^40 of headroom): hard stop. The send path refuses to encrypt and pins the counter so it can never wrap. `EncryptDanger` enforces the same ceiling so any caller fails loud instead of reusing a nonce. The headroom only needs to exceed the senders that can be mid-reservation at once (~2 per routine); `routines` is capped at 2^16 with a compile-time assert tying the two together.

Notes:
- `shouldSwapPrimary` refuses a tunnel past the rehandshake threshold so a completed rehandshake can't re-promote the spent key and flap.
- Exhaustion teardown deletes the tunnel locally (it can't encrypt a `CloseTunnel`); the peer recovers via recv_error.
- New `messages.tx.exhausted` metric counts drops.

Follow-up: a non-primary tunnel that keeps carrying relay traffic (including relayed handshakes) only reaches the reject backstop, not the 2^34 gate, so it can exceed the data-volume target under one key (still no nonce reuse). Narrow and self-healing, left for later.

<details>
<summary>Why 2^34 (AES-GCM data-volume math)</summary>

We use a deterministic counter nonce, so the classic 2^32 AES-GCM limit (a random-nonce birthday bound) does not apply. The binding constraint is data volume: AES-GCM confidentiality advantage is ~ sigma^2 / 2^128, where sigma is total 16-byte blocks encrypted under one key. Advantage depends only on the counter value and frame size, not link speed; link speed only sets how fast the counter is reached.

At `RehandshakeAfterMessages = 2^34`, sized to the worst realistic frame (64KB):

| Frame            | blocks/pkt | GCM advantage @2^34 | roll @100 Gbps | roll @50 Gbps |
|------------------|-----------:|--------------------:|---------------:|--------------:|
| 64KB (max IP)    |       2^12 |               2^-36 |         ~25 h  |        ~50 h  |
| 9000B (AWS jumbo)|      ~2^9.1 |               2^-42 |        ~3.4 h  |        ~6.9 h |
| 1300B (typical)  |      ~2^6.3 |               2^-47 |        ~30 min |         ~1 h  |

So even a saturated 100 Gbps link full of max-size frames rolls about daily at ~2^-36 advantage (roughly one in 7e10 after a full ~1 PB epoch), and everything realistic is far more conservative. ChaCha20-Poly1305 has no meaningful data-volume limit, so for it this is purely a key-lifetime bound.
</details>

### DIFF
--- a/connection_manager.go
+++ b/connection_manager.go
@@ -323,6 +323,12 @@ func (cm *connectionManager) makeTrafficDecision(localIndex uint32, now time.Tim
 		return closeTunnel, hostinfo, nil
 	}
 
+	if hostinfo.ConnectionState != nil && hostinfo.ConnectionState.messageCounter.Load() >= RejectAfterMessages {
+		// Send path can't encrypt a CloseTunnel notify, so just delete locally; the peer recovers via recv_error.
+		hostinfo.logger(cm.l).Error("Dropping tunnel, message counter is exhausted")
+		return deleteTunnel, hostinfo, nil
+	}
+
 	primary := cm.hostMap.Hosts[hostinfo.vpnAddrs[0]]
 	mainHostInfo := true
 	if primary != nil && primary != hostinfo {
@@ -448,6 +454,11 @@ func (cm *connectionManager) shouldSwapPrimary(current *HostInfo) bool {
 		return false
 	}
 
+	if current.ConnectionState.messageCounter.Load() >= RehandshakeAfterMessages {
+		// This tunnel is being rolled for counter exhaustion, never swap back onto its spent key.
+		return false
+	}
+
 	crt := cm.intf.pki.getCertState().getCertificate(current.ConnectionState.myCert.Version())
 	if crt == nil {
 		//my cert was reloaded away. We should definitely swap from this tunnel
@@ -542,6 +553,15 @@ func (cm *connectionManager) tryRehandshake(hostinfo *HostInfo) {
 		cm.l.Info("Re-handshaking with remote",
 			"vpnAddrs", hostinfo.vpnAddrs,
 			"reason", "current cert version < pki.initiatingVersion",
+		)
+
+		cm.intf.handshakeManager.StartHandshake(hostinfo.vpnAddrs[0], nil)
+		return
+	}
+	if hostinfo.ConnectionState.messageCounter.Load() >= RehandshakeAfterMessages {
+		cm.l.Info("Re-handshaking with remote",
+			"vpnAddrs", hostinfo.vpnAddrs,
+			"reason", "message counter rehandshake threshold reached",
 		)
 
 		cm.intf.handshakeManager.StartHandshake(hostinfo.vpnAddrs[0], nil)
--- a/connection_state.go
+++ b/connection_state.go
@@ -2,6 +2,7 @@ package nebula
 
 import (
 	"encoding/json"
+	"fmt"
 	"log/slog"
 	"sync"
 	"sync/atomic"
@@ -41,7 +42,12 @@ type ConnectionState struct {
 // completed handshake.Result. It seeds messageCounter and the replay window so
 // that the post-handshake message indices already used on the wire don't count
 // as missed traffic in the data plane.
-func newConnectionStateFromResult(r *handshake.Result) *ConnectionState {
+func newConnectionStateFromResult(r *handshake.Result) (*ConnectionState, error) {
+	// Refuse a MessageIndex too big for the replay window: it can only be a bug, and would spin the seed loop below.
+	if r.MessageIndex >= ReplayWindow {
+		return nil, fmt.Errorf("handshake message index %d exceeds replay window", r.MessageIndex)
+	}
+
 	ci := &ConnectionState{
 		myCert:    r.MyCert,
 		initiator: r.Initiator,
@@ -54,7 +60,7 @@ func newConnectionStateFromResult(r *handshake.Result) *ConnectionState {
 	for i := uint64(1); i <= r.MessageIndex; i++ {
 		ci.window.Update(nil, i)
 	}
-	return ci
+	return ci, nil
 }
 
 func (cs *ConnectionState) MarshalJSON() ([]byte, error) {
--- a/connection_state.go
+++ b/connection_state.go
@@ -12,7 +12,18 @@ import (
 	"github.com/slackhq/nebula/noiseutil"
 )
 
-const ReplayWindow = 1024
+const (
+	ReplayWindow = 1024
+
+	// RehandshakeAfterMessages is the outbound counter where the connection manager re-handshakes to roll keys.
+	RehandshakeAfterMessages = uint64(1) << 32
+
+	// RejectAfterMessages is the nonce ceiling enforced by noiseutil; a tunnel here is deleted locally, not notified.
+	RejectAfterMessages = noiseutil.RejectAfterMessages
+)
+
+// RehandshakeAfterMessages must stay below RejectAfterMessages so tunnels roll before the hard send stop.
+const _ = RejectAfterMessages - RehandshakeAfterMessages
 
 type ConnectionState struct {
 	eKey           noiseutil.CipherState
@@ -52,6 +63,16 @@ func (cs *ConnectionState) MarshalJSON() ([]byte, error) {
 		"initiator":       cs.initiator,
 		"message_counter": cs.messageCounter.Load(),
 	})
+}
+
+// NextMessageCounter reserves the next outbound counter, pinning at RejectAfterMessages so it can never wrap.
+func (cs *ConnectionState) NextMessageCounter() (uint64, bool) {
+	c := cs.messageCounter.Add(1)
+	if c >= RejectAfterMessages {
+		cs.messageCounter.Store(RejectAfterMessages)
+		return c, false
+	}
+	return c, true
 }
 
 func (cs *ConnectionState) Curve() cert.Curve {
--- a/connection_state.go
+++ b/connection_state.go
@@ -15,8 +15,8 @@ import (
 const (
 	ReplayWindow = 1024
 
-	// RehandshakeAfterMessages is the outbound counter where the connection manager re-handshakes to roll keys.
-	RehandshakeAfterMessages = uint64(1) << 32
+	// RehandshakeAfterMessages rolls keys inside the AES-GCM data-volume margin (~2^-36 advantage at 64KB frames).
+	RehandshakeAfterMessages = uint64(1) << 34
 
 	// RejectAfterMessages is the nonce ceiling enforced by noiseutil; a tunnel here is deleted locally, not notified.
 	RejectAfterMessages = noiseutil.RejectAfterMessages
@@ -65,7 +65,7 @@ func (cs *ConnectionState) MarshalJSON() ([]byte, error) {
 	})
 }
 
-// NextMessageCounter reserves the next outbound counter, pinning at RejectAfterMessages so it can never wrap.
+// NextMessageCounter reserves the next 1-based counter; RejectAfterMessages is the first we refuse, pinned to not wrap.
 func (cs *ConnectionState) NextMessageCounter() (uint64, bool) {
 	c := cs.messageCounter.Add(1)
 	if c >= RejectAfterMessages {
--- a/connection_state_test.go
+++ b/connection_state_test.go
@@ -10,6 +10,7 @@ import (
 	ct "github.com/slackhq/nebula/cert_test"
 	"github.com/slackhq/nebula/handshake"
 	"github.com/slackhq/nebula/header"
+	"github.com/slackhq/nebula/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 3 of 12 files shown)


connection_manager.go
```diff
@@ -448,6 +454,11 @@ func (cm *connectionManager) shouldSwapPrimary(current *HostInfo) bool {
 		return false
 	}
 
+	if current.ConnectionState.messageCounter.Load() >= RehandshakeAfterMessages {
```
jasikpark, line 457: **question:** This guard is necessary and should stay — I checked what happens without it. The old `shouldSwapPrimary` returns true whenever the tunnel is on a current cert, so the just-abandoned hostinfo gets re-promoted, `tryRehandshake` fires on it again, and it flaps.

What I'm less sure about is the threshold. It uses `RehandshakeAfterMessages`, but a non-primary hostinfo never reaches `tryRehandshake` — that decision requires `mainHostInfo` in `makeTrafficDecision`. So in the state where the peer treats a past-2^32 tunnel as primary and we don't, nothing converges: we refuse to swap, and since only the lower-addressed side swaps at all, the peer won't either. It stays split until the `RejectAfterMessages` backstop.

Narrow — a normal rehandshake promotes on both sides through `unlockedAddHostInfo`, so this needs a pre-existing disagreement from a simultaneous-handshake or roam race to bite. Would it be worth also starting a handshake when refusing, so the pair lands on a fresh tunnel?

— Caleb + Cache

jasikpark (commented):
Reviewed against the files rather than the diff alone, and verified the description's claims rather than trusting them. One question, one instance of your stated follow-up worth naming now, one non-blocking nit. Nothing blocking.

**Verified locally at `8ee8d64`:**

- `gofmt -l` and `go vet ./...` clean; full unit suite, `-race` on the root package and `noiseutil`, and `go test -tags=e2e_testing ./e2e/...` all pass.
- Mutation-tested the new guards so the tests aren't decorative. Neutering `NextMessageCounter` back to a plain `Add(1)` fails `TestConnectionState_NextMessageCounter`; deleting the `shouldSwapPrimary` guard fails `Test_NewConnectionManager_CounterLimits` at its `assert.False`.
- Went looking for a build-tagged `EncryptDanger` that would miss the new check. There isn't one — `boring.go` and `notboring.go` only define `EncryptLockNeeded` and the AESGCM constructor, so `aesgcm.go` and `chachapoly.go` are the complete set.
- No measurable hot-path cost from the two new comparisons. `-count=5`, base vs this branch, on an M1 Max: AESGCM 285-290 vs 285-295 ns/op, ChaChaPoly 1758-1770 vs 1757-1765. Within noise.
- The `deleteTunnel` case in `doTrafficCheck` has a body identical to `Interface.closeTunnel`, so "delete locally, no notify" is exactly what happens.
- An exhausted tunnel can still be rolled: handshakes go out via `hm.outside.WriteTo` directly, not the counter-gated send path. One exception, in a comment below.

The pin-instead-of-saturate choice holds up under concurrency, and it's worth stating why since it's load-bearing: after the crossing, every `Add` returns at least `RejectAfterMessages + 1` and stores back exactly `RejectAfterMessages`, so a pre-add value of `R - 1` occurs exactly once. That's what makes the `c == RejectAfterMessages` log-once condition in `dropExhausted` sound rather than merely probable, and 2^32 of headroom covers any plausible concurrent-sender count.

— Caleb + Cache

inside.go
```diff
@@ -294,7 +304,11 @@ func (f *Interface) SendVia(via *HostInfo,
 		// NOTE: for goboring AESGCMTLS we need to lock because of the nonce check
 		via.ConnectionState.writeLock.Lock()
 	}
-	c := via.ConnectionState.messageCounter.Add(1)
+	c, ok := via.ConnectionState.NextMessageCounter()
+	if !ok {
+		f.dropExhausted(via.ConnectionState, via, c, "Dropping outbound relay packets, tunnel message counter is exhausted")
```
jasikpark: **issue (non-blocking):** The relay follow-up in the description has one instance worth naming now: relayed handshakes come through here. `handshake_manager.go:1089` sends handshake stage0 via `SendVia`, so when the exhausted tunnel *is* the relay, this drop suppresses the handshake that would recover it.

It does self-heal — the relay tunnel reaches `RejectAfterMessages`, the connection manager tears it down, a fresh one comes up — but it's the one drop that delays its own recovery, so it may deserve a line in the follow-up's scope.

— Caleb + Cache

inside.go
```diff
@@ -275,6 +275,16 @@ func (f *Interface) sendTo(t header.MessageType, st header.MessageSubType, ci *C
 	f.sendNoMetrics(t, st, ci, hostinfo, remote, p, nb, out, 0)
 }
 
+// dropExhausted releases the caller's write lock and logs once, on the crossing send, for a spent tunnel.
+func (f *Interface) dropExhausted(cs *ConnectionState, hostinfo *HostInfo, c uint64, msg string) {
```
jasikpark: **nitpick (non-blocking):** No metric for exhaustion drops, and `send`/`sendTo` bump `messageMetrics.Tx` before `sendNoMetrics` reaches here, so Tx over-counts them. That matches the existing `eKey == nil` early return, so it's consistent with what's already here — raising it only because this particular drop is security-relevant enough that an operator might want it graphable.

— Caleb + Cache

connection_state.go
```diff
@@ -12,7 +12,18 @@ import (
...
+
+	// RehandshakeAfterMessages is the outbound counter where the connection manager re-handshakes to roll keys.
+	RehandshakeAfterMessages = uint64(1) << 32
+
+	// RejectAfterMessages is the nonce ceiling enforced by noiseutil; a tunnel here is deleted locally, not notified.
+	RejectAfterMessages = noiseutil.RejectAfterMessages
+)
+
+// RehandshakeAfterMessages must stay below RejectAfterMessages so tunnels roll before the hard send stop.
+const _ = RejectAfterMessages - RehandshakeAfterMessages
```
jasikpark, line 27: **praise:** I inverted the two constants to confirm this isn't decorative — it fails the build with `constant -1 of type uint64 overflows uint64`, and the comment above is what makes that message readable to whoever trips it.

— Caleb + Cache

noiseutil/cipher_state.go
```diff
@@ -1,11 +1,19 @@
...
 import (
+	"errors"
 	"fmt"
+	"math"
 
 	"github.com/flynn/noise"
 )
 
+// RejectAfterMessages is the nonce at which we refuse to encrypt, keeping 2^32 of headroom below wrap.
+const RejectAfterMessages = math.MaxUint64 - (uint64(1) << 32)
```
JackDoan: this is a weird way to write `0xffffffff`, but not a sin

noiseutil/cipher_state_test.go
```diff
@@ -89,6 +89,20 @@ func roundtrip(t *testing.T, enc, dec CipherState) {
 	assert.Equal(t, 16, enc.Overhead())
 }
 
+func TestEncryptRejectsExhaustedCounter(t *testing.T) {
+	encA, _ := buildCipherStates(t, CipherAESGCM)
+	encC, _ := buildCipherStates(t, noise.CipherChaChaPoly)
+	nb := make([]byte, 12)
+
+	for _, cs := range []CipherState{NewCipherStateAESGCM(encA), NewCipherStateChaChaPoly(encC)} {
+		_, err := cs.EncryptDanger(nil, nil, []byte("x"), RejectAfterMessages-1, nb)
```
JackDoan, line 103: this test is self-referential, it checks that the nonce doesn't exceed `RejectAfterMessages`, but doesn't confirm that `RejectAfterMessages` makes sense in-context

noiseutil/cipher_state.go
JackDoan: wait actually I think this is just wrong?  

```
>>> u64max=(1<<64)-1
>>> hex(u64max)
'0xffffffffffffffff'
>>> hex(u64max-(1<<32))
'0xfffffffeffffffff'
```

unless I'm missing something. Wraparounds always trip me up.

connection_manager.go
```diff
@@ -323,6 +323,12 @@ func (cm *connectionManager) makeTrafficDecision(localIndex uint32, now time.Tim
 		return closeTunnel, hostinfo, nil
 	}
 
+	if hostinfo.ConnectionState != nil && hostinfo.ConnectionState.messageCounter.Load() >= RejectAfterMessages {
+		// Send path can't encrypt a CloseTunnel notify, so just delete locally; the peer recovers via recv_error.
```
JackDoan, line 327: if we set the limit at u32-1 (or something) we _could_ send a closeTunnel, but it's probably not worth it for such an edge-case

JackDoan: same for triggering a new handshake?

inside.go
```diff
@@ -275,6 +275,16 @@ func (f *Interface) sendTo(t header.MessageType, st header.MessageSubType, ci *C
...
+	if noiseutil.EncryptLockNeeded {
+		cs.writeLock.Unlock()
```
JackDoan: I don't like that this unlock is so visually distant from the original .Lock() call

noiseutil/cipher_state.go
JackDoan: oh I see. This is to avoid wraparound in the edge case where there are somehow very very many in-flight atomic adds. The u32 rehandshake is defended separately.

Hmm

JackDoan: awh man the counter is u64 on the wire not u32, I'm really losing it lol

noiseutil/cipher_state_test.go
JackDoan: im silly

connection_manager.go
JackDoan: this is what I get for posting before I see the rehandshake mechanism

noiseutil/cipher_state.go
JackDoan: if we want to be air-tight we should limit `routines` to ~1M or something crazy but less than this gap

connection_state.go
matt-defined, line 59: I can't imagine this is a real problem, but if this were also guarded against wrapping I wouldn't have to wonder :)

connection_state.go
```diff
@@ -54,6 +65,16 @@ func (cs *ConnectionState) MarshalJSON() ([]byte, error) {
 	})
 }
 
+// NextMessageCounter reserves the next outbound counter, pinning at RejectAfterMessages so it can never wrap.
+func (cs *ConnectionState) NextMessageCounter() (uint64, bool) {
+	c := cs.messageCounter.Add(1)
+	if c >= RejectAfterMessages {
```
matt-defined, line 77: Nit, If `RejectAfterMessages` were set to 10, I would expect the 10th message to be "fine", but the 11th to be dropped.

connection_state_test.go
```diff
@@ -79,6 +80,43 @@ func runTestHandshake(t *testing.T) (initR, respR *handshake.Result) {
...
+	initR, _ := runTestHandshake(t)
+	ci := newConnectionStateFromResult(initR)
+	ci.messageCounter.Store(RejectAfterMessages - 1)
+
+	f := &Interface{l: test.NewLogger()}
+	hostinfo := &HostInfo{vpnAddrs: []netip.Addr{netip.MustParseAddr("10.0.0.1")}, ConnectionState: ci}
+
+	// The crossing send is refused and the counter is pinned rather than advancing to a real write.
+	f.sendNoMetrics(header.Message, 0, ci, hostinfo, netip.AddrPort{}, []byte{}, make([]byte, 12), make([]byte, mtu), 0)
+	assert.Equal(t, RejectAfterMessages, ci.messageCounter.Load())
```
matt-defined: Since we initialized to `RejectAfterMessages - 1` this would be `RejectAfterMessages` whether or not we pinned the counter.

connection_state_test.go
```diff
@@ -79,6 +80,43 @@ func runTestHandshake(t *testing.T) (initR, respR *handshake.Result) {
...
+	// Continued send attempts stay refused and the counter never wraps
+	for i := 0; i < 10; i++ {
+		_, ok = cs.NextMessageCounter()
+		assert.False(t, ok)
+	}
+	assert.Equal(t, RejectAfterMessages, cs.messageCounter.Load())
+}
+
+// TestSendNoMetricsDropsExhausted drives the real send path to the exhausted drop; nil f.writers means a
+// fall-through to a write panics. The boring-only writeLock unlock in dropExhausted is not exercised here.
```
matt-defined: I think that in the call to `f.sendNoMetrics`, neither `remote`, nor `hostinfo.GetRemote()` are valid, so we try to use a relay. But we have no relays, so we just silently fall off the end.

inside.go
matt-defined: Having a metric for this would also give us something better to check against in the tests.

matt-defined (changes requested):

connection_manager.go
nbrownus: Right, the rehandshake at 2^34 is what actually saves the tunnel here, and the reject ceiling is just the never-reached backstop, so I kept the local delete rather than reserving nonce space for a CloseTunnel.

inside.go
nbrownus: Re-handshaking on refusal is the wrong move: a non-primary tunnel only exists because a newer one already completed for that pair, so rolling it would be re-handshaking a rehandshake. In the split-brain case you describe the tunnel is non-primary for us, so we're barely sending on it and our counter is static, not climbing to the backstop, while the side that treats it as primary rolls it through its own mainHostInfo path. So it converges on whoever is actually using it. Keeping the guard, no extra handshake. (Threshold is RehandshakeAfterMessages, now 2^34.)

connection_state.go
nbrownus: There's a second assert now too, tying RejectHeadroom to maxRoutines so the headroom can't be undercut by a routine-count bump.

inside.go
nbrownus: Added the metric

connection_state.go
nbrownus: Fair, it's 1-based (Add(1), first message is counter 1), so >= `RejectAfterMessages` sends `RejectAfterMessages - 1` messages. Added a comment saying exactly that. Kept `>=` since the boundary is unreachable;

connection_state_test.go
nbrownus: Fixed

inside.go
nbrownus: Moved it

noiseutil/cipher_state.go
nbrownus: Yeah, the headroom is purely the "too many concurrent in-flight adds" guard, separate from the 2^34 rehandshake. Took your air-tight suggestion: added maxRoutines (2^16) with a clamp, plus a compile-time assert that RejectHeadroom stays above it. Also renamed the constant to RejectHeadroom and bumped it to 2^40 so wrap-safety holds for any routine count, not just the current cap, and it no longer looks like the rehandshake value.

connection_state_test.go
nbrownus: Same root cause as the other, the counter-value assert proved nothing since the Add reaches RejectAfterMessages regardless. The rewritten test asserts on the drop metric and the untouched out flag instead, so it actually distinguishes a drop from a fall-through.

connection_state.go
nbrownus: Agreed it's not real but defense in depth and the check is basically free, added it.

JackDoan (approved):

matt-defined (approved):

PR #1843, opened 2026-08-17
use stdlib maps.Values

wadey:
This has been in stdlib since go1.23, no need to use golang.org/x/exp anymore just for this:

- https://pkg.go.dev/maps#Values


jasikpark:
checks are failing

jrwren (approved):

PR #1861, opened 2026-08-26
respect setting for tun.pin_threads_key

JackDoan:


main.go
```diff
@@ -247,22 +247,24 @@ func Main(c *config.C, configTest bool, buildVersion string, l *slog.Logger, dev
 		// onto allowed[0]. The bound UDP port keys the per-instance spread:
 		// distinct across instances sharing a box, stable across restarts.
```
nbrownus: No longer stable across restarts since pid is the default

JackDoan: oof fixed

nbrownus (approved):

PR #1862, opened 2026-08-26
Recompute the transport checksum on self-forwarded packets

### LINKED ISSUE #1863: 🐛 BUG: TCP and UDP to a node's own IPv6 overlay address hang on macOS

master at `657f6ad`. Also reproduces on GitHub's `macos-latest` runner.

macOS (Darwin 25.6.0). Linux is not affected; other BSDs untested.

### Describe the Bug

A macOS node cannot open a TCP connection to, or exchange UDP datagrams with, its **own IPv6 overlay address**: the connect or send times out. Its own IPv4 overlay address works over TCP and UDP, and `ping6` to the own IPv6 address gets replies.

This bites as soon as overlay hosts have DNS names: a node resolving its own name gets an AAAA record, and any IPv6-preferring client hangs where the IPv4-only name would have worked.

On macOS the kernel routes a node's own overlay address through the tun rather than loopback, so this traffic passes through nebula, which has forwarded it back to the host since #501/#808. Those predate IPv6 overlay addresses.

### How to reproduce

On a macOS node with both an IPv4 and an IPv6 overlay address (`V4`, `V6`):

```sh
# TCP
nc -6 -l "$V6" 3000                  # terminal 1
echo self | nc -6 -w 3 "$V6" 3000    # terminal 2: times out

# UDP
nc -6 -u -l "$V6" 3000               # terminal 1
echo self | nc -6 -u -w 1 "$V6" 3000 # terminal 2: never arrives

# controls, all fine
ping6 -c 3 "$V6"
nc -l "$V4" 3000 &  echo self | nc -w 3 "$V4" 3000
```

Any TCP or UDP pair shows it (`python3 -m http.server -b "$V6"` + `curl -6` is how I hit it). #1862 adds `make smoke-self`, which runs these probes against a real nebula on the host; on macOS before the fix: `icmp4=ok tcp4=ok udp4=ok icmp6=ok tcp6=FAIL udp6=FAIL`.

### Logs from affected hosts

Nothing is logged: the node is up, traffic to other nodes is fine, and the timeouts leave no trace, so this is easy to mistake for a host firewall or routing problem.

Defaults, with a certificate that carries an IPv6 overlay address.

### Related

- #38 → #501, #808: the original "macOS can't reach its own nebula IP" report and the forward that fixed it for IPv4.
- #1862: diagnosis and fix, with failing tests.

— Caleb + Cache

jasikpark:
Fixes #1863.

A host that sends TCP or UDP to its own IPv6 overlay address on macOS gets nothing back. The kernel routes self traffic through the tun and nebula forwards it straight back to the kernel (`immediatelyForwardToSelf`, from [#501](https://github.com/slackhq/nebula/pull/501) and [#808](https://github.com/slackhq/nebula/pull/808)). On the way out, xnu has written only the pseudo-header sum into the transport checksum field and left the rest to hardware offload; that state does not survive the trip through userspace, so the packet re-enters with an unfinished checksum and is dropped as corrupt. ICMPv6 and all of IPv4 work.

Two commits, red then green:

1. **Add failing tests for self-addressed overlay traffic.** `inside_test.go` hands the self-forward a TCP and a UDP packet with a wrong checksum over each IP version and expects a valid one back (skips where the branch is compiled out). `make smoke-self` (`.github/workflows/smoke/smoke-self.sh`, plus a `macos-latest` job in `smoke.yml`) runs one nebula on the host and aims icmp/tcp/udp probes at its own v4 and v6 overlay addresses. The unit test states the contract; the smoke test checks it against a real kernel, since whether the kernel accepts what nebula hands back is only answerable there. On macOS today it reports `icmp4=ok tcp4=ok udp4=ok icmp6=ok tcp6=FAIL udp6=FAIL`.

2. **Recompute the transport checksum on self-forwarded packets.** `iputil.SetTransportChecksum` recomputes the TCP or UDP checksum of an IPv4 or IPv6 packet in place, walking IPv6 extension headers to the transport header and leaving fragments, ICMP and anything it cannot locate untouched. The self-forward calls it before writing back to the tun. For UDP the pseudo-header counts the datagram's own Length field rather than the IP payload extent ([RFC 768](https://www.rfc-editor.org/rfc/rfc768), [RFC 8200 section 8.1](https://www.rfc-editor.org/rfc/rfc8200#section-8.1)), so a datagram padded out past what it declares still checksums the way its receiver will. Recomputing unconditionally, rather than completing the partial sum the kernel left, means a checksum that is wrong for any reason comes back right, and the fix carries no assumption about what a particular kernel leaves in the field. With it, `make smoke-self` reports all six ok.

The same forward is compiled on the other BSDs; only macOS has been checked live. A fragmented datagram is declined rather than recomputed, and nothing exercises that path end to end.

The first commit was pushed alone so the macOS `make test` job and the new smoke-self job would show the failure: [Test macos](https://github.com/slackhq/nebula/actions/runs/33094861344) failed all four `TestConsumeInsidePacketSelfTrafficChecksum` cases and [smoke-self](https://github.com/slackhq/nebula/actions/runs/33094861312) reported `tcp6=FAIL udp6=FAIL`. Both are green on the second commit.

https://claude.ai/code/session_01UaJmdSVRvqSd9BitLgDPjQ

### DIFF
--- a/iputil/checksum.go
+++ b/iputil/checksum.go
@@ -1,0 +1,147 @@
+package iputil
+
+import (
+	"encoding/binary"
+
+	"github.com/google/gopacket/layers"
+	"github.com/slackhq/nebula/overlay/checksum"
+	"golang.org/x/net/ipv4"
+	"golang.org/x/net/ipv6"
+)
+
+const udpHeaderLen = 8
+
+// SetTransportChecksum recomputes the TCP or UDP checksum of an IPv4 or IPv6
+// packet in place.
+//
+// A kernel that offloads checksums to the NIC hands a packet to a tun with the
+// transport checksum unfinished: only the pseudo-header sum is in the field and
+// the rest is left for hardware that a tun does not have. A packet written
+// straight back to that tun is dropped on re-entry unless the checksum is
+// completed first. ICMP is left alone; it arrived complete on the kernels this
+// was measured against.
+//
+// So is any packet whose transport header cannot be located: fragments, unknown
+// extension headers and truncated packets. An IPv6 fragment header is declined
+// even when it carries the whole datagram (RFC 6946 atomic fragment), because
+// the walk reports only that a fragment header was present.
+func SetTransportChecksum(packet []byte) {
+	if len(packet) < 1 {
+		return
+	}
+	switch int(packet[0] >> 4) {
+	case ipv4.Version:
+		setTransportChecksum4(packet)
+	case ipv6.Version:
+		setTransportChecksum6(packet)
+	}
+}
+
+func setTransportChecksum4(packet []byte) {
+	if len(packet) < ipv4.HeaderLen {
+		return
+	}
+	ihl := int(packet[0]&0x0f) << 2
+	end := int(binary.BigEndian.Uint16(packet[2:4]))
+	if ihl < ipv4.HeaderLen || end < ihl || end > len(packet) {
+		return
+	}
+	// The checksum covers the whole datagram, which a fragment (MF set or a
+	// non-zero offset) does not carry.
+	if binary.BigEndian.Uint16(packet[6:8])&0x3fff != 0 {
+		return
+	}
+
+	transport, ok := transportExtent(packet[ihl:end], packet[9])
+	if !ok {
+		return
+	}
+	csum := ipv4PseudoheaderChecksum(packet[12:16], packet[16:20], uint32(packet[9]), uint32(len(transport)))
+	writeTransportChecksum(transport, layers.IPProtocol(packet[9]), csum)
+}
+
+func setTransportChecksum6(packet []byte) {
+	if len(packet) < ipv6.HeaderLen {
+		return
+	}
+	end := ipv6.HeaderLen + int(binary.BigEndian.Uint16(packet[4:6]))
+	if end > len(packet) {
+		return
+	}
+
+	// The checksum covers the whole datagram, which a fragment does not carry.
+	// An unknown extension header hides where the transport header starts. A
+	// chain longer than the walk's budget ends it early, at an offset that was
+	// never checked against the packet.
+	proto, offset, _, anyFragment, err := IPv6FindUpperProtocol(packet[:end])
+	if err != nil || anyFragment || offset >= end {
+		return
+	}
+
+	transport, ok := transportExtent(packet[offset:end], proto)
+	if !ok {
+		return
+	}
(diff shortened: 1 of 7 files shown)


iputil/checksum.go
```diff
@@ -0,0 +1,147 @@
...
+// the rest is left for hardware that a tun does not have. A packet written
+// straight back to that tun is dropped on re-entry unless the checksum is
+// completed first. ICMP is left alone; it arrived complete on the kernels this
+// was measured against.
+//
+// So is any packet whose transport header cannot be located: fragments, unknown
+// extension headers and truncated packets. An IPv6 fragment header is declined
+// even when it carries the whole datagram (RFC 6946 atomic fragment), because
+// the walk reports only that a fragment header was present.
+func SetTransportChecksum(packet []byte) {
```
JackDoan, line 28: perhaps you could make use of `FinishChecksum` from segment_linux.go?

jasikpark:
Field test on a real macOS host (Darwin 25.6.0): same host cert and utun, one client instance per run, probing the node's own overlay addresses over ICMP/TCP/UDP against a local echo server.

| nebula | v4 ICMP/TCP/UDP | v6 ICMP | v6 TCP | v6 UDP |
|---|---|---|---|---|
| `03421e3` (tests-only parent) | pass | pass | **timeout** | **timeout** |
| `31300340` (this PR) | pass | pass | pass | pass |

Matches the `smoke-self` CI job: only IPv6 transport traffic to the node's own address is affected, and only the checksum rewrite changes it.

— Caleb + Cache

iputil/checksum.go
jasikpark: `FinishChecksum` seeds from the partial sum the kernel left in the field, which is only right when `VIRTIO_NET_HDR_F_NEEDS_CSUM` says that field holds a pseudo-header sum (and it's `linux && !android`, with `virtio.Hdr` carrying the offsets). utun hands us a bare AF header, so we can't tell an unfinished checksum from a complete one, and seeding from a complete one would corrupt it; the test packets carry a deliberately wrong value for that reason. Recomputing from the IP header is right in both states. It does share `checksum.Checksum` with the offloads work. If you'd like, I can pull the common tail (zero the field, sum with seed, UDP 0→0xffff, store) into a helper both call, but it'd have to move out of the `_linux` file.

— Caleb + Cache

nbrownus (approved):
Tested on MacOS, FreeBSD, OpenBSD, and NetBSD. Only MacOS needs the fix in my testing but it's not hurting anything on the others (they don't ride the hairpin).

jasikpark:
can someone merge this pls? I've not got perms

Issue #1865, opened 2026-09-01
🐛 BUG: In certain configurations, lighthouse will process host queries for its own address


matt-defined:
### What version of `nebula` are you using? (`nebula -version`)

1.11.1

### What operating system are you using?

Linux

### Describe the Bug

In certain scenarios lighthouses can end up processing host queries for itself, and sending itself a punch notification. In this situation, it will then start to initiate a tunnel with itself (but then stop, without sending a handshake) until the timeout is reached.

### Logs from affected hosts

```
time=2026-09-01T14:16:53.531Z level=INFO msg="Handshake message received" vpnAddrs="[10.162.148.9 fdef:99:99:e2b9:aa70:83ee:8502:ec89]" from=172.20.0.1:34534 certName=probe certVersion=2 fingerprint=09596b2813fd2697543c00e619ffd1fb0d32584def24e47e8d52c0e81fe2f446 issuer=b045078d2e3902606653d85e46c3c2cf81ce663814e6f75f337ad00e35493aa8 initiatorIndex=937729865 responderIndex=581034947 handshake="map[stage:2 style:ix_psk0]"
time=2026-09-01T14:16:53.531Z level=INFO msg="Taking new handshake" vpnAddrs="[10.162.148.9 fdef:99:99:e2b9:aa70:83ee:8502:ec89]" localIndex=495786579 remoteIndex=1680675524 certName=probe
time=2026-09-01T14:16:53.531Z level=DEBUG msg="Hostmap vpnIp added" hostMap="map[hostinfo:map[existing:true localIndexId:581034947 vpnAddrs:[10.162.148.9 fdef:99:99:e2b9:aa70:83ee:8502:ec89]] mapTotalSize:2 vpnAddrs:[10.162.148.9 fdef:99:99:e2b9:aa70:83ee:8502:ec89]]"
time=2026-09-01T14:16:53.531Z level=INFO msg="Handshake message sent" vpnAddrs="[10.162.148.9 fdef:99:99:e2b9:aa70:83ee:8502:ec89]" handshake="map[stage:2 style:ix_psk0]" cached=false initiatorIndex=937729865 responderIndex=581034947 certName=probe certVersion=2 fingerprint=09596b2813fd2697543c00e619ffd1fb0d32584def24e47e8d52c0e81fe2f446 issuer=b045078d2e3902606653d85e46c3c2cf81ce663814e6f75f337ad00e35493aa8 from=172.20.0.1:34534
[...]

PR #1866, opened 2026-09-01
Drop hostQueries to our overlay addresses

matt-defined:
Don't bother responding to hostQuery's for our overlay addresses.

Addresses https://github.com/slackhq/nebula/issues/1865

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -34,7 +34,7 @@ type LightHouse struct {
 
 	myVpnNetworks      []netip.Prefix
 	myVpnNetworksTable *bart.Lite
-	// myVpnAddrsTable contaings our overlay host addrs, as opposed to the overlay networks
+	// myVpnAddrsTable contains our overlay host addrs, as opposed to the overlay networks
 	myVpnAddrsTable *bart.Lite
 	punchy          *Punchy
 
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -34,7 +34,9 @@ type LightHouse struct {
 
 	myVpnNetworks      []netip.Prefix
 	myVpnNetworksTable *bart.Lite
-	punchy             *Punchy
+	// myVpnAddrsTable contaings our overlay host addrs, as opposed to the overlay networks
+	myVpnAddrsTable *bart.Lite
+	punchy          *Punchy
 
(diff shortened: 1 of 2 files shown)


jasikpark (approved):
Fix is at the right layer — drops the query in `handleHostQuery` before any response or punch notification, killing the whole self-handshake chain.

Correct use of `myVpnAddrsTable` (host-length prefixes) over `myVpnNetworksTable` — only skips queries for the node's own exact address, not everything in the subnet. Consistent with the existing set-once-at-construction pattern. Test is well-structured: proves the precondition (lighthouse holds a cache entry for its own address), control case (normal query still answered + punch sent), then the real assertion (self-query emits nothing).

One typo nit: "contaings" → "contains" on the new `myVpnAddrsTable` comment.

— Caleb + Cache

matt-defined:
> Fix is at the right layer — drops the query in `handleHostQuery` before any response or punch notification, killing the whole self-handshake chain.
> 
> Correct use of `myVpnAddrsTable` (host-length prefixes) over `myVpnNetworksTable` — only skips queries for the node's own exact address, not everything in the subnet. Consistent with the existing set-once-at-construction pattern. Test is well-structured: proves the precondition (lighthouse holds a cache entry for its own address), control case (normal query still answered + punch sent), then the real assertion (self-query emits nothing).
> 
> One typo nit: "contaings" → "contains" on the new `myVpnAddrsTable` comment.
> 
> — Caleb + Cache

Typo fixed.

jasikpark (approved):

lighthouse.go
```diff
@@ -1158,6 +1161,17 @@ func (lhh *LightHouseHandler) handleHostQuery(n *NebulaMeta, fromVpnAddrs []neti
 		return
 	}
 
+	// Don't respond to requests for us.
+	if lhh.lh.myVpnAddrsTable.Contains(queryVpnAddr) {
+		if lhh.l.Enabled(context.Background(), slog.LevelDebug) {
+			lhh.l.Debug("Ignoring HostQuery for one of my own addresses",
+				"from", fromVpnAddrs,
```
nbrownus: The convention is to log this as `vpnAddrs`, so maybe `fromVpnAddrs` is your the way out. `from` will collide with json logging of udp/relay addr handling and break some folks log ingest.

nbrownus (approved):

Issue #1868, opened 2026-09-02
🐛 BUG: HostQuery fails to resolve underlay addresses for secondary overlay IPs in dual-stack certs (`udpAddrs=[]`)


cyberes3:
### What version of `nebula` are you using? (`nebula -version`)

1.11.0

### What operating system are you using?

Linux (Debian 13), amd64

### Describe the Bug

(I run a self-hosted control plane server that speaks the dnclient API and uses the same binaries.)

Dual-stack deployments use separate per-address tunnel lookups by design (migration). The issue is that lighthouse resolution for a given overlay address does not populate that address's lookup, so traffic to valid overlay IPs fails with udpAddrs=[] even while the same peer is reachable on its sibling overlay address.

This fails randomly. The same host pair can work for hours (or days) and then lose reachability on one overlay address with no config change, no cert rotation, and no obvious tunnel teardown in logs. It affects many hosts across the fleet, not a single bad node or a reproducible-on-demand setup. A dnclient restart on either side usually restores connectivity, which makes it look transient, but the underlying lookup gap can come back unpredictably.

My desktop lost connection to a large number of IPv6 overlay addresses and a dnclient restart fixed it.

Symptom: `Handshake timed out vpnAddrs=[<overlay-ip>] udpAddrs=[]` means no `Handshake message sent for that address`. Remote host receives nothing. The sibling overlay address for the same peer often still works at the same moment.

Why this feels random:
- Tunnels frequently establish on IPv4 first; IPv6 monitoring/targets keep working until they don't, with no Tunnel status dead / Close tunnel / Dropping tunnel for that peer beforehand (unlike peers that fail via normal active probing in the same time window).
- A peer can be registered at the lighthouse with both overlay addresses at boot, reachable on IPv4/LAN, and still time out on IPv6 ~14 hours later with zero traffic on the failing path.
- The same initiator can ping the IPv6 overlay at 100% loss while pinging the IPv4 overlay of the same peer in the same minute gives full udpAddrs, relay, success.
- Restarting dnclient on either side typically fixes it by re-registering addresses and repopulating caches=but there is no reliable way to predict when the next drop will happen.

Querying lighthouse for overlay address X (whether IPv4 or IPv6) should return the underlay addresses and fill X's RemoteList, enabling handshake to that address. What actually happens is that the lookup for the secondary/IPv6 overlay address never gets underlay addresses. Handshake sends nothing. Sibling address (usually cert-primary IPv4) resolves fine via the same lighthouse.

[...]

PR #1869, opened 2026-09-02
improve lighthouse addrMap correctness for multi-address certs

JackDoan:
addresses #1868

### DIFF
--- a/lighthouse.go
+++ b/lighthouse.go
@@ -513,17 +513,15 @@ func (lh *LightHouse) QueryServer(vpnAddr netip.Addr) {
 }
 
 func (lh *LightHouse) QueryCache(vpnAddrs []netip.Addr) *RemoteList {
-	lh.RLock()
-	if v, ok := lh.addrMap[vpnAddrs[0]]; ok {
-		lh.RUnlock()
-		return v
+	rl, ok := lh.findRemoteList(vpnAddrs)
+	if ok {
+		return rl
 	}
-	lh.RUnlock()
 
 	lh.Lock()
 	defer lh.Unlock()
 	// Add an entry if we don't already have one
-	return lh.unlockedGetRemoteList(vpnAddrs) //todo CERT-V2 this contains addrmap lookups we could potentially skip
+	return lh.unlockedGetRemoteList(vpnAddrs) //todo this re-calls unlockedFindRemoteList
 }
 
 // queryAndPrepMessage is a lock helper on RemoteList, assisting the caller to build a lighthouse message containing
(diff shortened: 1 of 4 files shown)


lighthouse.go
```diff
@@ -669,23 +667,58 @@ func (lh *LightHouse) addCalculatedRemotes(vpnAddr netip.Addr) bool {
...
+	return lh.unlockedFindRemoteList(vpnAddrs)
+}
+
+// unlockedFindRemoteList checks addrMap for each of vpnAddrs. It returns the first RemoteList found,
+// and true if that RemoteList is present for all vpnAddrs.
+// If false, it means the addrMap, and possibly the RemoteList, need to be corrected.
+func (lh *LightHouse) unlockedFindRemoteList(vpnAddrs []netip.Addr) (*RemoteList, bool) {
+	var am *RemoteList
+	//todo: if a host with addresses A and B is "split", so it only has address A and a new host has only address B
+	//todo: we don't handly that correctly, I'm pretty sure.
```
JackDoan, line 682: @nbrownus I think we need to trigger making a new RemoteList (or a clone/detach of an existing RemoteList?) if `vpnAddrs` doesn't match `am.vpnAddrs`?

lighthouse.go
```diff
@@ -513,17 +513,15 @@ func (lh *LightHouse) QueryServer(vpnAddr netip.Addr) {
 }
 
 func (lh *LightHouse) QueryCache(vpnAddrs []netip.Addr) *RemoteList {
-	lh.RLock()
-	if v, ok := lh.addrMap[vpnAddrs[0]]; ok {
-		lh.RUnlock()
-		return v
+	rl, ok := lh.findRemoteList(vpnAddrs)
```
jasikpark, line 516: **issue (blocking): reconcile `addrMap` when a v2 certificate's address set contracts**

If a v2 host first registers with `[A, B]` and later presents a v2 certificate containing only `[A]`, this fast path succeeds because `A` still maps to the existing `RemoteList`. `RefreshFromHandshake` then changes `RemoteList.vpnAddrs` to `[A]`, but nothing removes `addrMap[B]`. The retired address can continue reaching stale underlay data, and later reassignment of `B` can cause unrelated mappings to be merged or replaced.

Please reconcile both sides of this relationship for authoritative v2 address sets: add missing keys and remove keys that still point to this list but are absent from the new certificate. V1 handshakes should retain the current non-contracting behavior because their single address is incomplete. A regression test for `[A, B] -> [A]` should verify that `B` no longer reaches the old list.

lighthouse.go
```diff
@@ -669,23 +667,58 @@ func (lh *LightHouse) addCalculatedRemotes(vpnAddr netip.Addr) bool {
...
+	defer lh.RUnlock()
+	return lh.unlockedFindRemoteList(vpnAddrs)
+}
+
+// unlockedFindRemoteList checks addrMap for each of vpnAddrs. It returns the first RemoteList found,
+// and true if that RemoteList is present for all vpnAddrs.
+// If false, it means the addrMap, and possibly the RemoteList, need to be corrected.
+func (lh *LightHouse) unlockedFindRemoteList(vpnAddrs []netip.Addr) (*RemoteList, bool) {
+	var am *RemoteList
+	//todo: if a host with addresses A and B is "split", so it only has address A and a new host has only address B
```
jasikpark, line 681: **nitpick (non-blocking): remove or resolve the admitted incorrect-state TODO**

This is the same reconciliation path implicated by the contraction case above. Once the semantics are decided, could this become either enforced behavior with a test or a concrete tracked limitation rather than leaving a production TODO that says the state is probably handled incorrectly?

jasikpark (commented):
The expansion cases covered here look good, and CI is green. I found one blocking issue in contraction/reconciliation semantics, plus one related non-blocking TODO.

lighthouse.go
JackDoan: My position is that this is a rare enough case, and that there is still a net improvement to correctness in the PR, such that this is not a blocker.

PR #1870, opened 2026-09-04
Support multiple ports/ranges in firewall rules

johnmaguire:
It is very common to need to install firewall rules for a service that uses disparate ports (e.g. 80 & 443.)

Currently, the only way to do this is a rule per port or port range in the firewall config.

This PR makes the config more expressive, by allowing for a list in the `ports` field. During config parse time, we explode these rules into a rule per port range, matching the runtime firewall representation.

Slack thread: https://nebulaoss.slack.com/archives/CS01XE0KZ/p1787262330042939

### DIFF
--- a/firewall.go
+++ b/firewall.go
@@ -1105,6 +1105,9 @@ func (r *rule) sanity() error {
 
 func parsePort(s string) (int32, int32, error) {
 	const notAPort int32 = -2
+	if strings.ContainsRune(s, ',') {
+		return notAPort, notAPort, fmt.Errorf("contains a comma; use a YAML list such as `port: [80, 443]`; `%s`", s)
+	}
 	if s == "any" {
 		return firewall.PortAny, firewall.PortAny, nil
 	}
--- a/firewall.go
+++ b/firewall.go
@@ -329,7 +329,6 @@ func AddFirewallRulesFromConfig(l *slog.Logger, inbound bool, c *config.C, fw Fi
 	if r == nil {
 		return nil
 	}
-
 	rs, ok := r.([]any)
 	if !ok {
 		return fmt.Errorf("%s failed to parse, should be an array of rules", table)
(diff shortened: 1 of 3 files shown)


firewall.go
```diff
@@ -341,47 +340,74 @@ func AddFirewallRulesFromConfig(l *slog.Logger, inbound bool, c *config.C, fw Fi
...
+			portRanges = make([]portRange, len(ports))
+			for j, port := range ports {
+				startPort, endPort, err := parsePort(port)
+				if err != nil {
+					if len(ports) == 1 {
+						return fmt.Errorf("%s rule #%v; %s %s", table, i, errPort, err)
+					}
+					return fmt.Errorf("%s rule #%v; %s #%v %s", table, i, errPort, j, err)
+				}
+				if startPort > endPort {
```
jasikpark, line 402: This `startPort > endPort` check duplicates the one in `firewallPort.addRule` (line ~688). It's justified here since the mock firewall doesn't validate and list entries need per-entry context — but would it live more naturally inside `parsePort` itself (it's the one parsing the range)? Then both single and list paths get it with zero duplication and the `port #j` context still comes from the loop. At minimum, worth a comment explaining why it's checked in both places.

firewall.go
```diff
@@ -341,47 +340,74 @@ func AddFirewallRulesFromConfig(l *slog.Logger, inbound bool, c *config.C, fw Fi
...
+					err = errors.New("start port was lower than end port")
```
jasikpark, line 403: Nit: unlike the parse errors (which include the offending value, e.g. `; `nope``), this error drops the port string. Maybe include it for symmetry: `start port was lower than end port; `%s``, or fold it into the range-parse error in `parsePort`.

firewall.go
```diff
@@ -341,47 +340,74 @@ func AddFirewallRulesFromConfig(l *slog.Logger, inbound bool, c *config.C, fw Fi
...
 			proto = iputil.IPProtocolUDP
-			startPort, endPort, err = parsePort(sPort)
 		case "icmp":
 			proto = iputil.IPProtocolICMP
-			startPort = firewall.PortAny
-			endPort = firewall.PortAny
-			if sPort != "" {
-				l.Warn("ignoring port specification for ICMP firewall rule", "port", sPort)
+			if r.Ports != nil || r.Code != "" {
+				l.Warn("ignoring port specification for ICMP firewall rule", "port", ports)
```
jasikpark, line 378: Cosmetic: this now logs the whole slice, so `code: 3` on an ICMP rule logs `port=[3]` instead of `port=3`. Could log the singular value when it's a code, or join the list.

firewall.go
```diff
@@ -341,47 +340,74 @@ func AddFirewallRulesFromConfig(l *slog.Logger, inbound bool, c *config.C, fw Fi
...
 			}
 		default:
 			return fmt.Errorf("%s rule #%v; proto was not understood; `%s`", table, i, r.Proto)
 		}
-		if err != nil {
-			return fmt.Errorf("%s rule #%v; %s %s", table, i, errPort, err)
+
+		type portRange struct {
```
jasikpark, line 384: Nit: `portRange` is declared inside the rule loop but doesn't depend on loop variables — hoisting it above the `for` loop would read a little cleaner.

firewall.go
```diff
@@ -942,6 +973,21 @@ type rule struct {
 	CASha     string
 }
 
+func stringifyValues(v any) []string {
```
jasikpark, line 976: Edge case, totally acceptable to leave as-is: a nested list like `port: [[80, 443]]` stringifies to `"[80 443]"` and fails with `was not a number; `[80 443]`` — it does fail loudly, just with a slightly confusing message.

firewall_test.go
```diff
@@ -1313,9 +1316,48 @@ func TestNewFirewallFromConfig(t *testing.T) {
...
+	for _, port := range []uint16{22, 80, 81} {
+		assert.True(t, fw.OutRules.match(firewall.Packet{Protocol: iputil.IPProtocolTCP, RemotePort: port}, false, peerCert, caPool))
+	}
+	assert.False(t, fw.OutRules.match(firewall.Packet{Protocol: iputil.IPProtocolTCP, RemotePort: 23}, false, peerCert, caPool))
+	assert.True(t, fw.OutRules.match(firewall.Packet{Protocol: iputil.IPProtocolTCP, Fragment: true}, false, peerCert, caPool))
+
+	conf = config.NewC(test.NewLogger())
+	mf = &mockFirewall{}
+	conf.Settings["firewall"] = map[string]any{"outbound": []any{map[string]any{"port": []any{}, "proto": "tcp", "host": "a"}}}
+	require.EqualError(t, AddFirewallRulesFromConfig(l, false, conf, mf), "firewall.outbound rule #0; port list cannot be empty")
```
jasikpark, line 1348: Nice coverage. Two gaps if you want to round it out (minor): no test for `proto: any` with a port list, and no test exercising the `code`-with-ICMP warning path through the new code.

PR #1877, opened 2026-09-14
Honor tun.pin_threads_key=pid

jasikpark:
Noticed while reviewing the config docs for these options (DefinedNet/nebula-docs#376).

`tun.pin_threads_key` accepts `pid` or `port`, and `pid` is the documented default. The switch in `main.go` logs the operator's choice, but then unconditionally overwrites the key with the bound UDP port:

```go
if ap, err := udpConns[0].LocalAddr(); err == nil && ap.Port() != 0 {
    key = uint64(ap.Port())
}
```

So on any node with a bound port (i.e. nearly all of them) the key is always the port and `pid` is unreachable. That also means co-located instances sharing a listen port stack their readers onto the same cores, which is the case the per-instance key exists to avoid.

This gates the port lookup on the key actually being `port`, and falls back to the PID with a warning when no port is bound (`listen.port: 0` before assignment).


jasikpark:
Closing this — my checkout was stale and I missed that #1861 already fixed exactly this back in August. Sorry for the noise.

PR #1879, opened 2026-09-14
Enumerate Android interfaces without anet

JackDoan:
Follow-up to #1822, targeted at its branch.

## Why

anet needs `-ldflags=-checklinkname=0` because of two `//go:linkname` pulls into `net.zoneCache` and x/net's internal zone cache. It refreshes those after enumerating so that IPv6 zone names keep resolving inside the stdlib on Android. `localAddrs` never resolves a zone and discards link-local addresses anyway, so nebula was requiring the flag, and switching off the linker's check for the whole binary, for a feature it does not use. Every consumer linking an Android artifact would have had to carry the flag too. anet also has an open issue about this with no fix.

## What

- **Vendor the part of anet that matters** into `localaddrs_android.go`: an `RTM_GETADDR` dump on a netlink socket that is never bound, plus `SIOCGIFNAME` per index. Both are permitted to Android apps (`bind` and `RTM_GETLINK` are the denied operations, `SIOCGIFNAME` is in the unprivileged ioctl allowlist), and it is the same approach Bionic's `getifaddrs` takes for app UIDs. Written against `x/sys/unix`, which nebula already depends on. No flags, MTU or MAC are collected since `localAddrs` reads none of them.
- **Android only.** Linux and every other platform keep `net.Interfaces`.
- **Reshape the seam** to one enumerator returning `[]localInterface{Name, Addrs func()}`. One dump yields every interface with its addresses, so Android hands back what it already has instead of re-dumping per interface. Other platforms keep `net.Interfaces` and, because `Addrs` is deferred, still only pay the per-interface lookup for interfaces the allow list accepts. The existing `collectLocalAddrs` test keeps asserting that.
- **Take the bind-free path unconditionally on android.** The sepolicy is keyed to target SDK, not OS release, detecting either needs cgo, and the path is valid everywhere.
- Drop anet from `go.mod`, drop the ldflag from the Makefile's android link line but keep the link step.

## Verified

- `go build ./...`, `go vet`, full root package tests.
- `CGO_ENABLED=0 GOOS=android GOARCH=arm64 go build ./cmd/nebula` links with default ldflags.
- darwin, windows and freebsd builds unaffected.
- The enumerator was checked against `net.Interfaces` + `Addrs` on a Linux host with tun, veth, bridge and physical interfaces (same names, addresses and order) while it was temporarily built for linux. That test is not in the PR: the code is android-only, and the root package's android test binary does not currently compile anyway (`interface_emit_test.go` references `udp.StdConn`, which is not built for android).

Not verified on a real Android device. The syscalls are the same ones anet and Bionic use, but a run inside an app sandbox would be worth doing before merge.

https://claude.ai/code/session_01B15ivF2aJUtLFdqJ4iz6zs


JackDoan:
Superseded by #1881, which targets master directly and drops the hand-rolled netlink code in favor of two small additions to vishvananda/netlink.

PR #1881, opened 2026-09-15
Advertise underlay addresses on Android

JackDoan:
Supersedes #1822 and #1879, squashed into one commit. Co-authored with @johnmaguire, whose diagnosis and lighthouse changes this builds on.

## Problem

On Android 11+ the app sandbox denies `bind()` on netlink route sockets and the `RTM_GETLINK` message, so `net.Interfaces` fails with `EACCES`. `localAddrs` discarded that error and returned nothing, so an Android node advertised no underlay addresses and a peer on the same LAN could only reach it at whatever address a lighthouse observed.

## Change

- **New `nicfinder` package** owns local address discovery, with a per-platform `localInterfaces` finder behind a build tag. `Find` takes an exported `Filter` interface that `*LocalAllowList` satisfies.
- **Every platform but android** keeps `net.Interfaces`.
- **Android** lists addresses with `netlink.AddrList` over a socket that is opened here and never bound. The kernel autobinds it on the first send, which the sandbox permits, and each index is named with `SIOCGIFNAME`, which is on Android's unprivileged ioctl allowlist. This is the same approach Bionic's `getifaddrs` takes for app UIDs. No `unsafe`, no `//go:linkname`, no extra linker flags.
- **Errors are no longer discarded.** A finder returns the interfaces it could read alongside an error for the ones it could not. The lighthouse warns the first time discovery fails and demotes an unchanged failure to Debug so a permanent one does not warn on every update.

## Dependency

Every constructor in vishvananda/netlink binds its socket, so the Android path needs two small additions: `nl.NewNetlinkSocketFromFd` and `netlink.NewHandleFromSockets`. This PR switches nebula to `github.com/DefinedNet/netlink`, a fork that carries them with its module path renamed, imported directly rather than through a `replace` directive and pinned to its `v1.3.2-dn.1` tag. `overlay/tun_linux.go` moves to the fork as well. The fork is based on upstream master, which is ahead of the v1.3.1 tag nebula used before, so the Linux tun code picks up those upstream changes too.

## Verified

- `go build ./...`, `go vet`, and tests for `nicfinder` and the root package on Linux.
- `CGO_ENABLED=0 GOOS=android GOARCH=arm64 go build ./cmd/nebula` links with default ldflags.
- darwin, windows and freebsd builds unaffected.
- The Android finder was compiled for Linux in a throwaway harness and compared with `net.Interfaces` + `Addrs` on a host with tun, veth, bridge and physical interfaces: same names, addresses and order across 21 interfaces.

Not yet verified inside a real Android app sandbox.

https://claude.ai/code/session_01B15ivF2aJUtLFdqJ4iz6zs


go.mod
```diff
@@ -7,43 +7,52 @@ require (
...
+	github.com/prometheus/common v0.66.1 // indirect
+	github.com/prometheus/procfs v0.16.1 // indirect
 	github.com/vishvananda/netns v0.0.5 // indirect
+	go.yaml.in/yaml/v2 v2.4.2 // indirect
+	golang.org/x/mod v0.36.0 // indirect
 	golang.org/x/time v0.5.0 // indirect
+	golang.org/x/tools v0.45.0 // indirect
 )
+
+replace github.com/vishvananda/netlink => github.com/DefinedNet/netlink v0.0.0-20260915175302-69de5ea5c182
```
JackDoan, line 58: probably don't want to merge like this, I upstreamed my change, hopefully they pick it up?

nbrownus: This won't apply to mobile nebula, we need to swap out the import entirely for it to take effect there

nicfinder/localaddrs.go
```diff
@@ -0,0 +1,30 @@
...
+func localInterfaces() ([]localInterface, error) {
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return nil, fmt.Errorf("failed to enumerate local interfaces: %w", err)
+	}
+
+	var errs []error
+	out := make([]localInterface, 0, len(ifaces))
+	for _, i := range ifaces {
+		addrs, err := i.Addrs()
```
nbrownus, line 22: Previously we would filter out interfaces by name before grabbing their addresses. It would be ideal to avoid gathering addresses for ignored interfaces.

JackDoan: that would be nice, but afaik because of how netlink works, and how android has it locked down, we can't do this. We can say "give me the addresses" and then "which interface is this?" but not the other way around

nbrownus: That's a limitation on the android side, if `localInterfaces` took the allow list as input then the platform can choose how to apply it, and doing so would allow us to filter out the degenerative case of a k8s host with hundreds of veths before we fetch each address.

nicfinder/localaddrs_android.go
```diff
@@ -0,0 +1,104 @@
...
+
+	"github.com/DefinedNet/netlink"
+	"github.com/DefinedNet/netlink/nl"
+	"golang.org/x/sys/unix"
+)
+
+// Android denies bind() on netlink route sockets and RTM_GETLINK, so net.Interfaces fails with EACCES.
+// Enumerate with an RTM_GETADDR dump on an unbound socket instead.
+
+// localInterfaces returns every interface that has an address, with its addresses. It lists
```
nbrownus, line 21: nit: slightly misleading since the list is filtered

nbrownus (approved):

